Issue #29, opened 2017-02-03
Sharing boards with other users


lepidas:
As an option is available but not working though. It can find my NC members but when I press on a member name it's not sharing anything. NC server 11

juliushaertl:
Are there any errors in the server/browser log when you try add a new member?

lepidas:
NC server errors:
`Error	index	Doctrine\DBAL\Exception\DriverException: An exception occurred while executing 'INSERT INTO `oc_deck_board_acl`(`board_id`,`type`,`participant`,`permission_edit`,`permission_share`,`permission_manage`) VALUES(?,?,?,?,?,?)' with params [1, "user", "anastasia", null, null, null]: SQLSTATE[HY000]: General error: 1366 Incorrect integer value: 'user' for column 'type' at row 1`

`Debug	deck	DeckMapper SQL: INSERT INTO `*PREFIX*deck_board_acl`(`board_id`,`type`,`participant`,`permission_edit`,`permission_share`,`permission_manage`) VALUES(?,?,?,?,?,?) with 1|user|anastasia|||`

`Debug	deck	DeckMapper SQL: SELECT id, title, owner, color, archived FROM `*PREFIX*deck_boards` WHERE `id` = ? with 1`

juliushaertl:
Ah found the mistake. Fix is in #33. Thanks for reporting @lepidas, would be great if you could test if the fix works for you.

lepidas:
Well sort of working. I changed the 2 files, now it's available to "share" boards with other users but its problem with permissions (share/edit/manage) tickable, it's not persistent, I check "Manage" for example, refresh page and then its unchecked. And one more is if I share a board to X user then not possible to remove X user from the share.
Logs:
When I press the bin to remove a user from a share
`DeckMapper SQL: SELECT id, title, owner, color, archived FROM `*PREFIX*deck_boards` WHERE `id` = ? with 21`

When I check a permission to the shared user through the tick boxes
`DeckMapper SQL: SELECT id, title, owner, color, archived FROM `*PREFIX*deck_boards` WHERE `id` = ? with 21`

lepidas:
and this is from the browser log when I check the permissions boxes
`Possibly unhandled rejection: Error updating ACL [object Object]  angular.min.js:122:8
	f/< https://nextcloud.georgios-voulgarakis.info/apps/deck/js/vendor/angular/angular.min.js:122:8
[...]

PR #33, opened 2017-02-03
Use integer values for Acl type

juliushaertl:
Fix for #29


juliushaertl:
Note to myself ... never test anything on sqlite again :see_no_evil:

PR #56, opened 2017-02-26
Add on label tiles a tooltip with the label name.

roipoussiere:
On each card, there are colored "tiles" who inform which label is associated with this card. It is a good design, but sometimes it's difficult to remember the label color signification, so the user need to open the card to see the label names.

This pull request avoid this by adding on each label tile a tooltip describing the label name, improving the user experience:


juliushaertl:
@roipoussiere Thanks for your contribution. Works like a charm. :+1:

PR #75, opened 2017-03-22
Fix info.xml emoji usage

MariusBluem:
Our App Store does not render those emoji-codes we all know from GitHub 🙄 ...

bildschirmfoto 2017-03-22 um 13 59 58bildschirmfoto 2017-03-22 um 14 00 10

@juliushaertl 

Signed-off-by: Marius Blüm <marius@lineone.io>


juliushaertl:
:+1: Thanks @MariusBluem

Issue #92, opened 2017-04-25
Deck lists all users and all groups to anybody


oparoz:
### Steps to reproduce

1. Create groups
2. Assign different users to different groups
3. Log in as a user who is only a member of one group

### Expected behaviour

User should only be able to share with members of his group

### Actual behaviour

User can see all groups and all users
This is a big privacy issue.

### Server configuration

**Nextcloud version:** 11.0.2

juliushaertl:
@oparoz Thanks for reporting. I'll have a look. The best might be to just use the OCS api to fetch users and groups.

For reference:
/ocs/v1.php/apps/files_sharing/api/v1/sharees

juliushaertl:
Fixed with #94

PR #94, opened 2017-04-26
Use OCS API to search for users/groups

juliushaertl:
Use the OCS API to search for users/groups to share with in order to respect the sharing setting from server to "Restrict users to only share with users in their groups".

Fix for #92


js/service/BoardService.js
```diff
@@ -26,32 +26,82 @@ app.factory('BoardService', function(ApiService, $http, $q){
     };
     BoardService.prototype = angular.copy(ApiService.prototype);
 
-    BoardService.prototype.searchUsers = function(search) {
-        var url = OC.generateUrl('/apps/deck/share/search/'+search);
```
oparoz: I'm not familiar enough with the app, but if you're not calling this endpoint any more, shouldn't it be removed?

juliushaertl: :see_no_evil: Makes sense of course :)

PR #93, opened 2017-04-25
Moved "Add a new Stack" into the top Bar

artemanufrij:
#91 #89
Moved "Add a new Stack" into the top Bar

PR #95, opened 2017-04-27
Moved "Add a new Stack" into the top Bar [rebased]

juliushaertl:
rebased #93 and fixed an positioning issue with Nextcloud 12

Thanks for the contribution @artemanufrij


dirigit:
Confirm generation of stack, visibility of new stack is also given using files from #95 on 11.0.3.

Changing stack entry (edit because of typo) does not save.

juliushaertl:
@dirigit Please open a seperate issue. Would be nice if you could also have a look at the developers console of your browser to see if the click to save generates a request to the server.

Issue #88, opened 2017-04-05
Card title is capped / limited


Aesculapius:
When entering a long card title, for example a sentence, the text is first successfully saved - it seems - but later part of it is capped off. I think around 80 characters is the max length? 

I agree that the title should be kept short and put the rest in the description, but maybe a max length of 255 or something like that is more practical.

juliushaertl:
@Aesculapius Thanks for your report. I cannot reproduce this issue on my testing instances. Can you fill out the issue template for details about the database/server setup and maybe post a screenshot where the text is capped?

Aesculapius:
Sure.

### Steps to reproduce
1. Create a new card containing much text
2. Save the card and navigate away from Deck/Nextcloud
3. Come back to the place and notice a lot of characters are missing.

### Expected behaviour
All title text should be kept intact or - when that's really impossible - one should be notices while entering the title text.

### Actual behaviour
Text is capped off.

### Server configuration
### Server configuration
**Operating system**: Linux arc 3.13.0-53-generic #89-Ubuntu SMP Wed May 20 10:34:39 UTC 2015 x86_64

**Web server:** Apache/2.4.7 (Ubuntu) (apache2handler)

**Database:** mysql 5.5.53

**PHP version:** 7.0.1-4+deb.sury.org~trusty+1
Modules loaded: Core, date, libxml, openssl, pcre, zlib, bcmath, calendar, ctype, dba, dom, hash, fileinfo, filter, ftp, gettext, SPL, iconv, mbstring, PDO, session, posix, Reflection, standard, SimpleXML, soap, sockets, Phar, exif, sysvmsg, sysvsem, sysvshm, tokenizer, xml, xmlreader, xmlwriter, zip, mysqlnd, apache2handler, redis, curl, gd, json, mysqli, pdo_mysql, readline, Zend OPcache

**Nextcloud version:** 11.0.2 (stable) - 11.0.2.7

**Updated from an older Nextcloud/ownCloud or fresh install:**

**Where did you install Nextcloud from:** 

**Signing status:**
<details>
	<summary>Signing status</summary>

```
[]
```

</details>

**List of activated apps:**
<details>
	<summary>App list</summary>
```
Enabled:
 - activity: 2.4.1
 - activitylog: 0.0.1
 - admin_notifications: 1.0.0
 - calendar: 1.5.2
 - comments: 1.1.0
 - contacts: 1.5.3
 - dav: 1.1.1
 - deck: 0.1.2
[...]

PR #96, opened 2017-04-27
Move text fields to clob

juliushaertl:
Title and text fields should not be limited in length

Fix for #88

### DIFF
--- a/appinfo/database.xml
+++ b/appinfo/database.xml
@@ -19,7 +19,7 @@
                 <name>title</name>
                 <type>text</type>
                 <notnull>true</notnull>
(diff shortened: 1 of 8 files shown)


artemanufrij:
@juliushaertl I guess we need a limit for title. sidebar looks crazy with unlimited title:
screenshot_20170428_223629

juliushaertl:
@artemanufrij I've limited all title fields to 100 characters and made the header of the card stretch for multi line content. So entering longer text will no longer be possible.

artemanufrij:
It works fine!
screenshot_20170429_131549

Issue #73, opened 2017-03-21
Show reference to card description


Aesculapius:
It is hard to see if some cards have an optional description, but you wouldn't want to miss the information in there. So maybe a small icon or shade or anything on the card could point you in the direction that extra description is available for that card?

That would save endless opening and closing cards just to see if there's more information...

artemanufrij:
@juliushaertl please assign this issue to me....

juliushaertl:
I'll have a look to add you to the nextcloud github team. Then you can assign yourself to issues as well as push your branches directly to the nextcloud/deck repo. :wink:

juliushaertl:
@artemanufrij You just need to confirm the invitation.

juliushaertl:
Fixed with #98

PR #98, opened 2017-04-28
description marker

artemanufrij:
description marker


artemanufrij:
#73

pixelipo:
Lots of redundant code in that SVG - I suggest you clean it up with this online tool:
https://petercollingridge.appspot.com/svg-editor/

artemanufrij:
@pixelipo done

juliushaertl:
Thanks @artemanufrij 
What do you think about using an icon that is a bit more obvious what it stands for? 
There might be more indicators on a card in the future like due date, ... Can we move the icon just below the title of a card? Maybe .icon-rename that is included in Nextcloud would be ok for that?

artemanufrij:
@juliushaertl give me a moment :)

artemanufrij:
or like this:
screenshot_20170429_122418

juliushaertl:
The second one looks better :+1:

Let's keep it in the upper right corner for now. We can move it below the card title once we have more indicator icons to show. Some padding around the icon would although be nice. :wink:

artemanufrij:
screenshot_20170429_123715

juliushaertl:
Great. Let's merge this. :rocket:

PR #99, opened 2017-04-29
Bump bower dependencies

juliushaertl:

We can upgrade ng-sortable again since https://github.com/a5hik/ng-sortable/issues/365 is fixed with 1.3.8.


juliushaertl:
@artemanufrij Running `make install-deps` should make sure that bower downloads the new versions.

artemanufrij:
@juliushaertl ok. that works. but now I have 14k files. How can I make a "build"?

juliushaertl:
@artemanufrij `make appstore` will generate a .tar.gz file that you can deploy to your Nextcloud instance and doesn't include build dependencies. When running the app directly from git in your nextcloud you just need `make build` but then you will still have all node dependencies.

artemanufrij:
@juliushaertl 
1. `make install-deps` -> ok!
2. `make appstore` ->ok!
copy content of the `dock.tar.gz` file into fresh nextcloud instance 11.0.3
--> same failure `{{ b.title }}`

artemanufrij:
Works as expected

Issue #69, opened 2017-03-14
Click outside the card details page to close it


armaccloud:
### Steps to reproduce

1. Click on any card within your board - the card details appear on the right side of the screen
2. Click outside of the card details (somewhere on the left of the screen)

### Expected behaviour
The card details should close, like it does when clicking the X on the right-hand side

### Actual behaviour
The card stays opened, requiring to click on the little X on the right side of the screen.
In order to work more fluently and faster, it should close when clicked outside of the card.

juliushaertl:
Fixed with #102

PR #102, opened 2017-04-29
hide sidebar when clicked outside

artemanufrij:


artemanufrij:
#69

templates/part.board.mainView.php
```diff
@@ -28,7 +28,7 @@
 	</form>
 	</div>
 </div>
-<div id="board" class="scroll-container">
+<div id="board" class="scroll-container" ng-click="sidebar.show=false">
```
juliushaertl: This will break when you try to open up the same card again. Please use ui-sref to navigate to the board route. 

See: https://github.com/nextcloud/deck/blob/master/templates/part.card.php#L9

juliushaertl:
@artemanufrij btw. are you also on IRC?

templates/part.board.mainView.php
artemanufrij: @juliushaertl ui-sref doesn't work as expected. sidebar will never open.. It was my first try

artemanufrij:
@juliushaertl I'm on line #nextcloud@freenode.net

juliushaertl:
@artemanufrij Sign off message is missing, otherwise great improvement. :+1:

artemanufrij:
Signed-off-by: Artem Anufrij artem.anufrij@live.de

Issue #86, opened 2017-03-30
Start comment from everywhere in the comments section


Aesculapius:
It would be nice if you can click anywhere in the comments section to add a comment. Right now you have to click on the exact row of where the first line would start. It takes time to move you mouse precisely over there, while there's a big blank section that you can click in.

juliushaertl:
Fixed with #105

PR #105, opened 2017-04-30
better behavior for comment field

artemanufrij:
#86 

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -436,7 +436,7 @@ button:hover {
 }
 
 #card-meta {
-	padding-top: 0;
+	height: 100%;
 }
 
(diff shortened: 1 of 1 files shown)


juliushaertl:
The text indicating that there is a description to add is out of place (see [Screenshot](https://cloud.githubusercontent.com/assets/3404133/25567083/70c396ce-2de6-11e7-8d92-93e5996b79dd.png) )

And it would be great if we could get rid of the scrollbars if they are not needed.

artemanufrij:
@juliushaertl switched to flex box

juliushaertl:
Great work @artemanufrij

Issue #108, opened 2017-05-01
delete a project


himbeere:
Deck 0.1.3

Hello Devs.

Maybe a stupid question but how do i delete a project? Can't find any button / setting for that.

thanks and cheers
t.

artemanufrij:
@himbeere 

url/to/my/projectongithub/**settings**

juliushaertl:
@himbeere I guess you mean inside the deck app. Just click the 3-dots menu and then the delete icon.

![2017-05-01-193002_246x65_scrot](https://cloud.githubusercontent.com/assets/3404133/25587501/a4f8a6f0-2ea4-11e7-9792-ea9ecd34a89e.png)

Be aware, that there is currently no confirmation/undo functionality. This is requested in #85

himbeere:
Thanks @juliushaertl but that does not work here. I see the trash can too but when hitting the trash can nothing happens. Mouse over shows "Freigabe löschen" which means "cancel sharing this" or so I guess. 

My System: Apache2.4, PHP-7.1.4,Nextcloud 11.0.3

![img_20170501_202614](https://cloud.githubusercontent.com/assets/631865/25589659/b549da2a-2ead-11e7-94dd-7cd276212335.jpg)

juliushaertl:
Ok thanks for reporting. Seems there is a bug caused by #43. The checks in https://github.com/nextcloud/deck/blob/master/templates/part.navigation.php#L16-L18 are not working correctly. I'll look into that tomorrow.

juliushaertl:
Fixed with #113

Issue #85, opened 2017-03-29
Provide undo-operation on delete


stefan-niedermann:
When one deletes a board or a card, there should be an undo operation available for 4 seconds, so if a user deletes something by accident, it can be easily restored with a single click.

juliushaertl:
Makes sense. Also related to https://github.com/nextcloud/deck/issues/70

juliushaertl:
Maybe as a quick fix we can just delay the deletion for 10 sec, to provide that kind of undo functionality. A good starter issue I think.

juliushaertl:
Basic implementation is in #113 but we should just mark the board as deleted in the database and add a cleanup job to finally remove those boards after some time. Then the user doesn't have to wait the 10 sec. until the request for deleting is sent to the server.

juliushaertl:
implemented with #133

PR #113, opened 2017-05-02
Fix deleting boards and add a simple undo function

juliushaertl:
Fix for #108 and #85 for boards only

Please test @himbeere @artemanufrij

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -86,6 +86,13 @@ button:hover {
 	cursor: pointer;
 }
 
+#app-navigation .app-navigation-entry-utils li.board-delete-undo {
+	background-color: #aa0000;
+	color: #fff;
+	cursor: pointer;
+	width: auto !important;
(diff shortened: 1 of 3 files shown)


artemanufrij:
@juliushaertl will do it when I back home :)

artemanufrij:
@juliushaertl Can't test it. look here:
screenshot_20170502_201746

artemanufrij:
@juliushaertl 
Deleting works as expected for me... And I can stop it...

* it's not possible to delete board instand... I have to wait for the time...

* It's not really clear that you can cancel it by clicking timer label --> maybe a "stop" or "X" button...

juliushaertl:
@artemanufrij Good point, I'll add a Stop text when hovering maybe.

juliushaertl:
Let's merge, since it fixes #108.

Issue #120, opened 2017-05-04
Icons should be reordered...


artemanufrij:
[EDIT] [DELETE] or [DELETE] [EDIT]

![icons](https://cloud.githubusercontent.com/assets/16517095/25713858/6a71c09e-30f6-11e7-9c16-24c36886e30f.png)

artemanufrij:
@juliushaertl if you confirm this issue, so I can fix it...

juliushaertl:
Edit, Delete makes sense, as it is already done in #121 :+1:

pixelipo:
Close?

artemanufrij:
@pixelipo yes. it's already merged

PR #121, opened 2017-05-04
small css changes

artemanufrij:

same margin and order:
screenshot_20170504_201927

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -240,28 +240,30 @@ button:hover {
 	font-size: 12pt;
 	font-weight: 700;
 	overflow: hidden;
-	height: 40px;
+	display: flex;
+	min-height: 40px;
 }
 
 .stack h2 input {
 	padding: 0;
-	width: 80%;
+	width: 100%;
 	margin: 0;
 	font-size: 12pt;
 	font-weight: 700;
 	border: 0;
 	background-color: transparent;
-	float: left;
 	min-height: initial;
 }
 
 .stack h2 button,
 .stack .stack-actions {
 	float: right;
+	margin: 0px 0px 0px 20px;
 }
 
 .stack h2 span {
-	float: left;
+	width: 100%;
+	white-space: normal;
 }
 
 .stack h2 .stack-actions {
@@ -269,7 +271,7 @@ button:hover {
 }
 
 .stack h2:hover .stack-actions {
-	display: inline-block;
+	display: flex;
 	float: right;
 }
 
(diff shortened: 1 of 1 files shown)


artemanufrij:
screenshot_20170504_202706

artemanufrij:
#120

css/style.css
```diff
@@ -241,35 +241,36 @@ button:hover {
...
 	border: 0;
 	background-color: transparent;
-	float: left;
 	min-height: initial;
 }
 
 .stack h2 button,
 .stack .stack-actions {
 	float: right;
+	margin: 3px 1px 3px 19px;
```
juliushaertl: 19px look a bit to stretched. That should be just the same as the margin on the top (inside the grey box).

artemanufrij: @juliushaertl it looks very poor...

screenshot_20170504_215812

artemanufrij: @juliushaertl we could move "description-icon" 1px to the right... for the symmetry

artemanufrij: My favourite:
screenshot_20170504_220649

juliushaertl:
@artemanufrij Can you have a look? The delete button moves out of the container for long titles. Also the text should be vertically aligned to the center.

When fixed can you squash those commits into one and add the signoff stuff? ;)
See http://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git#5201642 for the squashing.

artemanufrij:
@juliushaertl do you mean like this:
screenshot_20170505_201322

css/style.css
```diff
@@ -240,36 +240,38 @@ button:hover {
...
 .stack h2 button,
 .stack .stack-actions {
 	float: right;
+	margin: 0px 0px 0px 20px;
 }
 
 .stack h2 span {
-	float: left;
+	width: 100%;
+    white-space: normal;
```
pixelipo: something wrong with indentation here

artemanufrij: you are right...

juliushaertl:
@artemanufrij I've squashed them for now.  

Btw. feel free to join #nextcloud-deck in IRC :wink:

jancborchardt:
By the way, that Edit and Delete icon (presumably) should ideally be stacked vertically in the popover and have text next to it. See how we did it in the Mail app: https://github.com/nextcloud/mail/pull/344 :)

juliushaertl:
@jancborchardt Makes sense.

@artemanufrij I think we already discussed that on irc. I'll open an issue for that.

PR #125, opened 2017-05-07
Header redesign - fixes #116

### LINKED ISSUE #116: Redesign header for Nextcloud 12
WIth the new app menu in Nextcloud 12 the header doesn't fit well in the design anymore. 

2017-05-02-180747_959x94_scrot

pixelipo:
I tried to take layout ideas from core Nextcloud apps for consistency while keeping all the existing features in the header.

new_header

I added the "Home" button which could be a first step towards eliminating 

It could use some love regarding responsiveness, but that is out of the scope of this PR. Also out of the scope is the idea that the "Add new stack" should be just a "+" icon-button that would trigger a popup menu which would include "New Stack" and "New Board") - similar to how it is handled in Files app.

Board color is now should in a more minimalistic way, with a line under the title.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -126,14 +126,21 @@ button:hover {
 	position: relative;
 	z-index: 120;
 	height: 44px;
-	background-color: #eee;
 }
 
-#board-header h1 {
+#board-header > * {
 	font-size: 14pt;
-	margin: 0;
-	padding: 12px;
+	padding: 7px;
+	margin: 7px;
 	float: left;
+}
+
+#board-header > a {
+	background: url('/core/img/breadcrumb.svg?v=1');
+	background-repeat: no-repeat;
+	background-position: right center;
+	background-size: auto 24px;
+	padding-right: 24px;
 }
 
 #board-actions {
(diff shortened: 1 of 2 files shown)


artemanufrij:
@pixelipo could you move "new stack" to the right, please.

pixelipo:
no problem, @artemanufrij 

I pulled it to the left to keep consistency of user interaction elements accross different Nexcloud apps, like here:
files_header

artemanufrij:
@pixelipo 
but in our case it's not a new board... it's a new stack...

pixelipo:
@artemanufrij the "+" button in Files created both new folders (subgroup of a folder) and new files - and both of those are inside the currently listed folder, so new stack (a subgroup of a board) is comparable to that.

artemanufrij:
@pixelipo ok, than we should move "input" into "tooltop" like files
screenshot_20170507_221124

pixelipo:
I agree. I didn't go that far with this PR because it seemed out of scope, but I can add it if you and @juliushaertl agree.

We would also need an icon to represent "stack".

juliushaertl:
@pixelipo Thank you very much for the pull request. Looks great so far. 

I'm against adding the new board function to a popover as described, since when a board is opened the context is clearly the board. so any actions inside the main content area should just have influence on the current board.

> We would also need an icon to represent "stack".

That would definitely make sense, then we could add a stack button to the top and add further functionality, like rearanging stacks, ... just to a dropdown on that button.

pixelipo:
Ok, I've only moved the input to the right for now. We can discuss enhancing it in a different issue.

JSoko:
Changes are looking good for me :-)

artemanufrij:
@pixelipo can I test it on NC11?

juliushaertl:
@pixelipo Nice work! Can you rebase this to master and sign off your commits?

pixelipo:
Closed in favour of https://github.com/nextcloud/deck/pull/130

PR #130, opened 2017-05-10
Header redesign - fixes #116

### LINKED ISSUE #116: Redesign header for Nextcloud 12

pixelipo:
Just decided to create a new pull request - all in a single commit, synced with master...
---
I tried to take layout ideas from core Nextcloud apps for consistency while keeping all the existing features in the header.

I added the "Home" button which could be a first step towards eliminating

It could use some love regarding responsiveness, but that is out of the scope of this PR. Also out of the scope is the idea that the "Add new stack" should be just a "+" icon-button that would trigger a popup menu which would include "New Stack" and "New Board") - similar to how it is handled in Files app.

Board color is now should in a more minimalistic way, with a line under the title.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -126,14 +126,21 @@ button:hover {
 	position: relative;
 	z-index: 120;
 	height: 44px;
-	background-color: #eee;
 }
 
-#board-header h1 {
+#board-header > * {
 	font-size: 14pt;
-	margin: 0;
-	padding: 12px;
+	padding: 7px;
+	margin: 7px;
 	float: left;
+}
+
+#board-header > a {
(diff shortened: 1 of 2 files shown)


pixelipo:
https://github.com/nextcloud/deck/pull/125 for reference

pixelipo:
is there something blocking the review here, @juliushaertl ?

juliushaertl:
@pixelipo No, just me busy doing other things :wink:

PR #133, opened 2017-05-16
Archive boards

juliushaertl:

- [x] get board permissions on boardlist
- [x] Apply archive filter to the sidebar
- [x] Add unit tests


artemanufrij:
@juliushaertl after moving board into archive it stays in the navigation bar. I guess it should be removed...

juliushaertl:
@artemanufrij This is ready to test. Only unit tests are missing for some new methods.

artemanufrij:
position of context menu is broken:
screenshot_20170528_153029

juliushaertl:
@artemanufrij @pixelipo Please review.

This also fixes https://github.com/nextcloud/deck/issues/85

artemanufrij:
@juliushaertl 
Undo works fine for board but not for card or stack. --> it's ok for me

But: Board will never be deleted. I can restore it every time. I'm waiting since more than 10 minutes.

css/style.css
```diff
@@ -614,23 +622,30 @@ button.button-inline:hover {
...
-	height: 27px;
-	float: left;
-	margin-right: 2px;
+	height: 100%;
+	flex-grow: 1;
 	border: none;
 }
 
 .colorselect .selected {
+	background-image: url(../../../core/img/actions/checkmark.svg);
```
pixelipo: It owuld be better to move his to a new class `icon-selected` or `icon-checkmark`. Then it can be reused. Also, next two lines could be removed.

P.S. do we need `../../..`? on my setup(s), `url(/core/img/actions/checkmarks.svg)` works fine.

css/style.css
```diff
@@ -614,23 +622,30 @@ button.button-inline:hover {
...
+	background-position: center center;
+	background-repeat: no-repeat;
 	opacity: 1.0;
-	border: 1px solid #333333;
+}
+
+.colorselect .dark.selected {
+	background-image: url(../../../core/img/actions/checkmark-white.svg);
```
pixelipo: same as above

css/style.css
```diff
@@ -662,6 +677,16 @@ button.button-inline:hover {
 	cursor: pointer;
 	display: block;
 }
+#boardlist .app-popover-menu-utils {
+	width: 30px;
+	display: inline;
+	position: relative;
+}
+
+.popovermenu ul {
```
pixelipo: This style already exists in core css that is served alongside, so perhaps it;s not needed?

```
.bubble ul, .app-navigation-entry-menu ul, .popovermenu ul {
    display: flex !important;
    flex-direction: column;
}
```

css/style.css
```diff
@@ -680,15 +705,34 @@ button.button-inline:hover {
 
 .cell-board-title {
 	width: 50%;
+}
 
+#boardlist tr.deleted td * {
```
pixelipo: why not just change the opacity of the whole row with `#boardlist tr.deleted {` ?

juliushaertl:
Thanks for testing :+1: I'll merge this now so we have it in todays nightly builds for further testing.

img/deck.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 512 512"><path d="M432.3 448.1h-348c-13.2 0-24-10.8-24-24V264.6c0-13.2 10.8-24 24-24h348c13.2 0 24 10.8 24 24v159.5c0 13.2-10.8 24-24 24zM380.4 89.8H127.8c-7.7 0-14-6.3-14-14v-6.3c0-7.7 6.3-14 14-14h252.6c7.7 0 14 6.3 14 14v6.3c0 7.7-6.3 14-14 14zm19.4 61.8H110.6c-7.7 0-14-6.3-14-14v-6.3c0-7.7 6.3-14 14-14h289.2c7.7 0 14 6.3 14 14v6.3c0 7.7-6.3 14-14 14zm21.6 61.4H94.6c-7.7 0-14-6.3-14-14v-6.3c0-7.7 6.3-14 14-14h326.8c7.7 0 14 6.3 14 14v6.3c0 7.7-6.3 14-14 14z"/></svg>
```
pixelipo: It would be good to keep this in line with the other Nextcloud icons. I will commit an optimized version of the icon to this PR - I hope you don't mind :wink:

pixelipo: Since you merged before my review, I created a new PR #175

css/style.css
```diff
@@ -614,23 +622,30 @@ button.button-inline:hover {
 
 .colorselect {
 	overflow: hidden;
```
juliushaertl: no think that is a leftover from using floats before using flexbox with this PR.

css/style.css
```diff
@@ -65,6 +65,14 @@ button.button-inline:hover {
  * Navigation sidebar
  */
 
+.app-navigation-entry-menu ul {
+	flex-direction: row;
+}
+
+.app-navigation-entry-utils-menu-button {
+	display: block !important;
```
juliushaertl: :+1:

css/style.css
```diff
@@ -614,23 +622,30 @@ button.button-inline:hover {
 
 .colorselect {
 	overflow: hidden;
-	clear: both;
-	padding-top: 4px;
-	padding-left: 4px;
+	border-radius:3px;
+	flex-direction: row;
+	min-width: 240px;
```
juliushaertl: Not sure about that, since we have 2 places with different width using this.

2017-06-08-230340_1155x200_scrot

css/style.css
juliushaertl: /core/ will not work on setups where nextcloud is running inside of a subdirectory, so the ../../../ are needed.

juliushaertl: but classes like icon-selected and icon-selected-dark makes sense.

css/style.css
juliushaertl: :+1: But i need to check that.

css/style.css
juliushaertl: It will also change the opacity of the border that separates the lines, that looked a bit odd. :wink:

img/deck.svg
juliushaertl: Sure. feel free to open a new PR, since this one is already merged :see_no_evil:

juliushaertl:
@pixelipo Thanks for the comments on the css changes. I'll try to address those in a follow up PR tomorrow.

jancborchardt:
Awesome work by you all on the Deck app by the way! :) I want to check it out more too, and I would actually use it for my Tasks if it would use the Nextcloud dav backend (so I can simply switch between Deck and the Tasks app whenever I like :)

juliushaertl:
@jancborchardt Thanks for checking it out :) Integration is also one of my main goals but until now I have not come up with a satisfying solution on how to integrate it with the dav backend. I'll look deeper into this after the next version is released.

TeknoShamin:
I appreciate all the work done here. Is there any thought about making Decks archivable the way Boards and Slides can be?

juliushaertl:
@TeknoShamin Yes, I think archiving stacks would make sense as well. Can you open a new issue for that?

Issue #135, opened 2017-05-17
Can't rename stack title


artemanufrij:
### Steps to reproduce
1. Add a new stack
2. Edit stack title by clicking on "edit" icon
3. Rename stack title
4. Click <ENTER> or on arrow icon
5. Click on Deck-App-Icon (Nextcloud navigation bar)
6. Choose last board
-> You will see old stack title

x9t9:
I think it is related, or same as #89 .
Also in this case , I can reproduce your issue, but pressing `enter` always does work for me..

artemanufrij:
I think we should hide all icons when board will be renamed - like renaming files in NC, FileExplorer (Dolphin, etc.)

PR #137, opened 2017-05-18
rename stack

artemanufrij:


artemanufrij:
Fix for: #135

pixelipo:
works, apart from a small bug - it's impossible to unfocus the input field if you don't make any changes.

artemanufrij:
@pixelipo fixed

pixelipo:
@artemanufrij fixes first part, but second one is still an issue. Please add this:

```
.stack h2 span {
    max-width: 100%;
    white-space: normal;
    overflow: hidden;
    text-overflow: ellipsis;
}
```

artemanufrij:
@pixelipo done

artemanufrij:
@pixelipo pls entry `#nextcloud-deck` on `freenode` :)

css/style.css
```diff
@@ -267,6 +271,8 @@ button:hover {
 .stack h2 span {
 	width: 100%;
 	white-space: normal;
+	overflow: hidden;
+    text-overflow: ellipsis;
```
juliushaertl: indentation is wrong here :wink:

css/style.css
```diff
@@ -254,10 +254,14 @@ button:hover {
...
+	background-color: white;
 	min-height: initial;
 }
 
+.stack form {
+	width: 100%;
+}
+
 .stack h2 button,
 .stack .stack-actions {
```
juliushaertl: the margin should ideally only apply to the buttons, not to the container. otherwise the spacing is a bit to big.

artemanufrij: without `widht: 100%` it looks like this:
screenshot_20170519_185818

juliushaertl: I mean the rule below that :Wink:

.stack h2 button,
 .stack .stack-actions

artemanufrij: ah... ok... i guess we should fix it in an other branch. what do you thin?

PR #139, opened 2017-05-19
added a card control bar

artemanufrij:
Signed-off-by: Artem Anufrij <artem.anufrij@live.de>

PR #140, opened 2017-05-19
Limit stack/card title overflow - fixes #131

### LINKED ISSUE #131: Card title overflows/goes under the card-options button
See screenshot for two (related) problems:

title-overflow

It's really simple to fix - I'll create a PR tomorrow

pixelipo:
Additionaly, this fixes positioning of the action buttons (always in top right corner) as well as an HTML validity issue (`<div>` inside an `<h2>`)

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -244,7 +244,7 @@ button:hover {
 	font-weight: 700;
 	overflow: hidden;
 	display: flex;
-	min-height: 40px;
+	min-height: 42px;
 }
 
 .stack h2 input {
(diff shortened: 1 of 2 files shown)


artemanufrij:
we should use 100% space for card title
screenshot_20170519_130605

artemanufrij:
@pixelipo check pls #139. I think it fixes that issue

pixelipo:
@artemanufrij #139 doesn't fix the stack title issue.

artemanufrij:
@pixelipo you are right. this is for stack issue #137

juliushaertl:
@pixelipo This is now integrated in the other PR, right?

pixelipo:
I think neither this nor #139 are perfect solutions.

I've been thinking about this problem a bit, and it seems that we are trying to fit stuff onto a card that shouldn't really belong there. 

Archive and Delete are **actions**. They are not even the most common actions a user performs on a card - I would argue those are "Edit (title)" and "Add description". So why are they even present in card view? Why don't we move them over to the modal sidebar?
Attachment, Date, Labels, Description indicator are **content**.

So the idea is that card only displays content, while sidebar is used for editing cards (title, labels, description, attachments, due date, archive, delete). What say you, @artemanufrij @juliushaertl ?

If we take that route, it would be good to plan the space for all the possible content indicators that we might want to place on the card, even if we still don't have those functions ready.

Issue #145, opened 2017-05-21
Use default style for header buttons


artemanufrij:
What do you think about following style (including home button)?
![screenshot_20170522_002002](https://cloud.githubusercontent.com/assets/16517095/26287883/73e50c58-3e84-11e7-9810-88998f7113ec.png)

Issue #35, opened 2017-02-03
Improved usability in mobile browsers


poVoq:
There are a few issues with usability and rendering in mobile browsers.

Show stoppers are mainly that there is no drag & drop to move cards, so maybe there could be an alternative way to move cards between stacks via a drop-down menu of some sort?

The other thing is that you have to scroll horizontally a lot, which is not working so well especially as the left-hand menu is triggering it's slide to open feature all the time. Maybe the stacks could be automatically rearranged vertically in a mobile browser?

juliushaertl:
@poVoq Thanks for the comments. Drag and drop works if you hit the right spot, but mobile really needs some improvements. 

> Maybe the stacks could be automatically rearranged vertically in a mobile browser?
I'm not sure if I get that right. Can you describe that a bit more in detail?

poVoq:
Instead of having to scroll horizontally through the stacks, just put them on top of each other in mobile mode so that you can scroll vertically.

mr-bolle:
Hi juliushaertl, thanks for your greath work. 
The last days i check the mobile browser usability.

- Board Details: it isn't possible to share a Board with other user/groups if i use the Deck in Portait-Mode from the Mobile Device. If i change the Screen rotation to Landscape-Mode i can use this Option.
![deck_nextcloud](https://cloud.githubusercontent.com/assets/22414212/23569917/2b821286-0062-11e7-9adb-0d95ae799358.png)

- delete (Stack, Boards): the rename icon is realy close to the trash basket. And it could easy happend to delete the whole Board. A additional confirmation will be helpfull, and don't destroy the work.

- Board Menu overlay: if i would change the Label/Card Description the Card Tile is overlayed from the Board Menu Icon
![deck_nextcloud 2](https://cloud.githubusercontent.com/assets/22414212/23571682/dcb89e32-006a-11e7-895a-3f22ff9f9615.png)

octo777:
Just discovered Deck. Great work. I am excited to see how it develops.  So far for me chrome in android, I can grab the card, but it takes off to the left and scrolls to the beginning of the page and I can drop it in the first deck, but I can never get it to move to the right. If I move to the right edge of the screen, it just doesn't scroll at all.  I could have managed to move it one stack at a time since you can see two at a time, if it wasn't for the fact that when you pick up a card, it auto scrolls you quickly to the left.

bjornnorman:
[...]

PR #146, opened 2017-05-21
header bar and mobile device improvements

artemanufrij:

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -42,12 +42,12 @@ input.input-inline {
 	margin: -4px 0;
 }
 
-button {
+button.button-inline {
 	border: 0;
 	background-color: transparent;
 }
 
-button:hover {
+button.button-inline:hover {
 	border: 0;
 	background-color: transparent;
 }
@@ -175,11 +175,14 @@ button:hover {
 }
 
 #board-header-controls-menu {
-	display: none;
-	padding: 5px 6px 5px 7px;
+  display: none;
 }
 
-#board-header-controls-menu  #popover-controls {
+#board-header-controls-menu button {
+	padding: 15px 20px 15px 20px;
+}
+
+#board-header-controls-menu #popover-controls {
 	display: flex;
 	align-items: center;
 	padding-left: 3px;
--- a/css/style.css
+++ b/css/style.css
@@ -165,12 +165,6 @@ button:hover {
 	margin-left: auto;
 }
 
-#board-header-controls .board-action-button,
-#board-header-controls .board-action-button * {
-	display: flex;
-	align-items: center;
-}
-
 #board-header-controls-menu {
 	display: none;
 	padding: 5px 6px 5px 7px;
@@ -185,6 +179,12 @@ button:hover {
 #board-header-controls-menu .popovermenu{
 	margin-top: 35px;
 	margin-right: 6px;
+}
+
+.board-action-button,
+.board-action-button * {
+	display: flex;
+	align-items: center;
 }
 
 .filter-select {
--- a/css/style.css
+++ b/css/style.css
@@ -140,11 +140,6 @@ button:hover {
 	padding: 0px 12px;
 }
 
-#board-header #home a{
-	position: relative;
-	padding: 0px;
-}
-
 #board-actions {
 	font-size: 10pt;
 	float: right;
@@ -174,6 +169,22 @@ button:hover {
 #board-header-controls .board-action-button * {
 	display: flex;
 	align-items: center;
+}
+
+#board-header-controls-menu {
+	display: none;
+	padding: 5px 6px 5px 7px;
+}
+
+#board-header-controls-menu  #popover-controls {
+	display: flex;
+	align-items: center;
+	padding-left: 3px;
+}
+
+#board-header-controls-menu .popovermenu{
+	margin-top: 35px;
+	margin-right: 6px;
 }
 
 .filter-select {
@@ -333,7 +344,7 @@ button:hover {
 	min-height: 16px;
 }
 
-.card .popovermenu {
+.popovermenu {
 	z-index: 999;
 	opacity: 1;
 	margin-top: 25px;
@@ -341,7 +352,7 @@ button:hover {
 	display: block;
 }
 
-.card .popovermenu.hidden {
+.popovermenu.hidden {
 	display: none;
 }
 
@@ -854,6 +865,10 @@ button:hover {
 	/* hide board actions on mobile */
 	#board-header-controls {
 		display: none !important;
+	}
+
+	#board-header-controls-menu {
+		display: flex;
 	}
 }
 
--- a/css/style.css
+++ b/css/style.css
@@ -128,9 +128,6 @@ button:hover {
 	height: 44px;
 	display: flex;
 	align-items: center;
-}
-
-#board-header > * {
 	font-size: 14pt;
 }
 
@@ -170,6 +167,7 @@ button:hover {
 	display: flex;
 	align-items: center;
 	position: relative;
+	margin-left: auto;
 }
 
 #board-header-controls .board-action-button,
@@ -330,8 +328,8 @@ button:hover {
 	opacity: 1;
 }
 
-.space {
-	width: 100%;
+.app-popover-menu-utils {
+	margin-left: auto;
 	min-height: 16px;
 }
 
--- a/css/style.css
+++ b/css/style.css
@@ -131,8 +131,17 @@ button:hover {
 	font-size: 14pt;
 }
 
+#board-header .crumb {
+	background-image: url(/core/img/breadcrumb.svg?v=1);
+	background-size: auto 24px;
+	background-position: right center;
+	background-repeat: no-repeat;
+	padding: 7px 14px;
+}
+
 #board-header > h1 {
 	white-space: nowrap;
+	padding: 7px;
 }
 
 #board-header #home {
--- a/css/style.css
+++ b/css/style.css
@@ -132,7 +132,7 @@ button:hover {
 }
 
 #board-header .crumb {
-	background-image: url(/core/img/breadcrumb.svg?v=1);
+	background-image: url('../../../core/img/breadcrumb.svg?v=1');
 	background-size: auto 24px;
 	background-position: right center;
 	background-repeat: no-repeat;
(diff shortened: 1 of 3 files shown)


artemanufrij:
Fix for #35 and #145

artemanufrij:
FYI: I tested this request only on NC12

css/style.css
```diff
@@ -126,21 +126,26 @@ button:hover {
 	position: relative;
 	z-index: 120;
 	height: 44px;
+	display: flex;
+	align-items: center;
 }
 
 #board-header > * {
 	font-size: 14pt;
```
pixelipo: you can move this over to `#board-header` declaration now.

css/style.css
```diff
@@ -321,7 +330,7 @@ button:hover {
 	opacity: 1;
 }
 
-.card .card-controls .space {
+.space {
```
pixelipo: `<space>` element is not needed. Instead, you should add `margin-left: auto;` to `#board-header-controls`

css/style.css
```diff
@@ -845,17 +854,11 @@ button:hover {
 
 @media (max-width: 480px) {
 	/* hide board actions on mobile */
-	#board-actions {
+	#board-header-controls {
 		display: none !important;
```
pixelipo: This is a serious deal-breaker for mobile usability. Why don't we move these actions into a ... `.icon-more` popover menu?

templates/part.board.mainView.php
```diff
@@ -5,36 +5,39 @@
 		<p>{{ statusservice.text }}</p></div>
 </div>
 <div id="board-header">
-	<a href="#" title="<?php p($l->t('All Boards')); ?>">
-		<i class="icon icon-home"></i>
-	</a>
+	<div id="home" class="crumb svg last">
```
pixelipo: is there any good reason for using a more complicated HTML?

templates/part.board.mainView.php
```diff
@@ -5,36 +5,39 @@
...
-	<div id="stack-add" class="board-action-button" ng-if="boardservice.canEdit() && checkCanEdit()">
-		<form class="ng-pristine ng-valid" ng-submit="createStack()">
-			<input type="text" placeholder="Add a new stack"
-				ng-focus="status.addStack=true"
-				ng-blur="status.addStack=false"
-				ng-model="newStack.title" required
-				maxlength="100" />
-			<button class="icon icon-add" style="opacity: {{status.addStack ? 1: 0.5}};" type="submit"></button>
-		</form>
+	<div class="space"></div>
```
pixelipo: not needed, as mentioned above.

artemanufrij:
@pixelipo thank you for your feedback. I'll check't when I'm back home...

css/style.css
artemanufrij: It's already in `#board-header`

```
#board-header {
 	width: inherit;
 	color: #333333;
  	position: relative;
  	z-index: 120;
  	height: 44px;
 	display: flex;
 	align-items: center;
  }
```

pixelipo: @artemanufrij I meant that the font-size declaration can also be moved over to there

css/style.css
juliushaertl: @pixelipo Yes, having a menu makes sense, but we should keep dedicated buttons if there is enough space.

templates/part.board.mainView.php
juliushaertl: Let's keep this as it was before.

css/style.css
artemanufrij: @pixelipo @juliushaertl what do you think about this suggestion
screenshot_20170522_181847

pixelipo: I agree 100%, @juliushaertl

@artemanufrij yeah, something like that. But maybe we should stack them vertically, because it seems to be the convention. Archive and Details buttons should then have title visible

artemanufrij:
Current design:
screenshot_20170522_194715

juliushaertl:
I've rebased this to current master. @artemanufrij can you undo the changes from here, as discussed: https://github.com/nextcloud/deck/pull/146/files#r117689496

../../../core/img/places/home.svg will break since the url might contain an index.php or not. Let's use the icon-home css class here here, as it was before.

Otherwise changes look really good.

templates/part.board.mainView.php
pixelipo: That icon is a decorator of the "Home" button (which, by the way, should probably be renamed to "All boards") so it should be part of the CSS, not HTML. `` is for content, `background-image` is for design. So Files app is doing it wrong.

But there is one more thing to consider - accessibility. From that point of view, `title` tag should suffice, but to be 100% compatible with screen readers, we might want to add a screen-reader-only text inside `<a>` tag. Botsrap is using this class for that purpose:
```
.sr-only {
  position: absolute;
  width: 1px;
  height: 1px;
  padding: 0;
  margin: -1px;
  overflow: hidden;
  clip: rect(0, 0, 0, 0);
  border: 0;
}
.sr-only-focusable:active,
.sr-only-focusable:focus {
  position: static;
  width: auto;
  height: auto;
  margin: 0;
  overflow: visible;
  clip: auto;
}
```

css/style.css
pixelipo: found a bug @artemanufrij  - clicking on the "Add new stack" immediately closes the popover...

templates/part.board.mainView.php
artemanufrij: hey @pixelipo If I use your example "home" icon is not visible (on NC12)
screenshot_20170523_192633

artemanufrij: that was the reason because I used a div container

artemanufrij:
It doesn't work on my system (Client: Archlinux; Browser: Chromium/Firefox)
screenshot_20170523_213941

templates/part.board.mainView.php
```diff
@@ -5,36 +5,69 @@
...
-			<input type="text" placeholder="Add a new stack"
-				ng-focus="status.addStack=true"
-				ng-blur="status.addStack=false"
-				ng-model="newStack.title" required
-				maxlength="100" />
-			<button class="icon icon-add" style="opacity: {{status.addStack ? 1: 0.5}};" type="submit"></button>
-		</form>
+	<div id="board-header-controls-menu" class="app-popover-menu-utils button">
+		<button class="card-options icon-more" ng-model="card"></button>
+		<div class="popovermenu hidden">
```
pixelipo: This toggles back to `hidden` when you click on the input field so it's impossible to  add new stack. 

Problem is in [js/directive/appPopoverMenuUtils.js](https://github.com/nextcloud/deck/blob/master/js/directive/appPopoverMenuUtils.js#L40). Possible solution is to modify it so:

`if (event.target !== button && event.target.tagName.toLowerCase() !== 'input') {`

this would prevent closing of a popover every time user clicks on an input element. @juliushaertl what do you think? Maybe we should instead search if popover contains input elements and specifically target those?

templates/part.board.mainView.php
```diff
@@ -5,36 +5,69 @@
...
-		<form class="ng-pristine ng-valid" ng-submit="createStack()">
-			<input type="text" placeholder="Add a new stack"
-				ng-focus="status.addStack=true"
-				ng-blur="status.addStack=false"
-				ng-model="newStack.title" required
-				maxlength="100" />
-			<button class="icon icon-add" style="opacity: {{status.addStack ? 1: 0.5}};" type="submit"></button>
-		</form>
+	<div id="board-header-controls-menu" class="app-popover-menu-utils button">
+		<button class="card-options icon-more" ng-model="card"></button>
```
artemanufrij: @pixelipo: thx... done

templates/part.board.mainView.php
juliushaertl: @pixelipo A better way would be to have a class .popovermenu-noclose and check if that exists on the target element. That way we are more flexible if we have any other elements on a popovermenu in the future.

artemanufrij: @pixelipo I have tryed your suggsestion... it hasn't any effects... same behavior

artemanufrij: `scope.$on('documentClicked', function (scope, event) {` will not be fired... I guess it happens on an other place.

artemanufrij: I think all events in `appPopoverMenuUtils.js` will be ignored... `button.click` does never fire an event

pixelipo: But it is fired in my case. Every time. It's not possible to add a new stack unless I add that `if` condition. button.click does fire that event - it effectively closes the popover. Otherwise, you would have a hanging popover when clicking on "Show archived cards" It is also fired every time you click outside of the popover

Try adding console.log in those functions and you will see how often they are triggered.

css/style.css
```diff
@@ -161,6 +167,35 @@ button:hover {
...
 }
 
+#board-header-controls {
+	display: flex;
+	align-items: center;
+	position: relative;
+	margin-left: auto;
+}
+
+#board-header-controls-menu {
```
pixelipo: This container is much larger than the button `icon-more`, which means that if you click somewhere close to the edge, popover will not be triggered.

templates/part.board.mainView.php
artemanufrij: could you push your branch please?

pixelipo: Will do, just give me an hour or ro

js/directive/appPopoverMenuUtils.js
```diff
@@ -37,12 +37,12 @@ app.directive('appPopoverMenuUtils', function () {
 				e.stopPropagation();
 			});
             scope.$on('documentClicked', function (scope, event) {
-                if (event.target !== button) {
+                /* prevent closing popover if target has no-close class */
+                if (event.target !== button && !$(event.target).hasClass('no-close')) {
```
pixelipo: @juliushaertl @artemanufrij this fixes the popover issue. Is that ok?

juliushaertl: Looks good, will test that later. :+1:

css/style.css
juliushaertl: Fixed with the latest commit.

artemanufrij:
@pixelipo @juliushaertl hey guys! it works! thank you for your support :1st_place_medal:

PR #163, opened 2017-05-26
Update part.card.php

rakekniven:
Improvement discussed at transifex


juliushaertl:
@rakekniven Thanks :wink:

PR #165, opened 2017-05-28
fix css for mobile devices

artemanufrij:


css/style.css
```diff
@@ -856,16 +856,17 @@ button.button-inline:hover {
 	.board-header-controls.app-popover-menu-utils {
 		display: none;
 	}
+	.stack h2 .stack-actions {
```
juliushaertl: This is only needed for mobile.

artemanufrij: @juliushaertl done

PR #166, opened 2017-05-28
Fixes few regressions caused by my last PR.

pixelipo:
Last PR transformed our css to the "mobile-first" paradigm so @media query now targets large screens.
This commit fixes that as well as a share icon in the left sidebar.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -55,21 +55,16 @@ button.button-inline:hover {
 }
 
 #searchbox {
-	display: block !important;
+	display: flex !important;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


artemanufrij:
@pixelipo could you pls pull current master into your branch?

pixelipo:
@artemanufrij I would rather if you reverted your last 2 commits since my PR does the same thing, but with less code...

Shoud I instead merge master and revert your changes?

artemanufrij:
@pixelipo yes

PR #169, opened 2017-05-30
missing translation for 'Add a new stack'

artemanufrij:


artemanufrij:
@pixelipo you are right! Thanks!

templates/part.boardlist.php
```diff
@@ -3,7 +3,7 @@
 		<thead>
 		<tr>
 			<td class="cell-board-bullet"></td>
-			<td class="cell-board-title" width="90%"><?php p($l->t('Board title')); ?></td>
+			<td class="cell-board-title" width="90%"><?php p($l->t('Boards')); ?></td>
```
pixelipo: I don't think this is correct, @artemanufrij - it's a row title, and the row displays board titles.

artemanufrij: @juliushaertl what do you think about?

artemanufrij: @pixelipo @juliushaertl may be just "Title"?

pixelipo: `Title` is better.

By the way, Files app uses `Name` for the same row in their table

artemanufrij: @pixelipo ok, I change it to"Title"

juliushaertl: Yep just title sounds better.

templates/part.board.mainView.php
```diff
@@ -34,10 +34,10 @@
 		<div class="stack" ng-repeat="s in stacks" data-as-sortable-item
 			 data-columnindex="{{$index}}" id="column{{$index}}"
 			 style="">
-			<h2 data-as-sortable-item-handle>
+			 <h2 data-as-sortable-item-handle>
```
pixelipo: you got some weird space left before the `<h2>` tag

artemanufrij: @pixelipo fixed

Issue #170, opened 2017-05-30
Move edit/delete stack icons to the popover menu


juliushaertl:
https://github.com/nextcloud/deck/pull/121#issuecomment-304907357

artemanufrij:
like this:
![screenshot_20170530_230245](https://cloud.githubusercontent.com/assets/16517095/26604894/219d3060-458c-11e7-90df-435fc18f9bb6.png)

pixelipo:
@jancborchardt referring to your comment in #121 - I definitely agree with it (better accessibility, lower learning curve, cleaner UI), but I made a PR #173 which deprecates of of the two existing icons (Edit). Should that single (Delete) icon still be moved to the popover?

@juliushaertl you probably have plans for more actions related to stack (Move, Copy, Archive, Archive All Cards...), so popover definitely makes sense, but should we wait until we populate it more?

jancborchardt:
@pixelipo nice, good call with making rename more seamless! :)

A single icon should never be in a popover, but please do add a tooltip to it which appears on hover (using the JS tooltip function). But yeah as you say maybe it makes most sense to do a popover right away when more actions come in.

jancborchardt:
@pixelipo @artemanufrij btw, if you folks come to Berlin some time, for example for our [Nextcloud conference](https://nextcloud.com/conf/) – we should make a bike tour! https://www.reddit.com/r/FixedGearBicycle/comments/6eib4c/breakbrake17_transfer_raw_cinelli_gazzetta/ ;)

artemanufrij:
@jancborchardt Cool! I plan to visit NC conference

PR #172, opened 2017-05-30
redesign stack context menu

artemanufrij:

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -309,7 +309,7 @@ button.button-inline:hover {
 }
 
 .card .card-controls .card-options, .stack-header .card-options {
-	opacity: 0;
+	opacity: 0.25;
 }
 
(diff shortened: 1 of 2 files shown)


artemanufrij:
Fix for #170

juliushaertl:
@artemanufrij Nice. One small adjustment, I think the 3-dots button should be in the same column alignment as the ones from the card.
2017-05-31-093538_62x158_scrot

artemanufrij:
@juliushaertl ok... I will change it. my focus was on right margin.

pixelipo:
@juliushaertl I'm not so sure these should be aligned. After all, card is visibly narrower than stack.

artemanufrij:
We can hide "..." - button if cursor is out of card/stack. currently it has opacity: 0.25.

artemanufrij:
fixed in #173

PR #173, opened 2017-05-31
Click title to edit

pixelipo:
Allows editing stack title by clicking on the title itself. Deprecates edit icon in the stack header.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -250,7 +250,7 @@ button.button-inline:hover {
 
 .stack h2 button {
 	margin-left: auto;
-    display: flex;
+	display: flex;
 	opacity: 0.25;
 }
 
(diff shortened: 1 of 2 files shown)


juliushaertl:
@pixelipo Thanks, that makes sense. Two small things:

- Can we introduce a css class .inline-editable that will show cursor: text; on hover so it is clear to the user that he can enter text when clicking (this should also apply to the card title in the sidebar
- Pressing enter should end the edit mode (atm. it just saves but the input stays active)

pixelipo:
@juliushaertl in my case it loses focus in both cases (blur and submit). Can you please check again?

cursor:text is showing on hover, but I have extracted it into a new class if you think it will be useful in other cases as well

artemanufrij:
@pixelipo Enter/Return doesn't work in Firefox or Chromium

artemanufrij:
@pixelipo I have fixed it

Issue #123, opened 2017-05-06
ER: Minimized Left Panel


aproposnix:
It would be nice if I could minimize the left panel -  perhaps only showing the round circles to represent the other kanbans (without text). 

I've already been in a situation where I was going over a kanban with a client who was able to see the names of my other kanbans (including one for a competitor).

artemanufrij:
@aproposnix @juliushaertl what do you think about a solution like files:
![screenshot_20170601_223559](https://cloud.githubusercontent.com/assets/16517095/26699779/3c5b2502-471b-11e7-91cf-4d0d59316deb.png)
![screenshot_20170601_223655](https://cloud.githubusercontent.com/assets/16517095/26699780/3c5c916c-471b-11e7-9ddb-7e5939f5ef51.png)

juliushaertl:
@artemanufrij Let's do it like this, since otherwise the sidebar is also not accessible on mobile which is required after #133. Do you want to prepare a PR?

artemanufrij:
I have already a PR on my local system

artemanufrij:
issue fixed in #177

aproposnix:
Sorry for the late reply, but yeah, this is great! Thank you!

PR #177, opened 2017-06-09
toggle navibar visibility

artemanufrij:

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -910,6 +910,12 @@ button.button-inline:hover {
 	}
 }
 
+@media (max-width: 768px) {
+	#app-navigation-toggle {
+		display: none !important;
+	}
+}
+
 /**
  * Markdown rendering
  */
--- a/css/style.css
+++ b/css/style.css
@@ -135,7 +135,7 @@ button.button-inline:hover {
 	color: #333333;
 	position: relative;
 	z-index: 120;
-	height: 44px;
+	min-height: 44px;
 	display: flex;
 	align-items: center;
 	font-size: 14pt;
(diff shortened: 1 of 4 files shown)


artemanufrij:
fix for #123

css/style.css
```diff
@@ -902,6 +910,12 @@ button.button-inline:hover {
 	}
 }
 
+@media (max-width: 768px) {
+	#app-navigation-toggle {
+		display: none !important;
```
pixelipo: @juliushaertl didn't you mention somewhere that navigation toggle should be available on mobile as well?

js/controller/AppController.js
```diff
@@ -24,6 +24,10 @@ app.controller('AppController', function ($scope, $location, $http, $route, $log
     $rootScope.sidebar = {
         show: false
     };
+    $rootScope.navibar = {
+        show: true
```
pixelipo: Any chance we could animate this?

artemanufrij: @pixelipo I have tried to animate it like files but I didn't found how it works in files.

It was my fist try :(

css/style.css
artemanufrij: I think we could dispense with navigation bar completely if we move all buttons/controls into board view.

js/controller/AppController.js
pixelipo: They have a different approach - They are using CSS to move (translate) the main container (right one) over the navigation pane :)

css/style.css
juliushaertl: Yes let's add it to mobile as well, since otherwise it is not possible to access archived boards there.

js/controller/AppController.js
juliushaertl: Maybe we can just use the same mechanism as in files. :wink:

css/style.css
juliushaertl: One more thing that needs to be fixed is that the button should also be visible on the boards list overview ;)

juliushaertl:
> App doesn't save "toggled" state, which might be issue for the case reported in #123

I guess that is ok for now, since it will stay hidden while using deck and just reappear on reload.

Issue #178, opened 2017-06-09
header right controls have broken margin compared to other apps


artemanufrij:
![spectacle wl8511](https://user-images.githubusercontent.com/16517095/26996449-96715310-4d72-11e7-89a5-b4f175302ddd.png)

**Top**: Deck
**Bottom**:  all other apps

PR #179, opened 2017-06-09
fixed avatar margin

artemanufrij:


artemanufrij:
fix for #178

artemanufrij:
@juliushaertl do you mean to add a new class `assigned-users` in board list?

PR #181, opened 2017-06-10
added board header for board list

artemanufrij:


artemanufrij:
New header for board list:
screenshot_20170610_104332

PR #182, opened 2017-06-10
some design changes in board list

artemanufrij:

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -718,7 +718,7 @@ button.button-inline:hover {
 	opacity: 0.5;
 }
 
-#boardlist tr:last-child td {
+#boardlist tbody tr:last-child td {
 	opacity: 0.5;
 }
 
--- a/css/style.css
+++ b/css/style.css
@@ -676,6 +676,10 @@ button.button-inline:hover {
 	overflow: scroll;
 }
 
+#boardlist thead td {
+	color: #999;
+}
+
 #boardlist td .board-bullet {
 	width: 30px;
 	height: 30px;
@@ -684,11 +688,6 @@ button.button-inline:hover {
 	position: relative;
 	cursor: pointer;
 	display: block;
-}
-#boardlist .app-popover-menu-utils {
-	width: 30px;
-	display: inline;
-	position: relative;
 }
 
 .popovermenu ul {
@@ -708,7 +707,7 @@ button.button-inline:hover {
 }
 
 .cell-board-bullet {
-	width: 60px;
+	width: 50px;
 }
 
 .cell-board-title {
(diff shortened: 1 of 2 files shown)


artemanufrij:
* column title: color like `files`
* '...' - menu: opacity like `files`
* dot-icon: margin and size like `files`

artemanufrij:
@juliushaertl done

css/style.css
```diff
@@ -738,8 +741,20 @@ button.button-inline:hover {
 
 #boardlist td .app-popover-menu-utils {
 	float: right;
+	width: 30px;
+	display: inline;
+	position: relative;
+}
+
+#boardlist td .app-popover-menu-utils button {
+	opacity: 0.3;
```
pixelipo: I don't really see the point in using 4 different opacity levels. Difference between 0.3 and 0.5 or 0.5 and 0.7 is barely noticable and it doesn't give away any additional message to the user.

I'm working on cleaning up opacity definitions in nextcloud/server and the idea is to use 0.5 and 1 for most of the situations, and I think this is a perfect example. Triple-dot button should  have 50% opacity normally and 100% when hovered/focused/selected wherever it is used.

css/style.css
```diff
@@ -738,8 +741,20 @@ button.button-inline:hover {
...
 }
 
+#boardlist td .app-popover-menu-utils:hover button {
+	opacity: 0.7;
```
pixelipo: Part of the above comment - `opacity:1`

In the long term, this should all be moved into `.button-inline` and `.button-inline:hover, .button-inline: focus`.

templates/part.boardlist.php
```diff
@@ -24,7 +24,7 @@
 			<td>
 				<div class="hint"></div>
 				<div class="app-popover-menu-utils" ng-if="b.deletedAt == 0">
-					<button class="icon icon-more button-inline" title="More actions"></button>
+					<button class="icon icon-more button-inline" title="<?php p($l->t('More actions')); ?>"></button>
```
juliushaertl: :+1:

juliushaertl:
Great @artemanufrij. Thank you :)

templates/part.boardlist.php
pixelipo: Great catch!

css/style.css
artemanufrij: @pixelipo currently it has same behavior like files

pixelipo: That doesn't make it correct, @artemanufrij - plus it gives us a third opacity behaviour for the same button:
- triple-dot in navigation bar: 0.5->1
- triple-dot on card: 0.25->1
- triple-dot on board list: 0.3->0.7

I will make a PR to Files to fix it there as well.

artemanufrij: @pixelipo you are right... i focused only board list

PR #190, opened 2017-06-16
Toggle navibar on mobile

artemanufrij:

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -118,48 +118,66 @@ button.button-inline:hover {
 }
 
 #board {
-	position: relative;
-	height: 100%;
+	position: absolute;
 	white-space: nowrap;
 	overflow: auto;
 	z-index: 100;
 	width: 100%;
+	bottom: 0px;
+	top: 44px;
 }
 
 #board #innerBoard {
 	padding: 10px;
 }
 
-#board-header {
+#controls {
 	width: inherit;
-	color: #333333;
-	position: relative;
 	z-index: 120;
-	min-height: 44px;
 	display: flex;
 	align-items: center;
 	font-size: 14pt;
+	position: inherit;
+	width: 100% !important;
+    padding-left: 44px;
 }
 
-#board-header .crumb {
-	background-image: url('../../../core/img/breadcrumb.svg?v=1');
-	background-size: auto 24px;
-	background-position: right center;
-	background-repeat: no-repeat;
-	padding: 7px 14px;
+#controls .crumb, #controls a {
+	top: 12px;
 }
 
-#board-header > h1 {
+#controls > h1 {
 	white-space: nowrap;
 	padding: 7px;
 }
 
+#controls button {
+	height: inherit;
+}
+
+#controls input[type='text'] {
+	padding: 4px 5px;
+	margin: 3px 3px 3px 0px;
+	border: 0px none transparent;
+	min-height: initial;
+	background-color: transparent;
+	height: inherit;
+}
+
+#controls #button-home {
+	background-image: url(../../../core/img/places/home.svg);
+	background-position: 8px center;
+	background-repeat: no-repeat;
+	padding: 10px 24px 10px 14px;
+}
+
 #app-navigation-toggle {
-	position: relative;
-	top: 0px;
-	width: 44px;
-	height: 44px;
+	width: 45px;
+	height: 40px;
 	cursor: pointer;
+	opacity: 1;
+	display: inline-block !important;
+	position: fixed;
 }
 
 .board-header-controls {
@@ -210,17 +228,13 @@ button.button-inline:hover {
 	background-color: rgba(240,240,240,.9);
 	border-radius: 3px;
 	margin: 3px;
+	display: flex;
+	align-content: center;
 }
 
 #stack-add > form {
 	display: flex;
-}
-
-#stack-add input {
-	padding: 4px 5px;
-	border: 0px none transparent;
-	min-height: initial;
-	background-color: transparent;
+	align-content: center;
 }
 
 #stack-add input:invalid {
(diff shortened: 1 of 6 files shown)


artemanufrij:
fix for #185

juliushaertl:
2017-06-17-102112_765x49_scrot

- New stack button has lost some padding 
- The menu toggle button doesn't look centered to the home icon / board title
- The toggle button should also show up on desktop systems, as it was before

Apart from that :+1:

artemanufrij:
> The menu toggle button doesn't look centered to the home icon / board title
I have same position like `files` and I know its not perfect but its have same look

artemanufrij:
@juliushaertl 
* 'new stack' has same margin/padding as before
* fixed: toggle in desktop mode

js/app/Run.js
```diff
@@ -52,4 +52,24 @@ app.run(function ($document, $rootScope, $transitions, BoardService) {
...
+    // App sidebar on mobile
+    var snapper = new Snap({
+        element: document.getElementById('app-content'),
+        disable: 'right',
+        maxPosition: 250,
+        minDragDistance: 100
+    });
+
+    $('#app-navigation-toggle').click(function(){
+        if($(window).width() > 768) {
```
pixelipo: remove this `if`, since the toggle should always be visible - then you won't need `!important` in CSS

templates/main.php
```diff
@@ -62,7 +62,8 @@
 		<?php print_unescaped($this->inc('part.navigation')); ?>
 		<?php /* print_unescaped($this->inc('part.settings')); */ ?>
 	</div>
-	<div id="app-content" ng-class="{ 'details-visible': sidebar.show }" ui-view>
+	<div id="app-content" ng-class="{ 'details-visible': sidebar.show }">
+		<div ui-view></div>
```
pixelipo: why do you need additional `<div>` here?

templates/part.board.mainView.php
```diff
@@ -5,12 +5,12 @@
 		<p>{{ statusservice.text }}</p></div>
 </div>
 
-<div id="board-header">
-	<div id="app-navigation-toggle" class="icon-menu" ng-click="navibar.show=!navibar.show"></div>
-	<a class="crumb" href="#" title="<?php p($l->t('All Boards')); ?>">
-		<i class="icon icon-home"></i>
-		<span class="hidden-visually"><?php p($l->t('All Boards')); ?></span>
-	</a>
+<div id="controls">
```
pixelipo: please revert to previous version of this code. It was decided some time ago that it's a good approach. CHanges here are not needed to make toggle work.

I do agree with renaming `#board-header` to `#controls`

templates/part.boardlist.php
```diff
@@ -1,9 +1,12 @@
-<div id="board-header">
-	<div id="app-navigation-toggle" class="icon-menu" ng-click="navibar.show=!navibar.show"></div>
-	<a class="crumb" href="#" title="<?php p($l->t('All Boards')); ?>">
-		<i class="icon icon-home"></i>
-		<span class="hidden-visually"><?php p($l->t('All Boards')); ?></span>
-	</a>
+<div id="controls">
+	<div class="breadcrumb">
```
pixelipo: same as above.

js/app/Run.js
artemanufrij: it for different behavior. on mobile devices app-content will be moved to the right. on desktop app-navigation will be hide.

templates/main.php
artemanufrij: otherwise it doesn't work with core-functions

templates/part.board.mainView.php
artemanufrij: no, but it's same code like `files`

pixelipo: "same code as files" I remember that discussion from a different PR :wink: files also has a wrong. This is what is causing issue with positioning of icons mentioned by @juliushaertl

templates/main.php
pixelipo: which function exactly? I tested locally and didn't find any errors.

js/app/Run.js
pixelipo: does it even work, @artemanufrij ? Look at the next line `toggle('hidde')` is not correct :wink:

templates/main.php
artemanufrij: the toggle button will be created by `js.js` from core (see line 1569) automatically. it works only if `ui-view` is in a separate div instead in `app-content`

js/app/Run.js
artemanufrij: it works fine....

templates/main.php
pixelipo: Why do we create a DOM element using JS if it has to be present in 100% of cases? Seems like a waste of resources. cc @juliushaertl

juliushaertl: @pixelipo This is part of the logic from core, so that apps don't need to care about adding this sidebar slide functionality. But since we use our own js code to toggle those of course we could also  just keep the existing structure and add the toggle button to the template.

templates/part.board.mainView.php
juliushaertl: I remember we discussed that some times ago, yes, although I'm not sure what exactly the reason was for that. :see_no_evil: Let's use our fixed approach, we can always go to the way files does it once it is implemented properly there.

css/style.css
```diff
@@ -173,7 +173,7 @@ button.button-inline:hover {
 
 #app-navigation-toggle {
 	width: 45px;
```
pixelipo: this should be 44 as well

js/app/Run.js
```diff
@@ -63,9 +63,9 @@ app.run(function ($document, $rootScope, $transitions, BoardService) {
 
     $('#app-navigation-toggle').click(function(){
         if($(window).width() > 768) {
-            $('#app-navigation').toggle('hidde');
+            $('#app-navigation').toggle('hidden');
```
pixelipo: :+1: that's what I was talking about :)

Issue #136, opened 2017-05-17
We should show board name in the window title


artemanufrij:
![screenshot_20170517_214453](https://cloud.githubusercontent.com/assets/16517095/26173074/d6b3f53e-3b4a-11e7-8d68-fab21060fa41.png)

artemanufrij:
fixed in #191

PR #191, opened 2017-06-19
Set board title as browser title

juliushaertl:
Fix for #136

### DIFF
--- a/js/controller/BoardController.js
+++ b/js/controller/BoardController.js
@@ -45,6 +45,13 @@ app.controller('BoardController', function ($rootScope, $scope, $stateParams, St
 
 	$scope.board = BoardService.getCurrent();
 	StackService.clear(); //FIXME: Is this still needed?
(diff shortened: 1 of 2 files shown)


js/controller/BoardController.js
```diff
@@ -45,6 +45,13 @@ app.controller('BoardController', function ($rootScope, $scope, $stateParams, St
 
 	$scope.board = BoardService.getCurrent();
 	StackService.clear(); //FIXME: Is this still needed?
+	$scope.setPageTitle = function() {
+		if(BoardService.getCurrent()) {
+			document.title = BoardService.getCurrent().title + " - Deck - " + oc_defaults.name;
```
pixelipo: It would look a bit better with a different divider between board title and app title, such as `·` or `|`

juliushaertl: Yes, looks better. :+1:

PR #193, opened 2017-06-19
Fix dependency for versioneye

juliushaertl:


juliushaertl:
Let's remove versioneye ... more annoying than helpful for now :see_no_evil:

Issue #84, opened 2017-03-29
Stack should get always 100% height


stefan-niedermann:
Otherwise it looks a little bit glitchy:

![bildschirmfoto vom 2017-03-29 22 23 09](https://cloud.githubusercontent.com/assets/4741199/24474877/602a25f2-14ce-11e7-8e23-7d32bee2c272.png)

(i mean the gray background)

juliushaertl:
Thanks for the idea, we should give it a try to see how it looks.

NinaNC:
not sure when I have time but I'll give it a try :)

juliushaertl:
@NinaNC Great to hear. :+1: If you have any questions feel free to ask.

x9t9:
This very easily tested , and would also be resolved ( similar to what suggested in #82) with 'display:flex' on `#innerBoard` .  
That said , I do not believe it would make it look less glitchy.
In fact , IMHO , it would even look more chaotic as layout : 
![2017-05-10_070514](https://cloud.githubusercontent.com/assets/23067248/25876334/1a0dbcea-354f-11e7-9794-cf9a0bb789d2.jpg)

Further more , ( again, IMHO ) Should swimlanes be implemented as per #32  ( and they should .. ) , this would be more or less a help to ease the pain with the unavoidable horizontal division of the lanes .

stefan-niedermann:
It looks just a little chaotic because of the huge primary cta-buttons (+ Buttons) at the bottom. Maybe (in a seperate issue) they could move in the column-header as a simple + Sign without blue background?

keunes:
Agree with @stefan-niedermann, but not sure this 'full height' would provide me with any actual (visual) improvements IMHO :)

juliushaertl:
Flexbox is introduced in #352 and we have 100% height at least when dragging icons.

Issue #82, opened 2017-03-29
Stacks should have flexible width if enough space


stefan-niedermann:
Currently each stack has a fixed width. This is a bad use of the given space.
If you have only one stack, 80% of your screen is unused while you have little space to view your cards.
If you have many stacks, you have to scroll vertical, which is an unnatural gesture and causes some layout issues (e. g. a dragged card has another position on the screen than the mouse).

Proposal: The stacks should stretch to maximum width when possible (this means, one stack: 100% width, two stacks: 50% width, ...) of course, when you have very much stacks (and "much" is depending on the screen width), each stack should have a minimum width, so they do not get too small.

juliushaertl:
Makes sense to me. I think that is a good starter issue using css flexbox for implementation.

stefan-niedermann:
Only one question left: where to display the input field and submit button for new stacks, if the present stacks consume all the horizontal space? maybe a plus button in the top right corner of the most right stack?

juliushaertl:
@stefan-niedermann The add stack input has been moved to the top with #95

x9t9:
I  Believe that only putting `display:flex `on `#innerBoard ` and maybe `.stack` with `width:100%` would already achieve that .  It would potentially create other problems ( like stack delete/edit icons, which IMHO are misplaced ) , but should work for this specific issue .

On the same note : I believe all the CSS should go flex now, which will also allow for support on mobile ( for display mainly , not editing )

artemanufrij:
@x9t9 and `min-width: XXpx;`

juliushaertl:
@x9t9 It would be great if you could open a pull request. We can figure out smaller issues during the review process. 

> On the same note : I believe all the CSS should go flex now, which will also allow for support on mobile ( for display mainly , not editing )

:+1: For moving to flex. The code here was mostly written before I discovered that it exists. :wink:

pixelipo:
+1 for flex, but just FYI, now there is Aldo `grid`:
https://www.mozilla.org/en-US/developer/css-grid/

x9t9:
@pixelipo - Right ,Honestly I never worked with `grid`, mainly because comparing http://caniuse.com#search=flex and http://caniuse.com#search=css-grid) reveals some support issues for now .

But the way I see it , they are not challenging , but more completing each other. grid is more about the big picture layout where flex has advantages on the positioning of children elements . Which means that one can start with the more-supported flex, and later add grid if needed .

[...]

PR #199, opened 2017-06-20
Simplify board view

juliushaertl:
- Remove background from stacks
- Make stack height match the largest one #84
- ~Fixed stack headers, so it is clear where you are when scrolling~
- Resize stack width between 320px and 380px if there is enough space #82 

Feedback is very welcome. I'm not sure if the resizing of the stack width is a good way or if we should just go for a fixed width there.

2017-06-20-213318_1152x862_scrot

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -127,7 +127,7 @@ button.button-inline:hover {
 	top: 44px;
 }
 
-#board #innerBoard {
+#innerBoard {
 	padding: 10px;
 }
 
@@ -251,9 +251,7 @@ button.button-inline:hover {
 
 .stack {
 	width: 100%;
-	margin-right: 10px;
 	vertical-align: top;
-	background-color: #f8f8f8;
 }
 
 .stack h2 {
@@ -981,10 +979,22 @@ button.button-inline:hover {
 		display: none;
 	}
 
+	#innerBoard {
+		display: flex;
+	}
 	.stack {
 		width: 320px;
+		min-width: 320px;
 		display: inline-block;
+		border-right: 1px solid #eee;
+		margin-right: 10px;
+		padding-right: 10px;
 	}
+
+	.stack:last-child {
+		border-right: 1px solid transparent;
+	}
+
 	.stack h2 button {
 		display: none;
 	}
--- a/templates/part.board.mainView.php
+++ b/templates/part.board.mainView.php
@@ -45,8 +45,7 @@
 						ng-if="!s.status.editStack"
                         ng-click="stackservice.delete(s.id)"></button>
 			</h2>
-			<ul data-as-sortable="sortOptions" is-disabled="!boardservice.canEdit() || filter==='archive'" data-ng-model="s.cards"
-				style="min-height: 40px;">
+			<ul data-as-sortable="sortOptions" is-disabled="!boardservice.canEdit() || filter==='archive'" data-ng-model="s.cards" class="card-list">
 				<li class="card as-sortable-item"
 					ng-repeat="c in s.cards"
 					data-as-sortable-item


artemanufrij:
Can't delete a stack. Delete-icon is no more visible on mouse over.

css/style.css
```diff
@@ -286,6 +299,10 @@ button.button-inline:hover {
 	overflow: hidden;
 	text-overflow: ellipsis;
 }
+.stack ul.card-list {
```
pixelipo: no need for both `ul` and `.card-list` - you can target this element specifically with `.stack > ul` or simply `.card-list`. Best not to have too many  queries in CSS

css/style.css
```diff
@@ -252,8 +253,16 @@ button.button-inline:hover {
 .stack {
 	width: 100%;
 	margin-right: 10px;
+	padding-right:10px;
 	vertical-align: top;
-	background-color: #f8f8f8;
+	background-color: #ffffff;
+	border-right: 1px solid #eee;
+	position: relative;
```
pixelipo: not needed

css/style.css
```diff
@@ -125,10 +125,11 @@ button.button-inline:hover {
 	width: 100%;
 	bottom: 0px;
 	top: 44px;
+	margin-top: 10px;
 }
 
-#board #innerBoard {
-	padding: 10px;
+#innerBoard {
+	padding: 0 10px 20px 10px;
```
pixelipo: why complicate padding? You essentially just move top padding over to margin-top for `#board`.

Edit: in fact, you could remove the whole declaration by just giving `padding: 10px` to `#board` element

css/style.css
```diff
@@ -125,10 +125,11 @@ button.button-inline:hover {
 	width: 100%;
 	bottom: 0px;
 	top: 44px;
+	margin-top: 10px;
```
pixelipo: margin for a `fixed` element? Why not just modify `top` value?

css/style.css
```diff
@@ -264,6 +273,10 @@ button.button-inline:hover {
 	overflow: hidden;
 	display: flex;
 	min-height: 40px;
+	position: fixed;
```
pixelipo: Why fixed position? It just creates problems:

1. image

2. you have to give "imaginary" `margin-top` to `.card-list` (line 303). Everything seems to work just as fine without those 3 declarations
3. mobile issue mentioned by @artemanufrij 

It works just fine if you remove those 3 lines.

css/style.css
```diff
@@ -441,7 +459,7 @@ button.button-inline:hover {
 .card.create {
 	text-align: center;
 	padding: 10px;
-	margin: 10px;
+	margin: 10px 10px 0 10px;
```
pixelipo: is there any particular reason for removing bottom margin on this button?

css/style.css
```diff
@@ -972,9 +990,15 @@ button.button-inline:hover {
 		display: none;
 	}
 
+	#innerBoard {
+		display: flex;
+	}
+
 	.stack {
 		width: 320px;
+		min-width: 320px;
```
pixelipo: this doesn't didn't you mention resizing from 200px? Left like this, it has 0 effect because of `width: 320px`

css/style.css
```diff
@@ -972,9 +990,15 @@ button.button-inline:hover {
...
 		display: inline-block;
+		position: relative;
```
pixelipo: not required - since no positioning values (e.g. `top`) are applied to this class.

css/style.css
```diff
@@ -252,8 +253,16 @@ button.button-inline:hover {
 .stack {
 	width: 100%;
 	margin-right: 10px;
+	padding-right:10px;
 	vertical-align: top;
-	background-color: #f8f8f8;
+	background-color: #ffffff;
+	border-right: 1px solid #eee;
```
pixelipo: border should not be visible on mobile

css/style.css
```diff
@@ -252,8 +253,16 @@ button.button-inline:hover {
 .stack {
 	width: 100%;
 	margin-right: 10px;
+	padding-right:10px;
 	vertical-align: top;
-	background-color: #f8f8f8;
+	background-color: #ffffff;
```
pixelipo: `.stack` doesn't need background-color if it's white, right?

css/style.css
juliushaertl: The idea behind that was to keep the card title always on top even when scrolling though larger task lists. But I guess I'll remove this for now and try to add this in a separate PR. I guess it will also require some javascript magic.

pixelipo: Ah, that was a good idea, but yeah, probably some JS will be needed.
I'm more worried about how it would look though. Fixed titles usually require some extra style to distinguish. from the rest of the DOM

artemanufrij:
fix right margin in mobile view:
screenshot_20170623_215743

juliushaertl:
@pixelipo Updated and removed the fixed header stuff. 

I'm not sure if the dynamic sizing proposed in #82 of stacks makes much sense. I don't see much value, since it does only depend on the number of stacks not the browser width. What do you think, @artemanufrij @pixelipo ?

artemanufrij:
I don't like think about dynamic size. I realy like current style (this PR). I vote for fix width for stacks :+1: !

pixelipo:
I agree with Artem. Only a single element of the card would benefit from dynamic width - card title. Everything else takes up already existing space. It would look real bad on large screens

jancborchardt:
Looks much nicer and lighter without the background, good job! :)

PR #214, opened 2017-06-23
Fix icon2

pixelipo:
@juliushaertl I've added viewBox as this was suggested by @jancborchardt as a must-have for app icons.

I've also modified deck.svg to be identical (but white) version of deck-dark.svg

I didn't run scour optimization script - which would basically only remove whitespaces (~40 bytes per file).

Made a PR to your branch so that you can decide if you want these changes or not :wink: 


pixelipo:
to be merged to branch https://github.com/nextcloud/deck/tree/fix-icon

juliushaertl:
Thanks @pixelipo I've added your commit to #213 

Feel free to commit directly to the original branch if it is that related next time ;)

PR #215, opened 2017-06-23
polish sidebar

artemanufrij:

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -918,7 +918,7 @@ button.button-inline:hover {
 	width: 100%;
 	padding: 4px;
 	border: none;
-	border-radius: 0px 0px 3px 3px;
+	border-radius: 0px 0px 5px 5px;
 }
 
 #board-detail-labels ul li a.icon {
@@ -979,7 +979,7 @@ button.button-inline:hover {
 	padding: 0;
 	float: left !important;
 	display: block;
-	border-radius: 0px 0px 3px 3px !important;
+	border-radius: 0px 0px 5px 5px !important;
 }
 
 .ui-select-match-item .select-label {
--- a/css/style.css
+++ b/css/style.css
@@ -979,6 +979,7 @@ button.button-inline:hover {
 	padding: 0;
 	float: left !important;
 	display: block;
+	border-radius: 0px 0px 3px 3px !important;
 }
 
 .ui-select-match-item .select-label {
--- a/css/style.css
+++ b/css/style.css
@@ -73,7 +73,7 @@ button.button-inline:hover {
 .app-navigation-entry-utils-menu-share {
 	display: flex !important;
 	padding: 14px;
-	opacity: 0.4;
+	opacity: 0.3;
 }
 
 #app-navigation .app-navigation-entry-edit {
@@ -293,7 +293,7 @@ button.button-inline:hover {
 .stack h2 button {
 	margin-left: auto;
 	display: flex;
-	opacity: 0.25;
+	opacity: 0.3;
 	padding-right: 0;
 }
 
@@ -328,7 +328,7 @@ button.button-inline:hover {
 }
 
 .card .card-controls .card-options {
-	opacity: 0.25;
+	opacity: 0.3;
 	position: absolute;
 	right: 10px;
 	top: 8px;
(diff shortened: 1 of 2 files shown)


artemanufrij:
Current design:
screenshot_20170624_003433

artemanufrij:
last version:
screenshot_20170626_215550

templates/part.card.php
```diff
@@ -28,54 +28,45 @@
 		<?php p($l->t('by')); ?>
 		<span>{{ cardservice.getCurrent().owner.displayname }}</span>
 	</div>
-
+	<h3 style="padding-top: 0px; margin-top: 0px;">
```
juliushaertl: No inline styles please ;)

templates/part.card.php
```diff
@@ -28,54 +28,45 @@
...
-		</ui-select>
-	</div>//-->
-
+	
 	<div id="card-description">
 		<h3>
-			<?php p($l->t('Description')); ?>
-			<a href="https://github.com/nextcloud/deck/wiki/Markdown-Help" target="_blank" class="icon-help" title="<?php p($l->t('Formatting help')); ?>"></a>
-			<span class="save-indicator"><?php p($l->t('Saved')); ?></span>
+			<div>
```
juliushaertl: What is the need of two wrapping divs here? h3 is already a block element that could be used.

templates/part.card.php
```diff
@@ -89,43 +80,5 @@
 			<div class="placeholder"
 				 ng-if="!cardservice.getCurrent().description"><?php p($l->t('Add a card description…')); ?></div>
 		</div>
-
 	</div>
-
-
```
juliushaertl: :see_no_evil: :+1:

css/style.css
```diff
@@ -926,6 +954,9 @@ button.button-inline:hover {
 	margin-bottom: 0;
 }
 
+.tabsContainer {
+	margin-top: 15px;
+}
```
juliushaertl: missing newline.

css/style.css
```diff
@@ -549,27 +557,51 @@ button.button-inline:hover {
...
+#card-meta .duedate .datepicker-input {
+	width: 85px;
+}
+
+#card-meta .duedate .icon {
+	opacity: 0.3;
+}
+
+#card-meta .duedate .icon:hover {
+	opacity: 0.7;
```
juliushaertl: Not really needed, since the hover effect is done by the icon-delete class.

css/style.css
```diff
@@ -549,27 +557,51 @@ button.button-inline:hover {
...
+	width: 50px;
+	margin: 3px;
+}
+
+#card-meta .duedate .datepicker-input {
+	width: 85px;
+}
+
+#card-meta .duedate .icon {
+	opacity: 0.3;
```
juliushaertl: I would go to 0.5 opacity, since it should be default for all icons on nextcloud: https://github.com/nextcloud/server/issues/5157

templates/part.card.php
artemanufrij: for `display: flex`.

css/style.css
artemanufrij: but not by edit

juliushaertl: There is just the delete button :wink:

https://github.com/nextcloud/deck/pull/215/files#diff-a175750cce89c445e0a34d25823924c2R56

templates/part.card.php
juliushaertl: But we can just apply the flex styles to `#card-meta h3` or am I missing something?

css/style.css
artemanufrij: but it doesn't change opacity. only color. for buttons we use always a different opacity on mouse over.

templates/part.card.php
artemanufrij: `display: flex` changes some margin/padding behavior. It looks some different in compare to standard h3

css/style.css
juliushaertl: Ah i see. Ok fine then.

artemanufrij:
@juliushaertl 
like this:
screenshot_20170627_210514

Issue #160, opened 2017-05-25
We should remove unused css code


artemanufrij:
For more transparency

artemanufrij:
request is already merged

pixelipo:
@artemanufrij stll some work on this task. I will make more PRs under the same issue, ok?

artemanufrij:
@pixelipo: Sure!

juliushaertl:
Let's remove that from the 0.2 milestone, since this is something ongoing and nothing that is really required for now. :wink:

artemanufrij:
fixed in #219

PR #219, opened 2017-06-27
Move to SCSS

juliushaertl:
I'll wait for #215 and try to restructure the css file then to make use of all the great features. :tada: 

Fallback style.css will be generated when running make so it will keep working when using Nextcloud 11.


pixelipo:
style.css is created but for some reason not loaded when I run the app...

Plus it doesn't keep line break before `}` closing brackets.

P.S. I can take care of converting to 'proper' scss, if you don't mind.

juliushaertl:
> style.css is created but for some reason not loaded when I run the app...

For Nextcloud >12 it will generate the file automatically and load it from something like `/index.php/css/deck/37d007a56d816107ce5b52c10342db37-style.css`

> Plus it doesn't keep line break before } closing brackets.

I guess that can be adjusted by using --output-style extended in the Makefile. But we should use compressed there anyway so we have a small file in the end.

> P.S. I can take care of converting to 'proper' scss, if you don't mind.

That would be great.

juliushaertl:
> icon is not visible:

That is basically because the Nextclouds scss functionality only rewrites urls that are using quotes like url('../../iconfile') and not url(../../iconfile). We should add quotes everywhere for now. Maybe @pixelipo can do this when he does the migration to scss. I'll create a PR in nextcloud/server to fix this later as well.

pixelipo:
No problem.

BTW, I have already grouped all styles, tomorrow I'll extract colors into variables and push here. That should be enough for first version of scss.

artemanufrij:
@juliushaertl I added quotes in pr #220

pixelipo:
@juliushaertl @artemanufrij please test this. I have converted to SCSS structure and added color variables. Also cleaned up few unused styles and simplified color palette (we literally had "50 shades of gray" :rofl: )

artemanufrij:
@pixelipo I will test today afternoon.

pixelipo:
We can also close #160 once we merge this. There is still some unused CSS left, but it's not too bad and it can be cleaned up when needed.

juliushaertl:
Pushed another commit to small issues with the board list. Otherwise looks great :+1: 

Awesome work @pixelipo. Thanks for that.

css/style.scss
```diff
@@ -755,10 +755,11 @@ input.input-inline {
 
 	thead td {
 		opacity: .5;
+		padding: 15px;
 	}
 
 	td {
-		padding: 0 10px;
+		padding: 3px 10px;
```
pixelipo: @juliushaertl please revert this line to `padding: 0 10px;`

If you think rows must be 50px high, then you can increase this to 25px: `#boardlist .icon { padding: 25px; }`

The difference between these 2 solutions is that with the latter, buttons fill full height of the row.

P.S. In my opinion, 44px is enough for row height anyway.

juliushaertl: Thanks. Fixed.

PR #220, opened 2017-06-28
added quotes for urls into css file

artemanufrij:


artemanufrij:
will be reopening after moving to scss

PR #226, opened 2017-07-04
changed space between cards

artemanufrij:

### DIFF
--- a/css/style.scss
+++ b/css/style.scss
@@ -312,7 +312,7 @@ input.input-inline {
 
 .card {
 	background-color: $color-main-background;
-	margin: 10px;
+	margin: 10px 10px 20px 10px;
 	white-space: normal;
 	position: relative;
 	opacity: 1.0;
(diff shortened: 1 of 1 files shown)


artemanufrij:
**before:**
cards_before

**after:**
cards_after

juliushaertl:
@artemanufrij Yes that looks better than before. :+1:

pixelipo:
I'm not so sure that we need that much space between cards - it will result in one card less per screen every 7 cards...

pixelipo:
Merged without review, @juliushaertl ?

juliushaertl:
Was counting my :+1: as review :wink: 

But I also recogniced that on decks with many cards this is quite much unused space that would be better used if we show more cards. Maybe we should revert this. :see_no_evil:

artemanufrij:
@juliushaertl @pixelipo what do you talk about :smiley_cat: ? At home 21:9 34" at work: 4k :smile:

Issue #229, opened 2017-07-05
Small layout issue if no due date is set


poVoq:
If there is no due date set for a card, the button for adding it looks like it is cut off on the side. This is likely due to the fact that if a due date gets added, this kind of overlapps with the button and extends it.
Not really a show stopper, but looks odd.

artemanufrij:
@poVoq you can check it under #230

poVoq:
Great, thanks!

PR #230, opened 2017-07-05
fix for due date control

artemanufrij:


artemanufrij:
fix for #229

artemanufrij:
before:
screenshot_20170705_210234

after:
screenshot_20170705_210137

Issue #235, opened 2017-07-06
 Impossible to drop the card in an empty column


oparoz:
### Steps to reproduce

1. Move a card from its current location to and empty column

### Expected behaviour

The dotted drop zone should show up

### Actual behaviour

Impossible to drop the card

On 0.2.1

juliushaertl:
@pixelipo Can you have a look, this might be caused by the migration to SCSS.

juliushaertl:
Fixed with #238

PR #238, opened 2017-07-06
Add min-height to card list and merge duplicate rules for .card

juliushaertl:
fix for #235 

also merging two .card declarations into one


juliushaertl:
> The problem is not min height, but the fact that ul.card-list is the target for mouse cursor. Proper solution would be to use the entire div.stack as mouse target instead.

The problem is that the mouse target will be automatically assigned to the item that contains the list of sortable items (in our case the ul element). I could not find a solution for this right now.

If you want to have a look, feel free. Otherwise I'd like to merge this for now since it fixes a bug. :wink:

juliushaertl:
>  I wonder what if we moved the "Add Card" button/form into the list (as always last element)

Let me check that later today. That sounds like it should work :wink:

jancborchardt:
Any update @juliushaertl? Can help reviewing for sure :)

pixelipo:
AFAIC, we can merge as-is and fix #239 in a separate PR

juliushaertl:
> I wonder what if we moved the "Add Card" button/form into the list (as always last element)

That did not went well, becasue ng-sortable will then add a placeholder to drag to after the add button.

I did some margin/padding magic to make the drag-target container also include the header/create button. Please review @jancborchardt @pixelipo

jancborchardt:
Which version was this fixed in? 0.2.2 I guess?  :)

jancborchardt:
Asking cause I encounter that issue with the version of the Deck app at Portknox – cc @Gomez :)

juliushaertl:
@jancborchardt Yes, it is in 0.2.2

jacestk:
Mhh in 0.2.3 I stil can't move card into empty stack!

juliushaertl:
@jacestk Works fine here with 0.2.3 and Nextcloud 12 on both firefox and chrome. Please try a hard reload to ensure that no files are cached (Ctrl +  F5). If it still doesn't work please open a new issue including details about your browser.

jacestk:
Nope don't work on IE11, FireFOX 55.0.3 (64 bits), Chrome  61.0.3163.100 ...

PR #258, opened 2017-07-14
Fix drag-zone when sorting cards

juliushaertl:
Follow up of #238 

> I wonder what if we moved the "Add Card" button/form into the list (as always last element)

That did not went well, becasue ng-sortable will then add a placeholder to drag to after the add button.

I did some margin/padding magic to make the drag-target container also include the header/create button. Please review @jancborchardt @pixelipo

### DIFF
--- a/css/style.scss
+++ b/css/style.scss
@@ -310,7 +310,9 @@ input.input-inline {
 	}
 
 	.card-list {
-		min-height: 40px;
+		padding-bottom: 100px;
+		padding-top: 40px;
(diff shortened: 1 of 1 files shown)


pixelipo:
Works as expected @juliushaertl 

I'm a bit afraid the padding "hack" might bite us in the ass later, but let's worry about it later :laughing:

juliushaertl:
Failing tests were unrelated, also broken on master.

Issue #83, opened 2017-03-29
Provide public REST-API (and or documentation)... [$25]


stefan-niedermann:
... so one can write native applications that are in sync with the web interface :)

srbaker:
Heya, is someone already working on this?

If not, I'll do it. This is my first work on Nextcloud stuff, though, so I'll benefit from some hand-holding.

Thanks!

juliushaertl:
@srbaker Sorry for the delay. Sure, It would be awesome if you would jump in and help us improve deck.

It would require implementing some extra API controllers similar to what the news app does: https://github.com/nextcloud/news/blob/master/lib/Controller/FeedApiController.php 

Although it would be great to have some concept before starting the implementation. You can have a look at the v2 API proposal of the news app here as well: https://github.com/nextcloud/news/blob/master/docs/externalapi/External-Api.md

If you have any further questions feel free to ask or ping me on irc.

srbaker:
Perfect, thanks. I'll take a look at this on the weekend. I'm very excited to be able to replace Trello with something that I can share with my family. And I really want there to be a desktop and Android app as well. :)

stefan-niedermann:
> Android app

And that's what i am interested in to develop ;-)

juliushaertl:
@srbaker I've invited you to be a repo contributer, so you can directly push branches when developing.

srbaker:
Perfect, thanks! I didn't get as far as I wanted this weekend, but I have several hours in the evenings this week. I'll push up the in-progress stuff once I get the first simplest endpoint working. 

Cheers!

moritz31:
@srbaker @juliushaertl 
If we have a REST-API it shouldn't be that complicated to add an iOs/Android application.
Have done similar to some home automation stuff

juliushaertl:
@srbaker Any news from your api implementation? Is there anything you need help with?

srbaker:
Heya, sorry for the radio silence. Got a bit slowed down, but I have vacation time I'm planning to wrap up my work.

juliushaertl:
@srbaker No problem :wink: Take your time.

srbaker:
Alright, it's actually moving. Lovely vacation project.

https://github.com/srbaker/deck/tree/add-json-api

I'd like help with the following:

1. I'm really struggling by not doing this test-first, because that's my habit. I'm using `curl` manually as a temporary solution: When I run `make test` from my `deck` checkout, I get: 
```
mkdir -p build/
[...]

Issue #30, opened 2017-02-03
Import from trello


brantje:
Hi there!
Awesome idea, currently i use trello.
It would be a pain to migrate by hand, however, trello offers a [API](https://developers.trello.com/advanced-reference).
Would it be possible to write some kind of importer?

juliushaertl:
That would be definitely super nice to have also for other apps, but it is no priority for me now. 

We could also offer a public API for deck so that people can build simple import scripts for various other tools similar to this one: https://github.com/matueranet/kanboard-import-trello

srbaker:
This is what I need the API for, so I can import from Trello. Happy to do the work, as I mentioned in #83.

weipah:
I +1 this feature request. It would really ease up the process to make the switch from cloud-based to local installation.

Trello also supports JSON exports in their free version.

I guess using this for a one-time import would be much easier than deep-diving into their api.

juliushaertl:
@weipah The API was basically meant to be for Deck so users can script their import from various platforms. 

I see the sense of importing trello data directly from the webinterface, but I will not have time to implement this in the near future. Several other features will be focused before. But as always, if anyone wants to contribute, pullrequests are always welcome.

krombel:
can this be reopened again? I would like to have that feature as well.

juliushaertl:
This would be a nice feature, however currently there are no plans to implement this in the near future. In general I would be a lot more happy if importing from various 3rd party services and software would be done using the API, so we keep the deck code clean from other dependencies. Thus I will keep this ticket closed for now. If somebody wants to implement this feature nevertheless we are happy to assist and help out.

If you wish to have this feature implemented by the Nextcloud GmbH there is the option for consulting work on top of your Nextcloud Enterprise subscription to get your features implemented.

BloodyIron:
Let's get this reopened please. I'm for it!

martinaware:
I'd like it to reopened as well.

brokencode64:
Would love to see this re-opened. Currently this is the only thing keeping me from migrating to Deck.

putt1ck:
@brokencode64 we definitely have this on our radar but will be via the API using Trello export JSON format (because last I checked that was only free export format).

juliushaertl:
> If somebody wants to implement this feature nevertheless we are happy to assist and help out.

This still counts even if the ticket is closed. 

**A general note:**
[...]

PR #271, opened 2017-07-29
Add JSON API

srbaker:
This is the first run at an API, as discussed in #83. It is also required for #30.


juliushaertl:
@srbaker Sorry for the long time without a reply. As you may have recognized, it has been silent around deck in the last weeks. I'll make sure to review this during the weekend so we can go on with that.

lib/Controller/StackApiController.php
```diff
@@ -0,0 +1,97 @@
...
+     * @CORS
+     * @NoCSRFRequired
+     *
+     * @params $title
+     * @params $order
+     *
+     * Create a stack with the specified title and order.
+     */
+    public function create($boardId, $title, $order) {
+        // this throws a StatusException that needs to be caught and handled
```
juliushaertl: We catch StatusExceptions in the "SharingMiddleware" (although the Sharing is not part of that anymore, so it might need some rename). https://github.com/nextcloud/deck/blob/15e5a432b631a9eb8a464f3d6082702e6cb348cc/lib/Middleware/SharingMiddleware.php#L61

You can either add some logic to return proper error responses there, or exclude any exceptions from ApiControllers there and add the handling to the controller.

lib/Controller/BoardApiController.php
```diff
@@ -0,0 +1,146 @@
...
+     * @NoCSRFRequired
+     *
+     * @params $id
+     *
+     * Return the board specified by $id.
+     */
+    public function get($id) {
+        $board = $this->service->find($id);
+
+        // FIXME: this should probably 404 if the board has been deleted
```
juliushaertl: It should be a 200 if it got deleted and a 404 if the entity for $id is not found. Otherwise you cannot separate between a successful deletion and a failing one, because the entity does not exist.

lib/Controller/BoardApiController.php
```diff
@@ -0,0 +1,146 @@
...
+     * Undo the deletion of the board specified by $id.
+     */
+    public function undoDelete($id) {
+        $board = $this->service->find($id);
+        $this->service->deleteUndo($id);
+
+        return new DataResponse($board);
+    }
+
+    // this is taken from BoardController, but it's not ideal
```
juliushaertl: You can move the method to the BoardService class, then we can use it from both controllers.

See: https://github.com/nextcloud/deck/blob/2c63bfb4b8a7d5b0c22dc1d753e56d1dd08b61a9/lib/Controller/BoardController.php#L53

appinfo/routes.php
```diff
@@ -62,5 +62,18 @@
 		['name' => 'label#update', 'url' => '/labels/{labelId}', 'verb' => 'PUT'],
 		['name' => 'label#delete', 'url' => '/labels/{labelId}', 'verb' => 'DELETE'],
 
+        // api
```
juliushaertl: indentation seems wrong here :wink:

srbaker:
Thanks! I've also been a bit distracted by other things; I'll plan to address these this week, and finish up the last bits. I'm hoping to use this to get a mobile app running before the end of the year. Deck is starting to become a rather important part of my personal workflows. :)

juliushaertl:
@srbaker Sure no need to hurry. I plan to do a 0.3.0 release in like 1 or 2 months, so we have some time if we want to get this in there.

robertauer:
Are you guys still working on this? Would be really great to sync Deck with external apps!

srbaker:
Heya, I haven't had the time I was hoping to. But I've got some time schedule to get this through to completion soon!

Thanks for your interest!

artemanufrij:
I'm also waiting for implementation of JSON API. I plan to develop a native application for elementary OS

juliushaertl:
@srbaker That is awesome to hear. I think the JSON API will be a huge benefit for the deck app. Thank you already for digging into it :+1:

srbaker:
@juliushaertl It is encouraging that I can drag out a contribution this long, and still have it be valued. I really want to see this, but sometimes real life gets in the way, you know? I've got some holidays coming up and a good set of tasks to work through. Cheers!

srbaker:
Okay! I am back on this, after some delay. :)

Do you know if there is a linter for PHP that can tell me when I've violated the nextcloud coding style? I'd really like to just set up my editor (emacs) to do the right thing for me, rather than thinking about it. Thanks!

srbaker:
I _think_ this solves the open issues. Please, comment on the latest. I'm going to be giving this more attention this week and next, so please comment if you have time.

lib/Controller/BoardApiController.php
```diff
@@ -0,0 +1,136 @@
...
+	 * @param IRequest $request
+	 * @param IUserManager $userManager
+	 * @param IGroupManager $groupManager
+	 * @param BoardService $service
+	 * @param $userId
+	 */
+	public function __construct($appName, IRequest $request, IUserManager $userManager, IGroupManager $groupManager, BoardService $service, $userId) {
+		parent::__construct($appName, $request);
+		$this->service = $service;
+		$this->userId = $userId;
```
juliushaertl: No need for the user/group manager anymore.

lib/Controller/BoardApiController.php
```diff
@@ -0,0 +1,136 @@
...
+		$this->userManager = $userManager;
+		$this->groupManager = $groupManager;
+		$this->userInfo = $this->service->getBoardPrerequisites();
```
juliushaertl: We can actually move that service to the BoardService now. :wink:

lib/Controller/BoardApiController.php
```diff
@@ -0,0 +1,136 @@
...
+
+	/**
+	 * @NoAdminRequired
+	 * @CORS
+	 * @NoCSRFRequired
+	 *
+	 * Return all of the boards that the current user has access to.
+	 */
+	public function index() {
+		$boards = $this->service->findAll($this->userInfo);
```
juliushaertl: As mentioned above, let's move the user info fetching to the board service and get rid of the parameter here.

juliushaertl:
> Do you know if there is a linter for PHP that can tell me when I've violated the nextcloud coding style? I'd really like to just set up my editor (emacs) to do the right thing for me, rather than thinking about it. Thanks!

I'm not aware of any linter with a config available for the Nextcloud coding style. https://docs.nextcloud.com/server/10/developer_manual/general/codingguidelines.html#php 

Since many developers use PHPStorm, the code style settings are available here: https://github.com/nextcloud/server/blob/master/.idea/codeStyleSettings.xml I'm sorry, I cannot tell anything about emacs, but in general the coding style looks fine on this PR.

srbaker:
Sorry, I missed that, it should have been obvious. Clearly I need to go back through and do a refactor.

srbaker:
I haven't done the commit signing: do I have to go back and amend all of my commits with a signature? Or is there something else I can do?

juliushaertl:
> I haven't done the commit signing: do I have to go back and amend all of my commits with a signature? Or is there something else I can do?

We can also squash those commits when merging, if you are fine with that.

Btw. feel free to join #nextcloud-deck on freenode IRC if you have some quick questions. :wink:

Braintelligence:
Looks like we can expect an API for Deck soon, very excited!

juliushaertl:
Closing in favour of #535 

Thanks again @srbaker for getting this started.

PR #278, opened 2017-08-08
Fix card description placeholder

### LINKED ISSUE #261: "Add a card description" in sidebar has to be clicked twice
In addition to the placeholder of the field "Enter your description here" there is text in it before clicking "Add a card description".

You need to click the field to make the text vanish, but the field doesnt get focused. So you need to click it again. The first text should not be there.

(Also, the better wording is "Add a detailed description …" - things like "Enter here" or "Click here" are always a bit strange.)

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/47262574-add-a-card-description-in-sidebar-has-to-be-clicked-twice?utm_campaign=plugin&utm_content=tracker%2F52758431&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F52758431&utm_medium=issues&utm_source=github).
</bountysource-plugin>

juliushaertl:
fixes #261

### DIFF
--- a/templates/part.card.php
+++ b/templates/part.card.php
@@ -69,7 +69,7 @@
 
 		</h3>
 		<textarea elastic ng-if="status.cardEditDescription"
-				  placeholder="Enter your description here…"
+				  placeholder="<?php p($l->t('Add a card description…')); ?>"
 				  ng-blur="cardUpdate(cardservice.getCurrent())"
 				  ng-model="cardservice.getCurrent().description"
 				  autofocus-on-insert> </textarea>


templates/part.card.php
```diff
@@ -69,7 +69,7 @@
 
 		</h3>
 		<textarea elastic ng-if="status.cardEditDescription"
-				  placeholder="Enter your description here…"
+				  placeholder="<?php p($l->t('Add a card description…')); ?>"
```
jancborchardt: Just a detail, but we use a space between the last word and the ellipsis always. :)

Issue #241, opened 2017-07-09
Click and drag in slide details causes UI issues


TeknoShamin:
### Steps to reproduce
1. Create new board.
2. Create new deck.
3. Create new slide.
4. Enter text into slide.
5. Click on text and drag to the right.
6. Observe how the right hand window appears.

### Expected behaviour
Nothing should happen.

### Actual behaviour
The right hand window pane appears.

### Server configuration
**Operating system**: Linux farnsworth 4.4.0-77-generic #98-Ubuntu SMP Wed Apr 26 08:34:02 UTC 2017 x86_64

**Web server:** nginx/1.10.0 (fpm-fcgi)

**Database:** mysql 10.0.29

**PHP version:** 7.0.15-0ubuntu0.16.04.4
Modules loaded: Core, date, libxml, openssl, pcre, zlib, filter, hash, Reflection, SPL, session, standard, cgi-fcgi, mysqlnd, PDO, xml, apcu, calendar, ctype, curl, dom, mbstring, fileinfo, ftp, gd, gettext, iconv, igbinary, json, exif, mysqli, pdo_mysql, Phar, posix, readline, redis, shmop, SimpleXML, sockets, sysvmsg, sysvsem, sysvshm, tokenizer, wddx, xmlreader, xmlwriter, xsl, zip, Zend OPcache

**Nextcloud version:** 11.0.3 (stable) - 11.0.3.2

**Updated from an older Nextcloud/ownCloud or fresh install:**

**Where did you install Nextcloud from:** 

**Signing status:**
<details>
	<summary>Signing status</summary>

```
[]
```

</details>

"No errors have been found."

**List of activated apps:**
<details>
	<summary>App list</summary>
```
Enabled:
 - activity: 2.4.1
 - activitylog: 0.0.1
 - admin_notifications: 1.0.0
 - announcementcenter: 3.0.0
 - apporder: 0.3.3
 - bookmarks: 0.10.0
 - calendar: 1.5.3
 - comments: 1.1.0
 - contacts: 1.5.3
 - dav: 1.1.1
 - deck: 0.2.1
 - defaultgroup: 0.3.0
 - direct_menu: 0.10.2
 - drawio: 0.8.8
 - external: 1
 - federatedfilesharing: 1.1.1
 - federation: 1.1.1
 - files: 1.6.1
 - files_automatedtagging: 1.1.1
 - files_markdown: 1.0.1
 - files_pdfviewer: 1.0.1
 - files_reader: 1.0.4
[...]

PR #284, opened 2017-08-11
Disable draging of the sidebar

### LINKED ISSUE #262: Use of snap.js in desktop view makes right sidebar disappear/glitch if you select text in the sidebar
If you drag-select text with your cursor from left to right in the sidebar, say from the description, it will prompt snap.js of the left navigation to move the content to the right, leading the right sidebar to disappear.

That's one reason why we don't use snap.js on desktop, only on mobile. ;)

We should probably fix this in core, but as Deck right now is the only app which overrides the mobile handling I'm opening it here. Probably a simple removal of the mechanic of content getting dragged to the right too far would fix it?

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/47263293-use-of-snap-js-in-desktop-view-makes-right-sidebar-disappear-glitch-if-you-select-text-in-the-sidebar?utm_campaign=plugin&utm_content=tracker%2F52758431&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F52758431&utm_medium=issues&utm_source=github).
</bountysource-plugin>

juliushaertl:
Fixes #262 #241

### DIFF
--- a/js/app/Run.js
+++ b/js/app/Run.js
@@ -59,7 +59,7 @@ app.run(function ($document, $rootScope, $transitions, BoardService) {
         element: document.getElementById('app-content'),
         disable: 'right',
         maxPosition: 250,
-        minDragDistance: 100
+        touchToDrag: false
     });
 
     $('#app-navigation-toggle').click(function(){


juliushaertl:
Let's merge this since I want to release a bugfix release, since many users coming up with reporting duplicate bugs for 0.2.1.

PR #301, opened 2017-09-13
Fix card list overlapping the header

### LINKED ISSUE #293: Button to delete stacks is missing
Something is wrong with the hover zone so the button to delete stacks is inaccessible.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/48899830-button-to-delete-stacks-is-missing?utm_campaign=plugin&utm_content=tracker%2F52758431&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F52758431&utm_medium=issues&utm_source=github).
</bountysource-plugin>

juliushaertl:
fixes #293

This also gets rid of the ugly negative margin introduced with #258

### DIFF
--- a/css/style.scss
+++ b/css/style.scss
@@ -139,6 +139,12 @@ input.input-inline {
 	width: 100%;
 	bottom: 0px;
 	top: 44px;
+
+	&.as-sortable-un-selectable {
+		.card-list {
+			min-height: 96px;
+		}
+	}
 }
 
 #innerBoard {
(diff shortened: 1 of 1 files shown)


juliushaertl:
backported to stable-0.2 with 91fb153

PR #305, opened 2017-09-26
Change the word labels to tags for consistency

reginasipos:
I proposed this change because they're called tags everywhere else.

### DIFF
--- a/templates/part.card.php
+++ b/templates/part.card.php
@@ -38,7 +38,7 @@
 				   placeholder="<?php p($l->t('Add a label')); ?>"
 				   on-select="labelAssign($item, $model)"
 				   on-remove="labelRemove($item, $model)" ng-disabled="!boardservice.canEdit() || archived">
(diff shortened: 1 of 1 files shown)


jancborchardt:
Looks good to me! :) @nextcloud/deck can you check?

pixelipo:
It seems to me that there are more places where 'label' is used (in part.card.php):
"Choose a label"
"Add a label"
"Select labels…"

It would be good to fix them all

juliushaertl:
Makes sense. Thanks for your contribution, @reginasipos. As @pixelipo said, it would be great if you could change the other occurrences as well.

reginasipos:
@pixelipo @juliushaertl fixed the others :)

juliushaertl:
Signoff message is missing for commits, but let's merge, since it is a very small and language string only change.

juliushaertl:
Thanks again @reginasipos

PR #310, opened 2017-09-29
Change style of add card buttons to be less distracting

jancborchardt:
The full-color add buttons are a bit much when you actually have some cards in the stack, tends to be very distracting. ;) How about this:

Before:
screenshot from 2017-09-29 10-10-39
After:
screenshot from 2017-09-29 10-04-23

Only issue is I can’t seem to get the font-color visible. Any idea @nextcloud/deck?


juliushaertl:
@jancborchardt Thanks for the PR, looks pretty good to me. 

The font color for the input field is being set by angular here: https://github.com/nextcloud/deck/pull/310/files#diff-a8325f055da9615ba3ebbb5e7da3d790R112 After removing the color from ng-style the rule from the css should be applied.

juliushaertl:
Ah one more thing, maybe we should also add a border radius to the cards then, looks a bit odd now with only the add buttons being rounded.

juliushaertl:
@jancborchardt I've made some changes to this. 

bildschirmfoto vom 2017-10-24 12-33-13

css/style.scss
```diff
@@ -334,7 +334,8 @@ input.input-inline {
 	white-space: normal;
 	position: relative;
 	opacity: 1.0;
-	-webkit-box-shadow: 0 0 5px $color-darkgrey;
+	-webkit-box-shadow: 0 0 3px $color-darkgrey;
```
pixelipo: box-shadow only for legacy webkit broewsers?

jancborchardt:
@juliushaertl looks good to me! :)

juliushaertl:
@pixelipo Thanks, fixed.

jancborchardt:
So, is this good to go? Needs rebase now also ;)

juliushaertl:
Rebased. Let's merge since CI was happy before. 👍 

Thanks @jancborchardt looks much cleaner now

pixelipo:
@juliushaertl sorry I had no time to revisit this PR

jancborchardt:
Weehuu! :) Thanks, and happy to help!

Issue #253, opened 2017-07-11
Consistent saving and save feedback of contents of description field


jancborchardt:
### Steps to reproduce
1. Write a whole bunch of text into the description of a card
2. Look for a save button, or existing save feedback
3. Be very hesitant about what happens, copy the text of the textarea just to be safe

### Expected behaviour
There should be permanent saving feedback. Good comparison is like how we do it in the Text editor app: https://github.com/nextcloud/files_texteditor – it saves every time you don’t type for 500ms (I think). While you edit it shows an asterisk next to the title (we could improve that, any ideas?) and it then shows a »Saving …« and then »Saved!« feedback.

### Actual behaviour
You have no idea about if your data is saved and if not, how you could do it.

What do you think @nextcloud/deck? :)

juliushaertl:
Makes sense. It should be clear that the work that users do is saved.

PR #325, opened 2017-10-09
Card: Auto save while typing

juliushaertl:
Fix for #253 
peek 2017-10-09 13-39

### DIFF
--- a/css/style.scss
+++ b/css/style.scss
@@ -613,14 +613,21 @@ input.input-inline {
 	}
 
 	.save-indicator {
-		background-color: $color-success;
-		color: $color-primary-text;
 		border-radius: 3px;
 		float: right;
-		padding: 0px 10px;
-		font-size: 8pt !important;
+		padding: 0 10px;
+		font-size: 8pt;
 		display: none;
 		align-self: flex-end;
+		text-align: center;
+		&.saved {
+			background-color: $color-success;
+			color: $color-primary-text;
+		}
+		&.unsaved {
+			background-color: $color-lightgrey;
+			color: $color-darkgrey;
+		}
 	}
 
 	.icon-help {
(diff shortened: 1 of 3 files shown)


jancborchardt:
Looks good to me! @reginasipos check out the saving feedback. :)

css/style.scss
```diff
@@ -621,6 +621,20 @@ input.input-inline {
 		font-size: 8pt !important;
 		display: none;
 		align-self: flex-end;
+		min-width: 120px;
+		text-align: center;
+	}
+	.unsaved-indicator {
```
pixelipo: I would prefer having a new common class for both indicators because they share 80% of the declarations.

css/style.scss
```diff
@@ -621,6 +621,20 @@ input.input-inline {
...
+		background-color: $color-lightgrey;
+		color: $color-darkgrey;
+		border-radius: 3px;
+		float: right;
+		padding: 0px 10px;
+		font-size: 8pt !important;
```
pixelipo: `!important` is not needed here.

P.S. We should create variables for font sizes (in a separate PR), because it's getting kind of messy.

css/style.scss
```diff
@@ -621,6 +621,20 @@ input.input-inline {
 		font-size: 8pt !important;
 		display: none;
 		align-self: flex-end;
+		min-width: 120px;
```
pixelipo: IMO this is not needed, because it doesn't look bad if `.save-indicator` and `.unsave-indicator` do not have identical with. Also, if identical with is what you were aiming at, min-width will not solve the problem - what if one of the strings is longer than120px in another language?

juliushaertl:
@pixelipo Thanks, good points. I've fixed them.

css/style.scss
juliushaertl: You're right, it also looks fine without.

Issue #306, opened 2017-09-26
Ability to save tag changes with enter


reginasipos:
### Steps to reproduce
1. change name or color of a tag
2. try to save it with enter

### Expected behaviour
Saving via enter should work for consistency, like with other functionalities

### Actual behaviour
Only check mark saves currently

rpadovani:
I tried to fix it in #331

juliushaertl:
Fixed in #331 
Thank you @rpadovani

PR #331, opened 2017-10-18
Ability to save tag changes with enter

rpadovani:
Fix issue #306 

Signed-off-by: Riccardo Padovani <riccardo@rpadovani.com>

### DIFF
--- a/templates/part.board.sidebarView.php
+++ b/templates/part.board.sidebarView.php
@@ -73,9 +73,9 @@
                         <span ng-if="label.title">{{ label.title }}</span><i ng-if="!label.title"><br /></i>
                     </span>
                     <div class="label-edit" ng-if="label.edit">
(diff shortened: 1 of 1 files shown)


templates/part.board.sidebarView.php
```diff
@@ -75,7 +75,7 @@
                     <div class="label-edit" ng-if="label.edit">
                         <div ng-style="{'background-color':'#{{label.color}}','color':'{{ textColor(label.color) }}','width':'100%'}>
                             <form ng-submit="labelUpdate(label)">
```
juliushaertl: Thanks for your contribution. 

Looks like the issue here is the missing " in line 76. Then it should work fine due to the form ng-submit and we don't require the ng-keyup.

Mind to change that?

rpadovani: You are right, thanks for pointing it out. 

I was confused by the ng-submit behavior: I thought it worked only if there is a submit button, but it also work if there is only one input

Issue #245, opened 2017-07-11
Color for all created boards is the same unless modified


jancborchardt:
### Steps to reproduce
1. Create a board
2. Create another board
3. Create another board

### Expected behaviour
The first board should be blue, the next should be (the next color in line), the third should be (third color in line) automatically.

### Actual behaviour
All are blue and need to be colored manually.

9662:
FWIW, I am not sure whether this is a bug or a feature.

One way to approach this is to ask: what is colour intended to communicate in the decks list?
 * If it is simply to highlight that there are multiple decks, is it adding much value?
 * If the choice of meaning (if any) of deck colour is left to the user, wouldn't it be better to apply a single default colour, and for the user to manually change it if they wish?

Just my thoughts. Funnily enough, I came here with a list of issues, and you have posted them all (and then some) a couple of hours ago. This was one which I considered but then decided I would not mention. 😛

jancborchardt:
The point of a color is to quickly visually differentiate different boards. In a test it came to lots of confusion and tasks added to wrong boards because the title alone is not enough to differentiate (you don’t really look at it). See iOS Reminders for example: When you add a new list, the default selected color is one you haven’t used for any of the other lists (if possible).

Defaults are rarely changed by users and hence the defaults should always be really great. :)


@9662 nice :) keep the issues coming, especially design issues and detail findings are very appreciated on any Nextcloud app.

juliushaertl:
> The first board should be blue, the next should be (the next color in line), the third should be (third color in line) automatically.

Good point. 

@jancborchardt @9662 Any issue reporting is very welcome here. :wink:  I sadly don't have much time to focus on deck right now.

This might be a good starter issue, just check for existing boards colors and select the first unused color when opening the view to create a new board.

juliushaertl:
Fixed in #334. Thanks to @rpadovani

frinanta:
I want every user can choose their own color in nextcloud templet, eg user a want yellow color, user b want green color, can you help me, how to make it?

juliushaertl:
@frinanta Not possible at the moment and I don't see that coming, since it is a very specific feature that would add a lot of additional code that needs to be maintained.

PR #334, opened 2017-10-25
When creating a board, choose the less used color

rpadovani:
Fix issue #245 

I first tried a simple approach, trying to calculate which color using just counting how many boards there are. Of course this approach had a lot of issues, so I then tried to calculate which color was not used yet. While this works, what happens when all the colors have been used at least once?

Therefore I choose to make the approach the most generic possible, using the less used color and, if there was a draw, using the most left on the $scope.colors array.

This implementation should work perfectly also in future, if we change default colors or if we allow custom colors

Signed-off-by: Riccardo Padovani <riccardo@rpadovani.com>

### DIFF
--- a/js/controller/ListController.js
+++ b/js/controller/ListController.js
@@ -28,20 +28,18 @@ app.controller('ListController', function ($scope, $location, $filter, BoardServ
         var boardKeys = Object.keys(boards);
         var colorOccurrences = [];
 
-        for (var i =0; i < boardKeys.length; i++) {
-            var key = boardKeys[i];
+        for (var i = 0; i < $scope.colors.length; i++) {
+            colorOccurrences.push(0);
+        }
+
+        for (var j = 0; j < boardKeys.length; j++) {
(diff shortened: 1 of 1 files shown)


jancborchardt:
Thank you @rpadovani, nice contribution! :)

jancborchardt:
@rpadovani also, I just saw you are based in Munich. You should come to our monthly Nextcloud meetups which are nice hack-meetings to sit together and work on improving Nextcloud: https://www.meetup.com/nextcloud-munich/ :) cc @Henni & @irgendwie from the meetup

rpadovani:
@jancborchardt thanks for the invitation! I haven't joined so far because my German is still very bad - I am not able to talk, and as far as I can see the evenings are in German, aren't they?

jancborchardt:
@rpadovani don't worry, @mabkenar is also there sometimes and speaks mainly English, and for @Henni and @irgendwie it's not a problem. ;) What counts is that you are a great contributor!

PR #336, opened 2017-10-26
App menu compatibiliy

### LINKED ISSUE #315: App navigation is broken on NC 13
https://github.com/nextcloud/server/pull/6399

Let's wait for https://github.com/nextcloud/documentation/pull/565 to see how we should implement it.

- [ ] Board editing not working
- [ ] Adding boards not working

We may also need a fallback for NC11/12

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/50014461-app-navigation-is-broken-on-nc-13?utm_campaign=plugin&utm_content=tracker%2F52758431&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F52758431&utm_medium=issues&utm_source=github).
</bountysource-plugin>

juliushaertl:
This PR fixes the menu für NC13. I also reworked some of the ListController logic to make it more performant. 

- [x] Check for NC 12 compatibility
- [x] Fix closing the popover menu

fixes #315


juliushaertl:
@skjnldsv Any hints on how backward compatibility to stable12/11 should be handled with the new flex app sidebar?

skjnldsv:
@juliushaertl Depends on what types of items you got. :)
I'll go for a patch per apps to be removed when lowest version supported is >=13

js/controller/ListController.js
```diff
@@ -53,44 +54,67 @@ app.controller('ListController', function ($scope, $location, $filter, BoardServ
...
+			$scope.boardservice.sidebar = $filter('orderBy')($scope.boardservice.sidebar, 'title');
+			$scope.boardservice.sidebar = $filter('cardFilter')($scope.boardservice.sidebar, {archived: false});
 		} else {
-			$scope.boardservice.sorted = $filter('cardFilter')($scope.boardservice.sorted, {archived: false});
+			$scope.boardservice.sorted = $scope.boardservice.getData();
+			if ($scope.status.filter === 'archived') {
+				var filter = {};
+				filter[$scope.status.filter] = true;
+				$scope.boardservice.sorted = $filter('cardFilter')($scope.boardservice.sorted, filter);
+			} else if ($scope.status.filter === 'shared') {
```
rpadovani: Isn't better to use enums instead of string to indicate the filter state? :)

Maybe it can go to a separated issue

skjnldsv:
@juliushaertl Please don't hesitate to ask for help! :)

juliushaertl:
@skjnldsv Thanks. Actually it already works quite nice with those few changes: https://github.com/nextcloud/deck/pull/336/commits/fa7d9e9545c9bc8d6fe4b1b36c70721d4ac74628#diff-be704e29a41bc5b36d0f4761c521b82d

Right now only works with NC12, since I plan to drop NC11 support once 13 is out.

skjnldsv:
Yeah, it's juste a matter of transition! :)
It should not require big patches, simple ones like you did are what's cool about this.

Once 13 will be the lowest supported, it should be fine ;)

js/controller/ListController.js
juliushaertl: Hmm, the filter state is set as a state param where we have it as a string anyway. So until we don't use it more often in the code, i guess that is fine for now. But if you want to improve here, feel free to do so. 😉

juliushaertl:
Please review @nextcloud/deck

css/style.scss
```diff
@@ -1166,52 +1147,13 @@ input.input-inline {
...
-}
 
 /**
  * Hotfix for https://github.com/angular-ui/ui-select/issues/1652
  */
 .ui-select-dropdown.select2-drop-active {
 	opacity: 1 !important;
 }
+
+@import 'comp-appnav.scss';
```
pixelipo: shouldn't import go to the top?

css/legacy.scss
```diff
@@ -20,5 +20,18 @@
...
+$color-main-background: #fff;
+$color-primary-text: #fff;
+$color-error: #e9322d;
+$color-warning: #fbd850;
+$color-warning-light: nc-lighten($color-warning, 15%);
+$color-success: #46ba61;
+$color-lightgrey: nc-darken($color-main-background, 4%);
+$color-grey: nc-darken($color-main-background, 7%);
+$color-darkgrey: nc-darken($color-main-background, 32%);
+
```
skjnldsv: Why do you add the variables.scss content here?
The scss compilator automatically add this to any file you load with add_style :)

juliushaertl: Not for stable11, but we can drop this since we will only support version 12 or later in the next release.

skjnldsv: Yes, I asked because the info.xml says min version to 12. :)

juliushaertl:
Merging so we can move on. :wink:

skjnldsv:
Well done!! 👌😉

PR #339, opened 2017-11-08
Drop support for NC11

juliushaertl:
Nextcloud 11 will be EOL when NC13 is released, so we don't support it with the upcoming version 0.3.0 of deck.


rpadovani:
LGTM :+1:

Issue #341, opened 2017-11-09
mysql datetime format still wrong, version 0.2.5


RainerEmrich:
### Steps to reproduce
1. Insert due date for a card.
2. Close card.
3. Reopen card.

### Expected behaviour
The date and time should be saved.

### Actual behaviour
The date and time are not saved.

### Server configuration

**Operating system:** Ubuntu 16.04.

**Web server:** Apache 2.4 (fpm-fcgi)

**Database:** Mariadb 10.2.10

**PHP version:** 7.1.11-1+ubuntu16.04.1+deb.sury.org+1
<details>
        <summary>PHP-modules loaded</summary>

```
 - Core
 - date
 - libxml
 - openssl
 - pcre
 - zlib
 - filter
 - hash
 - Reflection
 - SPL
 - session
 - standard
 - cgi-fcgi
 - mysqlnd
 - PDO
 - xml
 - apcu
 - apc
 - bz2
 - calendar
 - ctype
 - curl
 - dom
 - mbstring
 - fileinfo
 - ftp
 - gd
 - gettext
 - gmp
 - iconv
 - igbinary
 - imagick
 - imap
 - intl
 - json
 - ldap
 - exif
 - mcrypt
 - msgpack
 - mysqli
 - pdo_mysql
 - pdo_pgsql
 - pdo_sqlite
 - pgsql
 - Phar
 - posix
 - readline
 - redis
 - shmop
 - SimpleXML
 - smbclient
 - sockets
 - sqlite3
 - sysvmsg
 - sysvsem
 - sysvshm
 - tokenizer
 - wddx
 - xmlreader
 - xmlwriter
 - xsl
 - zip
 - memcached
 - libsmbclient
 - Zend OPcache
```
</details>

**Nextcloud version:** 12.0.3 - 12.0.3.3

**Where did you install Nextcloud from:** Updated from 12.0.2 via web updater

**Signing status:** No errors have been found.

**List of activated apps:**
<details>
        <summary>Enabled apps</summary>

```
 - activity: 2.5.2
 - admin_audit: 1.2.0
 - admin_notifications: 1.0.0
 - announcementcenter: 3.1.0
 - apporder: 0.4.0
 - audioplayer: 2.1.0
 - bruteforcesettings: 1.0.2
 - calendar: 1.5.6
 - checksum: 0.3.5
 - circles: 0.13.6
 - comments: 1.2.0
 - contacts: 2.0.1
 - dav: 1.3.0
[...]

PR #342, opened 2017-11-09
Card: Set dbtype when updating

juliushaertl:
Missed when backporting #322 to stable-0.2


juliushaertl:
Confirmed to work in #341 by @RainerEmrich

Issue #370, opened 2017-12-20
Avatars are misplaced after saving a board


splitt3r:
**Description:** Avatars are misplaced after saving board changes:

**Nc Version:** 13 (from git)

**App Version:** v0.3.0-beta1 (from git)

**Browser:** FF 57

![avatars](https://user-images.githubusercontent.com/1869584/34214824-db20cfee-e5a3-11e7-93fe-343bde15a2c5.png)

Nav bar overlaps the display name tooltip:

![title](https://user-images.githubusercontent.com/1869584/34214816-d5870ab2-e5a3-11e7-885a-373a674bd069.png)

Here too:

![tooltip](https://user-images.githubusercontent.com/1869584/34214962-3e47c5be-e5a4-11e7-83b2-b77e98d032b3.png)

juliushaertl:
Fixed with #376

PR #376, opened 2018-01-08
Fix avatar issues

juliushaertl:
Improve avatar directive and add delete user assignment button to the avatar menu


juliushaertl:
Fixes #370

PR #451, opened 2018-05-01
Adding Printing view, fixes #195

### LINKED ISSUE #195: Add Print/Export overview options
Feature request: Add a print or export overview option to allow the easy display all the information on a board (including each cards details and labels) so it could be discussed around a table rather than everyone looking at a screen in a meeting.

Previous discussion:
https://help.nextcloud.com/t/why-our-team-decided-not-to-use-deck-app-yet/14148

Dubidubiduu:
- Hide navigation stuff
- Printing first line of description
- Breaking page after 6 stacks

Code works if it is pasted in style.scss but i don't got it to get imported in a separate file, maybe you have an idea...?!

Still don't know how to make it properly scale by default, so you still have to scale up, for example up to 200 % in Firefox.

Fix for #195


juliushaertl:
Thank you very much @Dubidubiduu Can you try adding `@import 'print.scss';` to the styles.scss file? That should just work fine.

> Still don't know how to make it properly scale by default, so you still have to scale up, for example up to 200 % in Firefox.

Can you show me what looks wrong there, because to me the print view already looks ok so far :wink:

juliushaertl:
@Dubidubiduu Any news on this? Would be nice if we could at least have a first basic print stylesheet for now. :)

Dubidubiduu:
Sorry for not pushing this forward. :-/

But now the truth is that I really need to get some sleep and I have my problems with using git, should read a tutorial or so, and on top I've got some nasty caching problems, which aren't solved yet. If you already could use those lines and it's not an act for you you're very welcomed to just do it.

Otherwise it would take a few days to have some time to dig into this.

PR #488, opened 2018-06-11
File attachments for cards

juliushaertl:
This PR implements uploading file attachments to cards. The files are stored in AppData and can be downloaded by every user with read access to the board.

### List of attachments
image

### Dragging files to the board
bildschirmfoto vom 2018-06-14 11-34-29

### Insert attachments into description
image

## ToDo
- [x] Backend
  - [x] File listing
  - [x] File upload
  - [x] File delete
  - [x] Remove file on card deletion
  - [x] File update
  - [x] Move file deletion to background job
  - [x] Fix PDF display issue in chrome (related to CSP https://bugs.chromium.org/p/chromium/issues/detail?id=271452)
- [x] Frontend
  - [x] File listing in card details
  - [x] Use mime type icons
  - [x] File deletion
  - [x] Upload using a button
  - [x] Add drop-zone to the card details view
  - [x] Improve storing data in CardService
  - [x] Filename collision handling 
  - [x] Move file list to tabbed view
  - [x] Add drop-zone for upload to the board view
  - [x] Add indicator about file attachments to board view
  - [x] Insert attachment into card description
- [x] Unit tests
  - [x] Adjust Entity tests
  - [x] DeleteCron 
  - [x] AttachmentMapper
  - [x] AttachmentService
  - [x] FileService
  - [x] AttachmentController


juliushaertl:
@nextcloud/designers Some feedback on the UI would be great. I've updated the screenshots above with the current state.

cc @nextcloud/deck for review

jancborchardt:
Oh wow, that seems really nice! :) Have only small pieces of feedback:
- A bit more vertical whitespace between the rows of the attachments, both in the sidebar and in the popup
- Will file thumbnails be shown?
- Instead of tabbed interface, why not show the attachments directly below the description?
- Also, then we could simply show them bigger, not only as a thumbnail

juliushaertl:
> A bit more vertical whitespace between the rows of the attachments, both in the sidebar and in the popup

:+1: 

> Will file thumbnails be shown?

This is out of scope for now, since previews are not available for files stored in the AppData. I've discussed some possible implementation ideas with @rullzer but this is something for later.

> Instead of tabbed interface, why not show the attachments directly below the description?

I had that in the beginning, but in most cases, the description will contain the most important content of the card and links to attachments/images can also be embedded there, in case users want direct access to it. Also for mobile devices it makes sense to limit the amount of shown information to the description while providing a list of attachments in a separate tab. Otherwise we'll quickly end up with the notification being not visible since it has moved down outside of the visible area.

> Also, then we could simply show them bigger, not only as a thumbnail

@jancborchardt What do you mean by displaying them bigger?

As for the embedding of attachments into the description, this is how it looks like for images/regular files:

## Markdown
image
## Rendering
image

jancborchardt:
Yeah, so the thing is that simply displaying the images below the description is much simpler. No one needs to manually embed stuff then.

With them showing bigger, I basically mean how they are shown when embedded. Maybe we can limit the height a bit when there's many attachments though.

> Otherwise we'll quickly end up with the notification being not visible since it has moved down outside of the visible area.

What do you mean by that? No important stuff should be displayed _below_ the images.

juliushaertl:
> Yeah, so the thing is that simply displaying the images below the description is much simpler. No one needs to manually embed stuff then.

I'd say it should be up to the user which images are shown directly and wich are not that important. Also the attachments might in a lot of cases have some relation to sections of the description, so I think embedding is a better way then just listing them.

> With them showing bigger, I basically mean how they are shown when embedded. Maybe we can limit the height a bit when there's many attachments though.

Good point. Width is limited, but i'll also add a limit for the height.

>> Otherwise we'll quickly end up with the notification being not visible since it has moved down outside of the visible area.

> What do you mean by that? No important stuff should be displayed below the images.

I mean description not notification :see_no_evil:

juliushaertl:
Height limit for image and additional spacing is added. Let's merge this and iterate over it for further adjustments :wink:

pixelipo:
Awesome work @juliushaertl !

PR #535, opened 2018-07-14
Feature/83/json api

Nebri:
The big heavy development work is now done and ready for it's first review round.

I'm planning on writing a series of unit tests against the Api Controllers, and after doing the rebase a bunch of unit tests that were passing are no longer. I'll start taking a look at these units tests later this weekend.

When this PR gets merged it will close Issue #83

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -74,18 +74,18 @@ return [
 		['name' => 'label#update', 'url' => '/labels/{labelId}', 'verb' => 'PUT'],
 		['name' => 'label#delete', 'url' => '/labels/{labelId}', 'verb' => 'DELETE'],
 
-        // api
-        ['name' => 'board_api#index', 'url' => '/api/v1.0/boards', 'verb' => 'GET'],
-        ['name' => 'board_api#get', 'url' => '/api/v1.0/board/{id}', 'verb' => 'GET'],
-        ['name' => 'board_api#create', 'url' => '/api/v1.0/board', 'verb' => 'POST'],
-        ['name' => 'board_api#delete', 'url' => '/api/v1.0/board/{id}', 'verb' => 'DELETE'],
-        ['name' => 'board_api#undo_delete', 'url' => '/api/v1.0/board/{id}/undo_delete', 'verb' => 'POST'],
+		// api
+		['name' => 'board_api#index', 'url' => '/api/v1.0/boards', 'verb' => 'GET'],
+		['name' => 'board_api#get', 'url' => '/api/v1.0/board/{id}', 'verb' => 'GET'],
+		['name' => 'board_api#create', 'url' => '/api/v1.0/board', 'verb' => 'POST'],
+		['name' => 'board_api#delete', 'url' => '/api/v1.0/board/{id}', 'verb' => 'DELETE'],
+		['name' => 'board_api#undo_delete', 'url' => '/api/v1.0/board/{id}/undo_delete', 'verb' => 'POST'],
 
-        ['name' => 'stack_api#index', 'url' => '/api/v1.0/board/{boardId}/stacks', 'verb' => 'GET'],
-        ['name' => 'stack_api#create', 'url' => '/api/v1.0/board/{boardId}/stack', 'verb' => 'POST'],
-        ['name' => 'stack_api#delete', 'url' => '/api/v1.0/board/{boardId}/stack/{id}', 'verb' => 'DELETE'],
+		['name' => 'stack_api#index', 'url' => '/api/v1.0/board/{boardId}/stacks', 'verb' => 'GET'],
+		['name' => 'stack_api#create', 'url' => '/api/v1.0/board/{boardId}/stack', 'verb' => 'POST'],
+		['name' => 'stack_api#delete', 'url' => '/api/v1.0/board/{boardId}/stack/{id}', 'verb' => 'DELETE'],
 
-        ['name' => 'board_api#preflighted_cors', 'url' => '/api/v1.0/{path}',
-         'verb' => 'OPTIONS', 'requirements' => ['path' => '.+']],
+		['name' => 'board_api#preflighted_cors', 'url' => '/api/v1.0/{path}',
+			'verb' => 'OPTIONS', 'requirements' => ['path' => '.+']],
 	]
 ];
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -91,6 +91,7 @@ return [
 
 		['name' => 'card_api#get', 'url' => '/api/v1.0/boards/{boardId}/stacks/{stackId}/cards/{cardId}', 'verb' => 'GET'],
 		['name' => 'card_api#create', 'url' => '/api/v1.0/boards/{boardId}/stacks/{stackId}/cards', 'verb' => 'POST'],
+		['name' => 'card_api#update', 'url' => '/api/v1.0/boards/{boardId}/stacks/{stackId}/cards/{cardId}', 'verb' => 'PUT'],
 
 		['name' => 'board_api#preflighted_cors', 'url' => '/api/v1.0/{path}','verb' => 'OPTIONS', 'requirements' => ['path' => '.+']],
 	]
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -82,10 +82,10 @@ return [
 		['name' => 'board_api#update', 'url' => '/api/v1.0/boards/{id}', 'verb' => 'PUT'],
 		['name' => 'board_api#undo_delete', 'url' => '/api/v1.0/boards/{id}/undo_delete', 'verb' => 'POST'],
 
-		['name' => 'stack_api#index', 'url' => '/api/v1.0/board/{boardId}/stacks', 'verb' => 'GET'],
-		['name' => 'stack_api#create', 'url' => '/api/v1.0/board/{boardId}/stacks', 'verb' => 'POST'],
-		['name' => 'stack_api#update', 'url' => '/api/v1.0/board/{boardId}/stacks/{id}', 'verb' => 'PUT'],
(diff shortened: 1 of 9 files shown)


Nebri:
@juliushaertl I've managed to reduce the number of errors that were happening with the units tests down to 1. It was pretty simple, phpunit removed setExpectedException and replaced it with a new function. Here's my reference: https://github.com/Roave/BetterReflection/issues/170

The last error is eluding me at the moment, mind giving insights on how to proceed past that error? I'll be starting my own unit tests against the new API real soon :).

appinfo/routes.php
```diff
@@ -74,5 +75,30 @@
...
+		// api
+		['name' => 'board_api#index', 'url' => '/api/v1.0/boards', 'verb' => 'GET'],
+		['name' => 'board_api#get', 'url' => '/api/v1.0/boards/{boardId}', 'verb' => 'GET'],
+		['name' => 'board_api#create', 'url' => '/api/v1.0/boards', 'verb' => 'POST'],
+		['name' => 'board_api#delete', 'url' => '/api/v1.0/boards/{boardId}', 'verb' => 'DELETE'],
+		['name' => 'board_api#update', 'url' => '/api/v1.0/boards/{boardId}', 'verb' => 'PUT'],
+		['name' => 'board_api#undo_delete', 'url' => '/api/v1.0/boards/{boardId}/undo_delete', 'verb' => 'POST'],
+		
+		['name' => 'stack_api#index', 'url' => '/api/v1.0/boards/{boardId}/stacks', 'verb' => 'GET'],
+		['name' => 'stack_api#index', 'url' => '/api/v1.0/boards/{boardId}/stacks/{stackId}', 'verb' => 'GET'],
```
juliushaertl: Shouldn't that be handled by stack_api#get

Nebri: yup! a simple copy/paste oversight!

lib/Controller/CardApiController.php
```diff
@@ -0,0 +1,242 @@
...
+		$stackError = $this->apiHelper->entityHasError($this->request->params['stackId'], 'stack', $this->stackService);		
+		if ($stackError) {
+			return new DataResponse($stackError['message'], $stackError['status']);
+		}
+
+		if (is_numeric($this->request->params['cardId']) === false) {
+			return new DataResponse('card id must be a number', HTTP::STATUS_BAD_REQUEST);
+		}
+
+		$card = $this->cardService->find($this->request->params['cardId']);
```
juliushaertl: Something more general regarding the service classes. They actually throw various exceptions that come from either the permissionService or the db mappers. Sorry those are not correctly documented in the phpDoc comments. So we need to either add a try/catch block here or extend the service methods to return null in case there is no enity found.

Another approach would be to have a middleware for the API that catches the exceptions and returns a 404 for DoesNotExistException for example.

Nebri: Which would you prefer? I'm usually all for the try { catch } blocks as it makes the code that much more clear / self documenting, typically avoiding exactly this kind of scenario. I did a test for (false || null) just as a starting point.

All the services are able to throw these exceptions?

juliushaertl: Yes, that makes sense. Let's go that way. Basically yes, but I can push some commit to add proper annotations later, if that helps.

juliushaertl:
@Nebri I quickly pushed a commit to fix the one failing test.

juliushaertl:
@Nebri Any another one to add the exception annotations. At some point it would probably be a good idea to catch the database exceptions in the service classes and throw exceptions that better show what exactly went wrong.

E.g. for the CardServce::update method we currently have:
NoPermissionException which is fine
StatusException if the board or card is archived
MultipleObjectsReturnedException|DoesNotExistException if the query for the card goes wrong: Controllers don't need to know about that, a NotFoundException for the cardId would be the best here I guess.

Would that make sense from your POV?

Nebri:
@juliushaertl Mind taking a look over my commit: ede2144? Getting a bit lost in the weeds on mocking up a unit test that will cause the permissions to fail, and throw the permission exception. Once I figure that out, I'll be able to make a test to ensure the proper DataResponse is generated.

Nebri:
@juliushaertl 

> The board service should throw a NoPermissionEyception for this test case. Since we're unit testing there is no need to check if the PermissionService method is called. We can just assume that this is handled correctly.

Is this true for all the exceptions that could be occuring in the rest api? I've taken a look at SharingMiddleware and I've noticed it's taking exceptions and returning out json responses.

if you look at the Board / Stack api controllers I've defined, I'm also trying to build the same json responses using the DataResponse class. Am I essentially creating duplicating code that I don't need to?

The rest api code branch so far is at 2,000 lines, and based on our discussions this is seeming to get out of hand. Want to make 100% sure I'm not wasting effort re-inventing the wheel here :).

juliushaertl:
> Is this true for all the exceptions that could be occuring in the rest api? I've taken a look at SharingMiddleware and I've noticed it's taking exceptions and returning out json responses.

Yes, at least for the NoPermissionException, which is a StatusException that is catched by the Middleware.

> if you look at the Board / Stack api controllers I've defined, I'm also trying to build the same json responses using the DataResponse class. Am I essentially creating duplicating code that I don't need to?

You can get rid of those checks:
```
if ($board === false || $board === null) {
```

The middleware will automatically return a JSONResponse for any StatusException that is thrown in a controller. 

We probably should also move the other checks to the services and just throw a StatusException for those as well, but I need to think about that a bit.

Nebri:
Alright I'm going to pause work on this branch until we have a solid plan of attack for how to handle these exceptions, as it greatly affects the complexity and location of the relevant unit tests.

Nebri:
@juliushaertl This is ready for another review round.

lib/BadRequestException.php
```diff
@@ -0,0 +1,36 @@
...
+ *
+ *  You should have received a copy of the GNU Affero General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\Deck;
+
+use OCP\AppFramework\Http;
+class BadRequestException extends \Exception {
```
juliushaertl: extend from `StatusException` so we don't need an additional check for that in the Middleware.

lib/Middleware/SharingMiddleware.php
```diff
@@ -59,7 +60,7 @@ public function __construct(ILogger $logger, IConfig $config) {
 	 * @throws \Exception
 	 */
 	public function afterException($controller, $methodName, \Exception $exception) {
-		if ($exception instanceof StatusException) {
+		if ($exception instanceof StatusException || $exception instanceof BadRequestException) {
```
juliushaertl: Not needed if BadRequestException is a StatusException.

lib/Controller/LabelApiController.php
```diff
@@ -0,0 +1,108 @@
...
+	
+	/**
+	 * @NoAdminRequired
+	 * @CORS
+	 * @NoCSRFRequired
+	 *
+	 * Get a specific label.
+	 */
+	public function get() {
+		$label = $this->labelService->find($this->request->getParam('labelId'));
```
juliushaertl: We still need some checks as in the other services here as well :wink:

lib/Controller/LabelApiController.php
```diff
@@ -0,0 +1,108 @@
...
+	 * @NoAdminRequired
+	 * @CORS
+	 * @NoCSRFRequired
+	 *
+	 * @params $title
+	 * @params $color
+	 * Create a new label
+	 */
+	public function create($title, $color) {
+		$label = $this->labelService->create($title, $color, $this->request->getParam('boardId'));
```
juliushaertl: We still need some checks as in the other services here as well :wink:

juliushaertl:
@Nebri Looks good from the Service/Controller separation now.

Some things we still need endpoints for, or maybe we can adjust the current ones to offer an option for are:
- Card: Assign/unassign labels
- Card: Assign/unassign users
- Card: reorder
- Stacks: fetch archived
- Attachments

I'd be fine if we obmit the following for the first version of the API:
- Board: add/update/delete acl (the first api version should be fine without sharing/updating the permissions)

But it would of course nice to also have those :wink:

Nebri:
Of course, let's do it right! I have no intention of rushing this. I've been taking job interviews this week, next week is looking clear for the moment.

Nebri:
@juliushaertl I've just written the attachments api. I'm assuming that we also want to go through the services to put in the BadRequestException across the board?

I'd also like to request that we do the acl permissions in a separate issue. I'd like to keep the scope creep in check. Let's fix anything that may still be outstanding here, get it merged, and then build the permissions on top.

let me know what you think.

juliushaertl:
> I'm assuming that we also want to go through the services to put in the BadRequestException across the board? 

That would be good yes.

> I'd also like to request that we do the acl permissions in a separate issue. I'd like to keep the scope creep in check. Let's fix anything that may still be outstanding here, get it merged, and then build the permissions on top.

Sure, we can do this later on. Maybe you can just create an issue for that, so we don't forget about it. That will also make it a bit easier to review then. Better have several small steps there :wink:

srbaker:
Heya, thanks for jumping on this. @Nebri Every time I had planned to work on it, something else came up, and I just got stuck. Really glad to see this moving along!

Nebri:
@srbaker @juliushaertl Appreciate the kind words! I'll be out of town this week, need some time at my summer cottage to unwind. I'm planning to get back on this next week, maybe sooner if I find myself really bored on a rainy day and a coffee shop with wifi, hehe.

Nebri:
@juliushaertl added the additional checks into the services as discussed. Ready for a review / feedback.

juliushaertl:
@Nebri Great work, I've added some minor fixes and the CI is happy again. I did some testing and the endpoint work as expected. :+1:

PR #536, opened 2018-07-14
click on assign user or the plus button

philippe-levan:
Hi,

We often add users on cards. I believe the button to assign a user shoud be bigger.

I suggest with this PR to make all the header clickable to assign a new user (but I let the button with the hover to let the UI intuitive).

Best regards,
Phillippe

### DIFF
--- a/templates/part.card.php
+++ b/templates/part.card.php
@@ -53,9 +53,9 @@
 			</ui-select-choices>
 		</ui-select>
 	</div>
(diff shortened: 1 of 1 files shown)


juliushaertl:
@philippe-levan Thanks for the pull request. Makes a lot of sense to have the click handler on the whole header element.

philippe-levan:
thanks for the merge and the tag !

PR #545, opened 2018-07-22
Add notification on card assignment to user, fixes #475

### LINKED ISSUE #475: Notifications on card assignment
I want to use Deck in a large research project and unfortunately the following important features are missing to make that easy:

* Notifications when assigning a card: In a big project with many boards and cards it is important to notify the user when a card is assigned to her/him.

* Cards assigned to me: In a big project with many boards and cards it would be nice to quickly find (Active) cards assigned to self, either in a list or as a number behind each board name in the boards list page.

Thank you and best regards,

Maher

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/59069258-notifications-on-card-assignment?utm_campaign=plugin&utm_content=tracker%2F52758431&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F52758431&utm_medium=issues&utm_source=github).
</bountysource-plugin>

steav:


lib/Notification/Notifier.php
```diff
@@ -73,6 +73,18 @@ public function prepare(INotification $notification, $languageCode) {
...
+			case 'card-assigned':
+				$cardId = $notification->getObjectId();
+				$boardId = $this->cardMapper->findBoardId($cardId);
+				$initiator = $this->userManager->get($params[2]);
+				if ($initiator !== null) {
+					$params[2] = $initiator->getDisplayName();
+				}
+				$notification->setParsedSubject(
+					(string) $l->t('The card "%s" on "%s" has been assigned to you by %s.', $params)
+				);
```
juliushaertl: Can you add a rich subject as well, similar to https://github.com/nextcloud/deck/pull/545/files#diff-52c5c003756b8dcc9c5318257107346dL95? This way the user will also be linked properly

juliushaertl:
Thank you very much for your pull request @steav, looks good besides the inline comment. Can you also remove the changes to the language files. We are managing translations via transifex and those strings will be pulled from there: https://www.transifex.com/nextcloud/nextcloud

juliushaertl:
@steav Can you also add a sign-off message to the commits as described in https://github.com/nextcloud/deck#sign-your-work ?

steav:
Thanks you a lot @juliushaertl for your advices and your awesome work!

juliushaertl:
@steav Can you also have a look at the failing tests and add some basic ones for the new methods?

lib/Service/CardService.php
```diff
@@ -201,6 +204,11 @@ public function assignUser($cardId, $userId) {
 				return false;
 			}
 		}
+
+		/* Notify user about the card assignment */
+		$card = $this->cardMapper->find($cardId);
+		$this->notificationHelper->sendCardAssigned($card, $userId);
```
juliushaertl: We should only send a notification if the user assigns somebody else. If you assign yourself, there is no need for a notification.

juliushaertl:
@steav Really nice work done here. Tested and works fine. I just have one small enhancement commented inline. After that it is good to go. :wink:

lib/Notification/Notifier.php
steav: Done

lib/Service/CardService.php
steav: Done

juliushaertl:
merged. :rocket: Thank you very much for your contribution @steav Further pull requests are always welcome. :wink:

Issue #10, opened 2017-02-01
Comments [$120]


juliushaertl:
Similar to the files app it should be possible for users to comment on each card. 

- [ ] Check if it is possible to use the server comments app

joergmschulz:
I'm not sure how this could be done without cluttering the user interface. I like the simplicity of the current concept (well, dragging the cards would be a plus, you already opened an issue for this one) IMHO it is enough to have the description for the cards.

juliushaertl:
@joergmschulz I would prefer to keep the markdown description as a default and just use tabs for other stuff like the comments, just like it is handled in the files app.

![2017-02-01-193228_511x404_scrot](https://cloud.githubusercontent.com/assets/3404133/22520596/3ceab3f8-e8b5-11e6-80ee-eae98a28fdd3.png)

jancborchardt:
Yup, very important for collaboration too. cc @blizzz for comments.

blizzz:
Yes, it's possible to use comments in other apps. One example that does it is the [announcement center app](https://github.com/nextcloud/announcementcenter), related PRs: https://github.com/nextcloud/announcementcenter/pulls?q=is%3Apr+comment+is%3Aclosed

juliushaertl:
Thanks for the hint.

Nebri:
This seems pretty straight forward. @juliushaertl I'd be willing to give this a go. After reading this over it sounds like the majority wishes to see the comments as a tab under the details view that comes up on the right panel after clicking a card?

View screenshot below for how that comments tab would appear. Once you give me the go ahead I'll happily create a new branch for this issue and get to work. Been a long time since I last got to do any angularjs work!

![image](https://user-images.githubusercontent.com/3360697/42609181-3c85cb5e-8559-11e8-949a-d1c0a1007518.png)

jancborchardt:
> it sounds like the majority wishes to see the comments as a tab under the details view that comes up on the right panel after clicking a card?

Yes – just keep in mind that we also have a plan in the Files app to combine the chronological information (Activity, Comments, Versions) into one timeline. Not sure if anything else is chronological for now, but keep that in mind for the future @juliushaertl @Nebri :)

juliushaertl:
Good point @jancborchardt We need to remind that for comments and activity #211

JumpLink:
@Nebri Do you want to implement this feature? Did you need help?

Nebri:
[...]

PR #632, opened 2018-10-04
Comments support

juliushaertl:
* Resolves: #10 
* Target version: master 

### Summary

This adds suport for comments on cards

image

### TODO

- [x] Unified timeline view
- [x] Show comments in activity list
- [x] Editing comments 
- [x] Deleting comments
- [x] Show avatar instead of activity entry with subject
- [x] @-mentions
- [x] Notifications on mentions
- [ ] Code cleanup
  - [x] Move content editable directive to dedicated file
  - [x] Properly inject the comment manager in CardService and DeckProvider
- [ ] Tests

### Checklist


js/app/Run.js
```diff
@@ -4,23 +4,48 @@
...
- *  
+ *
  *  You should have received a copy of the GNU Affero General Public License
  *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
- *  
+ *
  */
 import app from './App.js';
 
+app.directive("contenteditable", function() {
```
juliushaertl: Move this to a separate file

juliushaertl:
This PR also introduces a unified tab for comments/activity as discussed in https://github.com/nextcloud/server/issues/658 / https://github.com/nextcloud/server/issues/10289

image

- How should the tab be called, maybe "Timeline"?
- How should we handle deleted comments. In my oppinion they should still appear in the activity history, but it doesn't make sense to show them in the combined view.

@nextcloud/designers Some feedback on the design would be great.

Issue #631, opened 2018-10-03
[Design feature request] Default view without left menu


biva:
Thank you for this great app that we use all days.

The way we use it is with specific URL for different boards, for example: https://mynextcloud.com/apps/deck/#!/board/1/ for the board 1

But when I open this URL, the left menu is present: If I want to see only my board on the full screen, I need to click on the 3-bars menu (next to the "Home" icon).

Is there a possibility to hide the left menu by default? Or to add a parameter in the URL to hide it?

Thank you!

juliushaertl:
I think the best way would be to just store the shown/hidden status when toggling the sidebar using the burger button.

weeman1337:
I'll have a look at this one

PR #723, opened 2018-11-21
Deck now remembers the side menu state

weeman1337:
* Resolves: #631 
* Target version: master 

### Summary

The side menu state is now stored in local storage.  
I've also updated the jshint file and at least the code I touched to make it compliant.

### Checklist

- [X] Tests (unit, integration, api and/or acceptance) are included - none
- [X] Documentation (manuals or wiki) has been updated or is not required - I don't think it's required

### DIFF
--- a/js/.jshintrc
+++ b/js/.jshintrc
@@ -1,4 +1,6 @@
 {
+  "esversion": 6,
+
   "globals": {
     "jasmine"    : false,
     "spyOn"      : false,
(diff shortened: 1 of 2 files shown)


js/.jshintrc
```diff
@@ -1,4 +1,6 @@
 {
+  "esversion": 6,
```
weeman1337: I switched JSHint to ES6, because Deck uses ES6 for it's code.

js/.jshintrc
```diff
@@ -39,7 +40,6 @@
   "plusplus"  : false,
   "quotmark"  : "single",
   "regexp"    : false,
-  "strict"    : true,
```
weeman1337: ES6 modules are always strict; so this isn't required any more

PR #776, opened 2018-12-08
Move to database migrations

juliushaertl:
- [x] BigInt migration
- [ ] Migrate due date to datetimetz


juliushaertl:
Let's redo this after 0.6 is out since there were some db changes.

Issue #771, opened 2018-12-07
API/JSON incosistencies


AndyScherzinger:
Hi @juliushaertl 
This is rather a question/discussion, can really say if it is to be considered a feature, a bug orr to be ignored.

Please see below the actual (0.5.1) JSON element for a card retrieved in the Android client, see https://github.com/stefan-niedermann/nextcloud-deck/pull/16#issuecomment-445237500
```
{
"title":"Fix some naughty bug",
"description":"getRandom always returns four\nThis is line two\nThis is line three",
"stackId":8,
"type":"plain",
"lastModified":1544169560,
"lastEditor":"andy",
"createdAt":1544169433,
"labels":[
{"title":"To review","color":"317CCC","boardId":"5","cardId":null,"id":18},
{"title":"Later","color":"F1DB50","boardId":"5","cardId":null,"id":20}
],
"assignedUsers":[
{"id":2,"participant":{"primaryKey":"andy","uid":"andy","displayname":"Andy"},"cardId":"24"}
],
"attachments":null,
"attachmentCount":0,
"owner":"andy",
"order":999,
"archived":false,
"duedate":"2018-12-14T07:59:20+00:00",
"deletedAt":0,
"commentsUnread":0,
"id":24,
"overdue":0
}
```
* the label/```cardId``` of the labels is always `null` and  #774 
* the label/`boardId` is a `String` instead of a `long` #774 
* the assignedUser/```cardId``` of the assignedUsers is is a `String` instead of a `long` #774 
* `lastModified`, `createdAt` and `duedate` are all dates/timestamps while being mixed types long/date-string where the date string ships the timezone (kudos for that!) #809 

Any feedback is highly appreciated, as always :) ❤️

juliushaertl:
Thanks for reporting. This is actually a bug. I have a fix ready for the first 3 list items (#774), the date type unification however need a deeper look, since changing the types will probably also need some adjustments at the frontend of deck. I'm looking into this.

AndyScherzinger:
Thanks a lot for your help and efforts @juliushaertl! Highly appreciate your work!

PR #774, opened 2018-12-07
Fix numeric types and missing card id in card detail results

juliushaertl:
Fixes #771

PR #809, opened 2018-12-20
Fix datetime inconsistencies in api responses

juliushaertl:
* Resolves: #771
* Target version: master 

### Summary

lastModified, createdAt and duedate are all dates/timestamps while being mixed types long/date-string where the date string ships the timezone (kudos for that!)

Separated from #774 since it requires some more work on the frontend.


juliushaertl:
We should unify the date format in the database first, once we have migrations available. Let's close this for now.

Issue #586, opened 2018-08-13
Prevent duplicate tag names


ashgoodman:
**Describe the bug**
I am able to create duplicate tags (identical in spelling, capitalization etc) for the same board

**To Reproduce**
1. Create tag for a board
2. Create another tag for the same board with an identical name

**Expected behavior**
A warning message that the tag already exists for that board

juliushaertl:
Makes sense, however this is nothing to urgent. Thanks for reportin :+1:

ashgoodman:
Yep, I understand.

Glad to help improve the project!

PR #814, opened 2018-12-20
Prevent duplicate tag names

jakobroehrl:
* Resolves: #586 
* Target version: master 

### Summary
entering a duplicate title is not possible

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/lib/Service/LabelService.php
+++ b/lib/Service/LabelService.php
@@ -90,15 +90,10 @@ class LabelService {
 			throw new BadRequestException('board id must be a number');
 		}
 
-		$boardLabels = $this->labelMapper->findAll($boardId);
-		foreach($boardLabels as $boardLabel) {
-			if ($boardLabel->getTitle() === $title) {
-				throw new BadRequestException('title must be unique');
-				break;
-			}
-		}
-
 		$this->permissionService->checkPermission(null, $boardId, Acl::PERMISSION_MANAGE);
+
+		$this->checkDuplicateTitle($boardId, $title);
+
 		if ($this->boardService->isArchived(null, $boardId)) {
 			throw new StatusException('Operation not allowed. This board is archived.');
 		}
@@ -160,14 +155,33 @@ class LabelService {
 		}
 
 		$this->permissionService->checkPermission($this->labelMapper, $id, Acl::PERMISSION_MANAGE);
+		
+		$label = $this->find($id);
+		$this->checkDuplicateTitle($label->getBoardId(), $title);
+		
 		if ($this->boardService->isArchived($this->labelMapper, $id)) {
 			throw new StatusException('Operation not allowed. This board is archived.');
 		}
-		$label = $this->find($id);
+
 		$label->setTitle($title);
 		$label->setColor($color);
 		$this->changeHelper->boardChanged($label->getBoardId());
 		return $this->labelMapper->update($label);
+	}
+
+	/**
+	 * @param $boardId
+	 * @param $title
+	 * @throws BadRequestException
+	 */
+	private function checkDuplicateTitle($boardId, $title) {
+		$boardLabels = $this->labelMapper->findAll($boardId);
+		foreach($boardLabels as $boardLabel) {
+			if ($boardLabel->getTitle() === $title) {
+				throw new BadRequestException('title must be unique');
+				break;
+			}
+		}
 	}
 
 }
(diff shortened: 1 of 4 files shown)


juliushaertl:
Thank you very much for your contribution @jakobroehrl. Can you also add this to the update method for labels? Otherwise creating duplicate labels by renaming them is still possible. Maybe you can move the check to a separate private method, so we don't have duplicate code.

jakobroehrl:
@juliushaertl
Thanks for your advice. I added a new function and refered it from update and create.

- My check in the update function has a problem: In the UI remains the duplicate title until reloading the page. Do you have an idea where I can fix that behaviour?

- visual indicator in the UI: The visual indicator for the other label checks (title === false, color === false, ...) is missing, too. I would love to implement it. Are there any code snips oder examples I can use?

Thanks and Greetings!

juliushaertl:
You should be able to add a second callback to the then call here, to catch the error when sending the create request to the server:
https://github.com/nextcloud/deck/blob/master/js/controller/BoardController.js#L330

Although it isn't nice, i would just use `OC.Notification.showTemporary('Duplicate label name is not allowed');` to render a notification to the user, since we are currently reworking our UI anyway.

juliushaertl:
> How can I distinguish the different error messages in the BoardController? I get the same message on every error. BoardController.js line 335

This seems to be an issue with the current way our service layer works. The generic error message is coming from https://github.com/nextcloud/deck/blob/master/js/service/ApiService.js#L122 so we would need to return the proper error from the server here.

> If the label update fails I want to set the value from before. Where can I get it 
from? In $scope and label is only the new value. BoardController.js line 343

You need to add variable to store the previous value during the editing process, similar to what we do when editing the board title: https://github.com/nextcloud/deck/blob/master/js/controller/BoardController.js#L47

jakobroehrl:
@juliushaertl: I hope, it's ready to merge now :)

juliushaertl:
@jakobroehrl I rebased this to let the tests run against current master. Any change you could look into the failing unit tests:

```

There were 2 errors:
1) OCA\Deck\Service\LabelServiceTest::testCreate
Invalid argument supplied for foreach()
/home/travis/build/nextcloud/server/apps/deck/lib/Service/LabelService.php:97
/home/travis/build/nextcloud/server/apps/deck/tests/unit/Service/LabelServiceTest.php:74
2) OCA\Deck\Service\LabelServiceTest::testUpdate
Invalid argument supplied for foreach()
/home/travis/build/nextcloud/server/apps/deck/lib/Service/LabelService.php:170
/home/travis/build/nextcloud/server/apps/deck/tests/unit/Service/LabelServiceTest.php:95
ERRORS!
```

lib/Service/LabelService.php
```diff
@@ -166,6 +167,7 @@ public function update($id, $title, $color) {
 		$label = $this->find($id);
 		//$this->checkDuplicateTitle($label->getBoardId(), $title);
 		
+		$boardLabels = array();
```
juliushaertl: you will need to create a mock in the tests for the findAll method on labelMapper, so that this either returns an empty array or better (to properly test the code) contain an actual list of labels that will then be used in the foreach loop during the test.

lib/Service/LabelService.php
```diff
@@ -91,6 +91,18 @@ public function create($title, $color, $boardId) {
 		}
 
 		$this->permissionService->checkPermission(null, $boardId, Acl::PERMISSION_MANAGE);
+
+		//$this->checkDuplicateTitle($boardId, $title);
+		$boardLabels = $this->labelMapper->findAll($boardId);
+		if (is_array($boardLabels) || is_object($boardLabels)) {
```
juliushaertl: No need for is_object, $boardLabels will either be an array or null

lib/Service/LabelService.php
```diff
@@ -91,6 +91,18 @@ public function create($title, $color, $boardId) {
 		}
 
 		$this->permissionService->checkPermission(null, $boardId, Acl::PERMISSION_MANAGE);
+
+		//$this->checkDuplicateTitle($boardId, $title);
```
juliushaertl: Could you still move this to a separate method. Then we don't have duplicate code.

juliushaertl:
@jakobroehrl I rebase your changes and pushed a small commit to fix the codacy warnings. Let's get this in :+1:

Issue #780, opened 2018-12-08
in the left sidebar "delete board" generates 2 entries in timeline


cloud2018:
**Describe the bug**
When you delete a board with  the more-menue in the left sidebar in, two entries in the timeline are displayed.
When you delete a board in the main boardlist, just one entry in the timeline are displayed as expected

**To Reproduce**
1. In the left sidebar open 3-dots-menu and delete a board 
2. Restore the board
3. Open timeline of the board

**Expected behavior**
Only one entry ("Du hast das Board testboard gelöscht") should displayed 

**Screenshots**
![testboard deck nextcloud](https://user-images.githubusercontent.com/34692335/49685806-a443d400-faea-11e8-9511-085ab49c6759.png)

juliushaertl:
It seems that two delete requests are send to the server.

PR #816, opened 2018-12-20
Fix duplicate call to delete

juliushaertl:
* Resolves: #780
* Target version: master


weeman1337:
Does it in addition make sense to change the API, so deleting a deleted board isn't possible?

juliushaertl:
Yes, good point. Will adjust.

Issue #254, opened 2017-07-11
For icons, use Javascript tooltips which directly show


jancborchardt:
Currently the buttons like »Show archived cards«, »Board details«, »Remove due date«, »Formatting help« etc have tooltips, but they do not show directly.

They should use the JS tooltip function we use in Files and other apps too, which show a tooltip above the icon you are hovering.

PR #857, opened 2019-01-25
Use JS tooltips for board controls

jakobroehrl:

* Resolves: #254 
* Target version: master 

It's working but the icons now overlaps, how can I change it?
grafik

Is there a documentation that covers eg the tooltip function? I only copy and pasted.

### DIFF
--- a/templates/part.board.headerControls.php
+++ b/templates/part.board.headerControls.php
@@ -13,30 +13,19 @@
     </form>
 </div>
 
-<div class="section-label" data-toggle="tooltip" data-placement="bottom" title="<?php p($l->t('Show archived cards')); ?>">
-    <button ng-if="params.filter!='archive'" ng-click="switchFilter('archive')" title="<?php p($l->t('Show archived cards')); ?>">
-        <i class="icon icon-archive" style="opacity:0.5;"></i>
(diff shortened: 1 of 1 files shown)


templates/part.board.headerControls.php
```diff
@@ -13,19 +13,30 @@
...
-</button>
-<button ng-click="toggleCompactMode()" title="<?php p($l->t('Toggle compact mode')); ?>">
-	<i class="icon" ng-class="{ 'icon-toggle-compact-collapsed': compactMode, 'icon-toggle-compact-expanded': !compactMode }"></i>
-	<span class="hidden-visually"><?php p($l->t('Toggle compact mode')); ?></span>
-</button>
-<button ui-sref="board.detail({ id: id, tab: 0})"  title="<?php p($l->t('Show board details')); ?>">
-    <i class="icon icon-settings"></i>
-    <span class="hidden-visually"><?php p($l->t('Show board details')); ?></span>
-</button>
+<div class="section-label" data-toggle="tooltip" data-placement="bottom" title="<?php p($l->t('Show archived cards')); ?>">
```
juliushaertl: There should not be a wrapping div, you can also add the `data-toggle` and `data-placement` attributes directly to the `<button>` elements

jakobroehrl:
Now it's looking nice :)

PR #885, opened 2019-02-11
[nc16-only] Implementing fulltextsearch

daita:
This will make the app fail on NC13/14, this needs to be on a specific version for NC16 (but should also works for NC15)

It does not catch the search from the searchbar within the deck app, but allow to search from the fulltextsearch app.

## ToDo
- [x] Fix tests
- [x] Test on 15/16
- [ ] Waiting for nextcloud/server#14905 nextcloud/appstore#602

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -16,8 +16,8 @@ pipeline:
       - bash ./before_install.sh $APP_NAME $CORE_BRANCH $DB
       - cd ../server
       # Code checker
-      - ./occ app:check-code $APP_NAME -c strong-comparison
-      - ./occ app:check-code $APP_NAME -c deprecation
+      - ./occ app:check-code $APP_NAME -c strong-comparison || true
+      - ./occ app:check-code $APP_NAME -c deprecation || true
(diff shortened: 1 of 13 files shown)


daita:
selection_064

juliushaertl:
Thanks a lot @daita Looks good in general, will need to setup the fulltestsearch app to do some testing. Could you look into the test failures?

juliushaertl:
@daita Any chance you can have a look at the failing tests? :wink:

juliushaertl:
@daita Any news? I plan to push out a first beta at the end of the week, so it would be good to have this in there.

juliushaertl:
- [ ] ./occ app:check-code $APP_NAME -c strong-comparison
Invalid appinfo.xml file found: Element 'provider', attribute 'min-version': The attribute 'min-version' is not allowed.
- [ ] Add EventDispatcherInterface to unit tests

juliushaertl:
3ebe7949 needs to be reverted after https://github.com/nextcloud/server/pull/14905 has been merged

PR #966, opened 2019-03-29
Revert "Disable occ app:check-code"

juliushaertl:
Requires nextcloud/server#14905 nextcloud/appstore#602


juliushaertl:
Fixed with #977

PR #991, opened 2019-04-10
using fulltextsearchmanager::isAvailable()

daita:


daita:
would be good for the next beta of Nextcloud

lib/AppInfo/Application.php
```diff
@@ -191,17 +196,22 @@ protected function registerCollaborationResources() {
...
 	public function registerFullTextSearch() {
-		if (Util::getVersion()[0] < 16 || !\OC::$server->getAppManager()->isEnabledForUser('fulltextsearch')) {
+		if (Util::getVersion()[0] < 16) {
 			return;
 		}
 
 		$c = $this->getContainer();
 		try {
 			$this->fullTextSearchService = $c->query(FullTextSearchService::class);
+			$this->fullTextSearchManager = $c->query(FullTextSearchManager::class);
```
juliushaertl: Should be IFullTextSearchManager, FullTextSearchManager is not part of the public API

daita: Yeah, I changed like 3 times before getting it wrong...

PR #1049, opened 2019-05-16
returns string on null

daita:
should fix nextcloud/fulltextsearch#492


truelai:
This patch fixed the exception being thrown for me.

NC 16
Deck  0.6.1
FTS  1.3.1

PR #1114, opened 2019-07-03
Vue board sidebar sharing

jakobroehrl:
* Target version: vue
* Implements board sharing 
It's working in general. 
But the UI Change does not work. Maybe someone could help me with these 2 methods:
1. https://github.com/nextcloud/deck/blob/dfd33108accbc08fa4b480de19267c2589e93bd4/src/store/main.js#L172
2. https://github.com/nextcloud/deck/blob/dfd33108accbc08fa4b480de19267c2589e93bd4/src/store/main.js#L183

* Implements deleting/creating stacks
Why does the stack update API Call fails?
https://github.com/nextcloud/deck/blob/dfd33108accbc08fa4b480de19267c2589e93bd4/src/services/StackApi.js#L96

* Implements creating cards
Why does the card post API Call fails?
https://github.com/nextcloud/deck/blob/dfd33108accbc08fa4b480de19267c2589e93bd4/src/services/CardApi.js#L32


juliushaertl:
Nice progress :rocket: I'm really glad to have you on board for the vue migration.

> But the UI Change does not work. Maybe someone could help me with these 2 methods

You need to use Vue.set and Vue.delete in those cases, since otherwise Vue is not aware of the changes. For further details checkout https://vuejs.org/v2/guide/reactivity.html#Change-Detection-Caveats

> Why does the stack update API Call fails?

You are sending a nested object as data with the request. It should be `{id: 1}` instead of `{stack: {id: 1}}`:

		return axios.put(this.url(`/stacks/${stack.id}`), stack )

instead of 

https://github.com/nextcloud/deck/blob/dfd33108accbc08fa4b480de19267c2589e93bd4/src/services/StackApi.js#L97

> Why does the card post API Call fails?

This was caused by the card type missing. This is a currently unused parameter, so I pushed a commit to set it by default in the controller.

jakobroehrl:
@juliushaertl 
Thank you very much! Now I was able to fix all of my mentioned problems. 👍 
I'm glad to give someting back to this great project! 

Would you like to merge before I do the next changes for the vue migration?

juliushaertl:
Let me give it a quick review today. :wink:

src/components/board/Board.vue
```diff
@@ -26,7 +26,11 @@
 		<div v-if="board" class="board">
 			<container lock-axix="y" orientation="horizontal" @drop="onDropStack">
 				<draggable v-for="stack in stacksByBoard" :key="stack.id" class="stack">
-					<h3>{{ stack.title }}</h3>
+					<stack :stack="stack" />
+					<!-- <h3>{{ stack.title }}
```
juliushaertl: Moved the stack component, so we can remove it here :wink:

src/components/board/Board.vue
```diff
@@ -55,15 +59,17 @@
 import { Container, Draggable } from 'vue-smooth-dnd'
 import { mapState } from 'vuex'
 import Controls from '../Controls'
-import CardItem from '../cards/CardItem'
+/* import CardItem from '../cards/CardItem' */
```
juliushaertl: :fire:

src/components/board/Board.vue
```diff
@@ -55,15 +59,17 @@
...
+import Stack from './Stack'
 
 export default {
 	name: 'Board',
 	components: {
-		CardItem,
+		/* CardItem, */
```
juliushaertl: :fire:

juliushaertl: Same goes for the other commented code blocks :wink:

src/components/board/SharingTabSidebard.vue
```diff
@@ -33,24 +33,25 @@
...
+
+		<collection-list v-if="board.id" :id="&quot;board.id&quot;" :name="board.title"
+			type="deck" />
 	</div>
 </template>
 
 <script>
 import { Avatar, Multiselect } from 'nextcloud-vue'
-import CollaborationView from '../CollaborationView'
+import { CollectionList } from 'nextcloud-vue-collections'
```
juliushaertl: Let's delete the CollaborationView component then, as it is not needed any longer.

src/components/board/SharingTabSidebard.vue
```diff
@@ -33,24 +33,25 @@
 				<label for="checkbox">{{ t('deck', 'Manage') }}</label>
 
 				<button v-tooltip="t('deck', 'Delete')" class="icon-delete" @click="clickDeleteAcl(acl)" />
-
 			</li>
 		</ul>
-		<CollaborationView />
+
+		<collection-list v-if="board.id" :id="&quot;board.id&quot;" :name="board.title"
```
juliushaertl:
```suggestion
		<collection-list v-if="board.id" :id="`${board.id}`" :name="board.title"
```

src/components/board/SharingTabSidebard.vue
```diff
@@ -68,7 +69,24 @@ export default {
...
+		}),
+		unallocatedSharees() {
+			let ret = []
+
+			let allocatedSharees = []
+			for (var user in this.board.acl) {
+				allocatedSharees.push(this.board.acl[user].participant.uid)
+			}
+
+			this.sharees.forEach(function(sharee) {
```
juliushaertl: This could be a bit simplified by using the filter method https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter

Maybe also get rid of the additional allocatedSharees array by using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex to check if the sharee is part of the board acl list.

src/components/board/Stack.vue
```diff
@@ -0,0 +1,155 @@
+<!--
+  - @copyright Copyright (c) 2018 Julius Härtl <jus@bitgrid.net>
+  -
+  - @author Julius Härtl <jus@bitgrid.net>
```
juliushaertl: Don't forget to add you as an author :wink:

src/services/CardApi.js
```diff
@@ -0,0 +1,47 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+import axios from 'nextcloud-axios'
+
+export class CardApi {
+
+	url(url) {
```
juliushaertl: We probably should have a base Api class with that url method, but we can do that later.

src/components/board/SharingTabSidebard.vue
jakobroehrl: What das this means? Where can I read more about this?

juliushaertl: We require a string as an id of the collections list. This notation is basically a string cast, done by using template literals: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

jakobroehrl:
@juliushaertl 
Thanks for the review, I changed almost all of you points. 

Added further functions:

- Archiv/Unarchiv cards
- Archiv/Unarchiv Filter (is there a grey filter icon to show the user if the filter is active?)
- Card Details are shown now

Questions

- "Assign to me" throws an error: "Method Not Allowed". Do you have an idea what's the problem is?
https://github.com/nextcloud/deck/blob/7be7b3a5c8dca110de3346317ef4d445870f6e9f/src/services/CardApi.js#L77
- How could I get the "deleted items" and "timeline" for the boards?

Thanks!

juliushaertl:
> "Assign to me" throws an error: "Method Not Allowed". Do you have an idea what's the problem is?

Assign user needs to be a post request instead of put: https://github.com/nextcloud/deck/blob/1811dc73b76348e8bee76900fb09aee423ee104c/appinfo/routes.php#L65

> How could I get the "deleted items" and "timeline" for the boards?

- Deleted boards should already be handled by the sidebar, i think they are returned by the regular api endpoint
- Stacks: https://github.com/nextcloud/deck/blob/1811dc73b76348e8bee76900fb09aee423ee104c/appinfo/routes.php#L50
- Cards: https://github.com/nextcloud/deck/blob/1811dc73b76348e8bee76900fb09aee423ee104c/appinfo/routes.php#L58

The timeline data can be fetched though the activity apps api endpoint:
https://github.com/nextcloud/activity/blob/master/docs/endpoint-v2.md

You can have a look at the existing implementation here: https://github.com/nextcloud/deck/blob/40fd0c14016d30335098bd7141eff8456a488389/js/service/ActivityService.js#L95-L151

I'd say we get this in once the user assignment is fixed. :wink:

jakobroehrl:
Hi @juliushaertl 

- simplfied user filtering in the sharing tab
- changed the typ in the assign User API, but I does not work. Now I get a nextcloud page back. Can I "debug" these API calls? 
 - Maybe you could help me with the deleted items. I get the data, but I could not show it in the UI: https://github.com/nextcloud/deck/blob/a16101443a29862714d1634084c7dcd645f5c70f/src/components/board/DeletedTabSidebard.vue#L47 
mapState and deletedCards is not working, I can't get the data. 

Thanks!

jakobroehrl:
@juliushaertl 
https://github.com/nextcloud/deck/blob/vue/src/components/card/CardSidebar.vue#L24
How could I here add tags, assign user and due to fields? It should be appear before AppSidebarTab
Thanks!

juliushaertl:
> How could I here add tags, assign user and due to fields? It should be appear before AppSidebarTab

Seems we need to extend the vue components for that. The best would be to add a dedicated slot for that here: https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/AppSidebar/AppSidebar.vue#L70 Mind to open an issue over at the nextcloud-vue repo so we can discuss this?

juliushaertl:
Answered the rest via email :wink:

juliushaertl:
Enhancements from the nextcloud-vue library:
- https://github.com/nextcloud/nextcloud-vue/pull/536
- https://github.com/nextcloud/nextcloud-vue/pull/535

PR #1135, opened 2019-07-16
Adjust notifier to Nextcloud 17

nickvergessen:
Ref nextcloud/server#15040


MorrisJobke:
This needs a bump of the version requirement as well: https://github.com/nextcloud/deck/blob/master/appinfo/info.xml#L39

juliushaertl:
Master is now stable17 only (which will be released as version 0.7 then)

MorrisJobke:
> Master is now stable17 only (which will be released as version 0.7 then)

Feel free to merge then ;)

juliushaertl:
Yep, just waiting for CI :wink:

PR #1173, opened 2019-08-14
Open card sidebar

jakobroehrl:
* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


src/components/cards/CardItem.vue
```diff
@@ -133,8 +133,8 @@ export default {
 			this.copiedCard.newUserUid = this.card.owner.uid
 			this.$store.dispatch('assignCardToUser', this.copiedCard)
 		},
-		setCurrentCard() {
-			this.$store.dispatch('setCurrentCard', this.card)
+		openCardSidebar() {
```
juliushaertl: This is now a duplicate with https://github.com/nextcloud/deck/pull/1173/files#diff-f1b63cfb35147eb8827861656c1da442R104 so we should remove either one of both. 

I just tested the open though clicking the card (not the actions menu) that's why i mentioned earlier that I though it has already worked :wink:

PR #1175, opened 2019-08-16
changed time calculation

jakobroehrl:

* Resolves: # 
* Target version: vue

### DIFF
--- a/src/components/card/CardSidebar.vue
+++ b/src/components/card/CardSidebar.vue
@@ -108,8 +108,8 @@ export default {
 			return this.$store.getters.cardById(this.id)
 		},
 		subtitle() {
-			let lastModified = this.currentCard.lastModified
-			let createdAt = this.currentCard.createdAt
+			let lastModified = OC.Util.relativeModifiedDate(this.currentCard.lastModified * 1000)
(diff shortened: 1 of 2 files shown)


juliushaertl:
At some point we should also make sure the computed property is update every minute or so, since otherwise keeping the page open without any update will lead to outdated relative timestamps.

juliushaertl:
@jakobroehrl Mind to look into this as well? :)

jakobroehrl:
@juliushaertl: Yes, sure. Good idea. 
How can I retrigger the computed property?

src/components/card/CardSidebar.vue
```diff
@@ -133,21 +138,30 @@ export default {
...
 	watch: {
-		'currentCard': {
-			immediate: true,
-			handler() {
-				this.copiedCard = JSON.parse(JSON.stringify(this.currentCard))
-				this.allLabels = this.currentCard.labels
-				this.assignedUsers = this.currentCard.assignedUsers.map((item) => item.participant)
-				this.desc = this.currentCard.description
-			}
+		currentCard() {
```
juliushaertl: Any reason you removed the immediate property of the watcher? Iirc that was needed to properly initialize the data on initial load.

jakobroehrl: readded

Issue #1138, opened 2019-07-18
Add description from curl POST request doesn't work


a11exandru:
I tried to add a new card via API with curl POST request <curl -u username:password  -X PUT -H "OCS-APIRequest:true" -H "Content-Type:application/json" "https://nextcloud.local/index.php/apps/deck/api/v1.0/
boards/9/stacks/27/cards/132?title=thisistitle&description=thisisthedescription&type=plain&order=999&duedate=null">  but the description remains empty even if I do add the description attribute.

putt1ck:
Does the new card get added, just no description?

a11exandru:
Yes, @putt1ck !

putt1ck:
@juliushaertl is anyone leading on API stuff at the moment?

juliushaertl:
Yes, the initial api implementation only allows to set the title, type and order (https://deck.readthedocs.io/en/latest/API/#cards) However it of course would make sense to also set the other details on creation. cc @desperateCoder since we discussed this as well for the android app.

@putt1ck I try to catch up on issues that come up, but my time for deck is unfortunately rather limited at the moment. Contributions are always very welcome.

juliushaertl:
Some code hints if someone wants to look into implementing this:

https://github.com/nextcloud/deck/blob/master/lib/Controller/CardApiController.php#L76
https://github.com/nextcloud/deck/blob/master/lib/Service/CardService.php#L158-L192

putt1ck:
@juliushaertl the hints should help @a11exandru fix his own bug report :)

a11exandru:
Thanks @juliushaertl ! I will look into it.

desperateCoder:
@a11exandru @juliushaertl  that'd be great, that's where i'm stuck right now. Thanks in advance!

PR #1182, opened 2019-08-19
Able to set description on creation card via API.

a11exandru:
Signed-off-by: Alexandru Puiu <alexpuiu20@yahoo.com>

* Resolves: #1138 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/lib/Controller/CardApiController.php
+++ b/lib/Controller/CardApiController.php
@@ -2,6 +2,8 @@
 /**
  * @copyright Copyright (c) 2018 Ryan Fletcher <ryan.fletcher@codepassion.ca>
  *
+ * @copyright Copyright (c) 2019, Alexandru Puiu (alexpuiu20@yahoo.com)
+ *
  * @author Ryan Fletcher <ryan.fletcher@codepassion.ca>
  *
  * @license GNU AGPL version 3 or any later version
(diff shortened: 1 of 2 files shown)


lib/Controller/CardApiController.php
```diff
@@ -70,11 +72,12 @@ public function get() {
 	 * @params $title
 	 * @params $type
 	 * @params $order
+	 * @params $description
 	 *
 	 * Get a specific card.
 	 */
-	public function create($title, $type = 'plain', $order = 999) {
-		$card = $this->cardService->create($title, $this->request->getParam('stackId'), $type, $order, $this->userId);
+	public function create($title, $type = 'plain', $order = 999, $description) {
```
juliushaertl: This should have an empty default value

lib/Service/CardService.php
```diff
@@ -594,7 +597,7 @@ public function assignUser($cardId, $userId) {
 			/* Notifyuser about the card assignment */
 			$this->notificationHelper->sendCardAssigned($card, $userId);
 		}
-
+
```
juliushaertl: Please remove the tabs in empty lines

lib/Controller/CardApiController.php
a11exandru: Ok, it's done!

lib/Service/CardService.php
a11exandru: Done

PR #1188, opened 2019-08-22
members are shown now

jakobroehrl:
* Resolves: # 
* Target version: vue


src/components/boards/BoardItem.vue
```diff
@@ -30,6 +30,8 @@
 		<div class="board-list-title-cell">{{ board.text }}</div>
 		<div class="board-list-avatars-cell">
 			<avatar :user="board.owner.uid" class="board-list-avatar" />
+			<avatar v-for="user in board.board.acl" :key="user.id" :user="user.participant.uid"
```
juliushaertl: The underlying issue is that the filteredBoards getter maps to a menu item object, which is not needed anymore: https://github.com/nextcloud/deck/blob/ac9fef5f054a8ede6b76f9d66b02f3d4044f4ead/src/store/main.js#L89

juliushaertl: Just returning boards should be fine here, then we can use `board.acl`

jakobroehrl: @juliushaertl 
If I change main.js L89 to return boards
I can use board.acl but board.title and board.acl are gone

juliushaertl: I quickly pushed a commit to fix it. Works just fine, but the text is now named title as in the board object.

PR #1196, opened 2019-08-25
Add date format to If-Modified-Since documentation

monokles:
* Resolves: # 
* Target version: master 

### Summary
Fixes an error in the curl example in API.md, and adds a small list with supported date formats for the If-Modified-Since headers.

See https://github.com/sabre-io/http/blob/4.2/lib/functions.php#L16 for the definition of the invoked function.

### Checklist

- [X] Tests (unit, integration, api and/or acceptance) are included

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -52,7 +52,7 @@ Example curl request:
 curl -u admin:admin -X GET \
     'http://localhost:8000/index.php/apps/deck/api/v1.0/boards/2/stacks' \
     -H "OCS-APIRequest: true" \
-    -H "If-Modified-Since: Mon, 5 Nov 2018 09:28:00 GMT"
+    -H "If-Modified-Since: Mon, 05 Nov 2018 09:28:00 GMT"
 ```
 
 # Endpoints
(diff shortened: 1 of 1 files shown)


docs/API.md
```diff
@@ -45,14 +45,19 @@ In any case a user doesn't have access to a requested entity, a 403 error will b
 ### If-Modified-Since
 
 Some index endpoints support limiting the result set to entries that have been changed since the given time.
+The supported date formats are:
+
+* IMF-fixdate:      `Sun, 03 Aug 2019 10:34:12 GMT`
```
juliushaertl: Could you try to highlight tat IMF-fixdate should be used (the others are considered obsolete) according to https://tools.ietf.org/html/rfc7231#section-7.1.1.1

monokles: Done! See [this commit](https://github.com/nextcloud/deck/pull/1196/commits/b7a36851de82542f7634e68356a22bc96cdf4f20).

juliushaertl:
Thanks a lot :+1:

PR #1210, opened 2019-08-29
label multi shows color, sharing shows avatars

jakobroehrl:
* Resolves: 
Label multiselect should also show the color
[SharingTab] [CardDetailsTab] User multiselect should show avatars as well
[SharingTab] List of acl entries should show a group icon if the acl entry is of type group

* Target version: vue

### DIFF
--- a/src/components/board/SharingTabSidebard.vue
+++ b/src/components/board/SharingTabSidebard.vue
@@ -1,12 +1,21 @@
 <template>
 	<div>
-		<multiselect v-model="addAcl" :options="unallocatedSharees" label="label"
+		<!-- <multiselect v-model="addAcl" :options="unallocatedSharees" label="label"
 			@input="clickAddAcl" @search-change="asyncFind">
 			<template #option="scope">
 				<avatar :user="scope.option.value.shareWith" />
 				<span class="avatarLabel">{{ scope.option.label }} </span>
 			</template>
-		</multiselect>
+		</multiselect> -->
+
+		<multiselect
+			v-model="addAcl"
+			:options="formatedSharees"
+			:user-select="true"
+			label="displayName"
+			track-by="user"
+			@input="clickAddAcl"
+			@search-change="asyncFind" />
 
 		<ul
 			id="shareWithList"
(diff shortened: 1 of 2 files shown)


src/components/board/SharingTabSidebard.vue
```diff
@@ -3,7 +3,8 @@
 		<multiselect v-model="addAcl" :options="unallocatedSharees" label="label"
 			@input="clickAddAcl" @search-change="asyncFind">
 			<template #option="scope">
-				{{ scope.option.label }}
+				<avatar :user="scope.option.value.shareWith" />
+				<span class="avatarLabel">{{ scope.option.label }} </span>
```
juliushaertl: You could just use the User layout option of our multiselect component for this:
https://nextcloud-vue-components.netlify.com/#/Components?id=avatarselectoption

Then we also have nice icons to differentiate between users, groups or circles 
image

src/components/board/SharingTabSidebard.vue
```diff
@@ -71,6 +81,17 @@ export default {
 		...mapGetters({
 			sharees: 'sharees'
 		}),
+		formatedSharees() {
+			return this.unallocatedSharees.map(item => {
+				return {
+					user: item.label,
+					displayName: item.label,
+					// desc: 'desc',
+					icon: item.value.shareType === 1 ? 'icon-group' : 'icon-user',
```
juliushaertl: would be good to also handle `7` which is a share to a circle

src/components/board/SharingTabSidebard.vue
```diff
@@ -1,11 +1,21 @@
 <template>
 	<div>
-		<multiselect v-model="addAcl" :options="unallocatedSharees" label="label"
+		<!-- <multiselect v-model="addAcl" :options="unallocatedSharees" label="label"
```
juliushaertl: Should be removed before merging

src/components/board/SharingTabSidebard.vue
```diff
@@ -71,6 +81,17 @@ export default {
 		...mapGetters({
 			sharees: 'sharees'
 		}),
+		formatedSharees() {
+			return this.unallocatedSharees.map(item => {
+				return {
+					user: item.label,
+					displayName: item.label,
+					// desc: 'desc',
```
juliushaertl: should be removed before merging

PR #1213, opened 2019-08-29
active is highlighted

jakobroehrl:
questions:
- What is the right grey tone?
- If I open a card and reload the page the card is not highlighted after the reload. What could that be?

Thanks!

### DIFF
--- a/src/components/cards/CardItem.vue
+++ b/src/components/cards/CardItem.vue
@@ -21,7 +21,7 @@
   -->
 
 <template>
-	<div :class="{'compact': compactMode}" tag="div" class="card"
+	<div :class="{'compact': compactMode, 'current-card': currentCard}" tag="div" class="card"
 		@click.self="openCard">
 		<div class="card-upper">
 			<h3 @click.stop="startEditing(card)">{{ card.title }}</h3>
@@ -98,6 +98,9 @@ export default {
 					color: this.textColor(label.color)
 				}
 			}
+		},
+		currentCard() {
+			return this.$route.params.cardId === this.id
 		}
 	},
 	methods: {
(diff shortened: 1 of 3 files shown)


src/components/cards/CardItem.vue
```diff
@@ -148,6 +151,10 @@ export default {
 		opacity: 0;
 	}
 
+	.current-card {
+		box-shadow: 0 0 3px 0 grey !important;
```
juliushaertl: I'd say we use the same color but a bit bigger shadow:

```suggestion
		box-shadow: 0 0 5px 0 var(--color-box-shadow);
```

juliushaertl:
> If I open a card and reload the page the card is not highlighted after the reload. What could that be?

When reloading the param seems to end up as a string while when navigating though clicking the parameter is an integer value

src/components/cards/CardItem.vue
jakobroehrl: I need the important Expression otherwise it won't work
Could we go up to 6px for better recognition?

jakobroehrl:
> 
> 
> > If I open a card and reload the page the card is not highlighted after the reload. What could that be?
> 
> When reloading the param seems to end up as a string while when navigating though clicking the parameter is an integer value

@juliushaertl 
That's odd. I can retrace it. Should I change the comparison here to ==
https://github.com/nextcloud/deck/blob/d08bd23578ea4c05e27d3744094ee8f6e07ad356/src/components/cards/CardItem.vue#L103

src/components/cards/CardItem.vue
juliushaertl: 6px is also fine for me
try to move it into the actual .card section and use &.current-card to apply the class selector on top of that, so it gets applied after the regular card style in line 159

jakobroehrl:
@juliushaertl 
This is still open "Should I change the comparison here to =="
The rest is working

PR #1217, opened 2019-09-03
avatars now in a good looking way

jakobroehrl:
* Resolves: # 567
* Target version: vue

Avatars are really looking good now :)

### DIFF
--- a/src/components/cards/CardBadges.vue
+++ b/src/components/cards/CardBadges.vue
@@ -111,9 +111,6 @@ export default {
 		display: flex;
 		flex-grow: 1;
 
-		/* & > div {
-			margin-right: 10px;
-		} */
 		.icon {
 			opacity: 0.5;
 			padding: 12px 3px;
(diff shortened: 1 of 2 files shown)


juliushaertl:
@jakobroehrl In general, a screenshot for visual changes will always make @jancborchardt or the other @nextcloud/designers happy so he can check it out as well :wink:

juliushaertl:
- Please rename the classes as session list doesn't really fit the use case here
- The 3-dots icon should be aligned with the regular one
image
- The popovermenu is not positioned properly
image


Besides that it looks really nice now :)

src/components/cards/CardBadges.vue
```diff
@@ -36,17 +36,16 @@ cardbadges
 
 		<div v-if="card.attachments" class="card-files icon icon-files-dark" />
 
-		<div v-if="card.assignedUsers" class="card-assigned-users">
-			<avatar v-for="user in card.assignedUsers" :key="user.id" :user="user.participant.primaryKey" />
-		</div>
+		<avatar-list :sessions="card.assignedUsers" />
```
juliushaertl: lets call this users

src/components/cards/CardBadges.vue
```diff
@@ -112,9 +111,9 @@ export default {
 		display: flex;
 		flex-grow: 1;
 
-		& > div {
+		/* & > div {
```
juliushaertl: Please remove code that is no longer needed

juliushaertl: Pushed a commit to fix.

PR #1219, opened 2019-09-04
tags are now in one line

jakobroehrl:


src/components/cards/CardItem.vue
```diff
@@ -41,9 +41,9 @@
 
 			</transition>
 		</div>
-		<ul class="labels" @click="openCard">
```
juliushaertl: No need to change the structure, a unordered list is the better fit here I'd say.

PR #1235, opened 2019-09-10
Comments

jakobroehrl:
* Target version: vue

Todos
- [x] There still needs to be incremental fetching of more than 21 comments.
- [x] mention a user. @juliushaertl How should I solve the user mention? 
Btw: I tested the mention with the files app. There it is possible to mention all users even if they can't  access the file. We should only mention accessible users, what do you think?


src/store/comment.js
```diff
@@ -0,0 +1,71 @@
...
+
+const apiClient = new CommentApi()
+
+export default {
+	state: {
+		comments: []
+	},
+	mutations: {
+		listComments(state, comment) {
+			state.comments = []
```
juliushaertl: I think for comments it would be nice to keep them even for multiple cards in the store. Otherwise we need to refetch them every time the user switches though cards.

src/store/comment.js
```diff
@@ -0,0 +1,71 @@
...
+			state.comments = comment
+		},
+		updateComment(state, comment) {
+			let existingIndex = state.comments.findIndex(_comment => _comment.id === comment.commentId)
+			if (existingIndex !== -1) {
+				state.comments[existingIndex].message = comment.comment
```
juliushaertl: use Vue.set

src/services/CommentApi.js
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	listComments(cardId) {
+		return axios({
+			method: 'REPORT',
+			url: this.url(`${cardId}`),
+			data: `<?xml version="1.0" encoding="utf-8" ?>
+			<oc:filter-comments xmlns:D="DAV:" xmlns:oc="http://owncloud.org/ns">
+				<oc:limit>21</oc:limit>
+				<oc:offset>0</oc:offset>
```
juliushaertl: There still needs to be incremental fetching of more than 21 comments.

src/store/comment.js
```diff
@@ -0,0 +1,83 @@
...
+
+const apiClient = new CommentApi()
+
+export default {
+	state: {
+		comments: {}
+	},
+	mutations: {
+		listComments(state, commentObj) {
+			state.comments[commentObj.cardId] = commentObj.comments
```
juliushaertl: Use Vue.set

src/store/comment.js
```diff
@@ -0,0 +1,83 @@
...
+		},
+		createComment(state, newComment) {
+			state.comments[newComment.cardId].push(newComment)
```
juliushaertl: Make sure to check if `state.comments[newComment.cardId]` actually exists. If the listComments  mutation has never been called before, this will fail.

src/store/comment.js
```diff
@@ -0,0 +1,83 @@
...
+import xmlToTagList from '../helpers/xml'
+
+const apiClient = new CommentApi()
+
+export default {
+	state: {
+		comments: {}
+	},
+	mutations: {
+		listComments(state, commentObj) {
```
juliushaertl: listComments sounds a bit confusing here as the mutiation adds a list of comments.

src/store/comment.js
```diff
@@ -0,0 +1,83 @@
...
+			if (existingIndex !== -1) {
+				state.comments[comment.cardId].splice(existingIndex, 1)
+			}
+		}
+	},
+	actions: {
+		listComments({ commit }, card) {
+			apiClient.listComments(card.id)
+				.then((comments) => {
+					let commentsJson = xmlToTagList(comments)
```
juliushaertl: Please use const if the variable is not reassigned

juliushaertl:
> How should I solve the user mention?
> Btw: I tested the mention with the files app. There it is possible to mention all users even if they can't access the file. We should only mention accessible users, what do you think?

Since we have the list of "mentionable" users available i would suggest to use that and limit the mentioning

For the implementation, we don't need to do any special formatting, the comment parsing will be done by the backend. Maybe we can just use https://tiptap.scrumpy.io/suggestions for the frontend. Either in this PR or in a separate one, up to you :wink:

jakobroehrl:
@juliushaertl 
If I change here https://github.com/nextcloud/deck/blob/d54cfad49fa1818145b88f8eb5c0252ebbea6374/src/components/card/CommentsTabSidebar.vue#L107-L112

to this.currentBoard.users then the this.currentBoard is undefined. Why is that? I can print out this.currentBoard.users properly, thank you.

jakobroehrl:
@juliushaertl Comments are working expect some polish, what do you think?

src/components/card/CommentItem.vue
```diff
@@ -0,0 +1,87 @@
+<template>
+	<li>
+		<hr>
+		<form v-if="edit" @submit.prevent="updateComment">
+			<input v-model="commentMsg" type="text" autofocus
```
juliushaertl: This should also use the rich editor with mentions. Please move that to a separate component so it can be reused.

src/components/card/CommentItem.vue
```diff
@@ -0,0 +1,87 @@
...
+				required>
+			<input v-tooltip="t('deck', 'Save')" class="icon-confirm" type="submit"
+				value="">
+			<input type="submit" value="" class="icon-close"
+				@click.stop.prevent="hideUpdateForm">
+		</form>
+
+		<template v-else>
+			<avatar :user="comment.uId" />
+			<span>{{ comment.uId }}</span>
```
juliushaertl: Should be the display name

src/components/card/CommentItem.vue
```diff
@@ -0,0 +1,87 @@
...
+			<Actions @click.stop.prevent>
+				<ActionButton icon="icon-rename" @click="showUpdateForm()">{{ t('deck', 'Update') }}</ActionButton>
+				<ActionButton icon="icon-delete" @click="deleteComment(comment.id)">{{ t('deck', 'Delete') }}</ActionButton>
+			</Actions>
+			<span v-html="comment.message" />
```
juliushaertl: v-html should not be used for unsanitized user input

src/components/card/CommentsTabSidebar.vue
```diff
@@ -0,0 +1,320 @@
+<template>
+	<div>
+		<div id="userDiv">
+			<avatar :user="card.owner.uid" />
+			<span class="has-tooltip username">
+				{{ card.owner.displayname }}
+			</span>
+		</div>
+
+		<!-- <div id="commentForm">
```
juliushaertl: :wastebasket:

src/components/card/CommentsTabSidebar.vue
```diff
@@ -0,0 +1,320 @@
...
+					:key="user.uid"
+					:class="{ 'is-selected': navigatedUserIndex === index }"
+					class="suggestion-list__item"
+					@click="selectUser(user)"
+				>
+					{{ user.displayname }}
+				</div>
+			</template>
+			<div v-else class="suggestion-list__item is-empty">
+				No users found
```
juliushaertl: Make this translatable

src/components/card/CommentsTabSidebar.vue
```diff
@@ -0,0 +1,320 @@
...
+								threshold: 0.2,
+								keys: ['uid', 'displayname']
+							})
+							return fuse.search(query)
+						}
+					})
+				],
+				content: '',
+				onUpdate: ({ getHTML }) => {
+					this.newComment = getHTML().replace(/(<p>|<\/p>)/g, '')
```
juliushaertl: Please make sure all HTML is removed not just p tags and we get the plain text

jakobroehrl: Is there a easy way or just so with replace?

src/components/card/CommentsTabSidebar.vue
```diff
@@ -0,0 +1,320 @@
...
+			}
+		}
+	},
+	created() {
+	},
+
+	methods: {
+		loadComments() {
+			this.isLoading = true
+			this.card.limit = this.limit
```
juliushaertl: I'm not a fan of this approach, the limit should be moved in the store itself. The offset should also be kept in the store, so the component doesn't need to care about it. Also we need some logic to handle updating for newer comments

src/store/comment.js
```diff
@@ -0,0 +1,93 @@
...
+export default {
+	state: {
+		comments: {}
+	},
+	mutations: {
+		addComments(state, commentObj) {
+			if (state.comments[commentObj.cardId] === undefined) {
+				Vue.set(state.comments, commentObj.cardId, commentObj.comments)
+			} else {
+				// FIXME append comments once incremental fetching is implemented
```
juliushaertl: This needs some more adjustments, The Vue.set was just a way to avoid duplicate entries when testing. We need to add elements to the existing list, but only if they are not already part of it.

src/store/comment.js
```diff
@@ -0,0 +1,93 @@
...
+				// state.comments[commentObj.cardId].push(...commentObj.comments)
+				Vue.set(state.comments, commentObj.cardId, commentObj.comments)
+			}
+		},
+		createComment(state, newComment) {
+			if (state.comments[newComment.cardId] === undefined) {
+				state.comments[newComment.cardId] = []
```
juliushaertl: Use Vue.set

src/store/comment.js
```diff
@@ -0,0 +1,93 @@
...
+			}
+			state.comments[newComment.cardId].push(newComment)
+		},
+		updateComment(state, comment) {
+			let existingIndex = state.comments[comment.cardId].findIndex(_comment => _comment.id === comment.commentId)
+			if (existingIndex !== -1) {
+				state.comments[comment.cardId][existingIndex].message = comment.comment
```
juliushaertl: Vue.set

src/components/card/CommentsTabSidebar.vue
juliushaertl:
```
const element = document.getElementById(getHTML())
this.newComment = element.innerText || element.textContent
```
should do the trick

juliushaertl:
For the UI polishing this should look like comments in the files app:

image

src/store/comment.js
jakobroehrl: How can I use set with these complex command?
state.comments[comment.cardId][existingIndex].message = comment.comment

This is not working
```
Vue.set(state.comments, comment.cardId.existingIndex.message, comment.comment)
Vue.set(state.comments, comment.cardId[existingIndex].message, comment.comment)
Vue.set(state.comments[comment.cardId][existingIndex], message, comment.comment)
```

src/store/comment.js
jakobroehrl: How can I use set with these complex command?
state.comments[comment.cardId][existingIndex].message = comment.comment

This is not working
```
Vue.set(state.comments, comment.cardId.existingIndex.message, comment.comment)
Vue.set(state.comments, comment.cardId[existingIndex].message, comment.comment)
Vue.set(state.comments[comment.cardId][existingIndex], message, comment.comment)
```

juliushaertl: https://vuejs.org/v2/api/#Vue-set

jakobroehrl:
Todo
- [ ] Edit Comment should also use the rich editor with mentions

src/store/comment.js
```diff
@@ -60,7 +60,15 @@ export default {
 		}
 	},
 	actions: {
-		listComments({ commit }, card) {
+		listComments({ commit, state }, card) {
+			card.limit = 5
```
juliushaertl: no need to add the limit to the card object

src/store/comment.js
```diff
@@ -60,7 +60,15 @@ export default {
...
+			card.offset = 0
```
juliushaertl: I'd prefer to have an object in the comments store, that stores the actual offset. That way the regular card object is not polluted with metadata.

src/store/comment.js
```diff
@@ -60,7 +60,15 @@ export default {
...
+			if (state.comments[card.id] !== undefined) {
+				if (state.comments[card.id].length === card.limit) {
+					card.offset = card.limit
```
juliushaertl: wouldn't this only allow fetching the comments once? What if you have >10 comments.

PR #1242, opened 2019-09-13
move card to another board

jakobroehrl:
@juliushaertl 
How can I retrieve the stacks from another board?
In the stacks store are only the stacks from the current board.
Thanks!

### DIFF
--- a/src/components/cards/CardItem.vue
+++ b/src/components/cards/CardItem.vue
@@ -37,9 +37,18 @@
 					<ActionButton icon="icon-archive" @click="archiveUnarchiveCard()">{{ t('deck', (showArchived ? 'Unarchive card' : 'Archive card')) }}</ActionButton>
 					<ActionButton v-if="showArchived === false" icon="icon-delete" @click="deleteCard()">{{ t('deck', 'Delete card') }}</ActionButton>
 					<ActionButton icon="icon-settings-dark" @click="openCard">{{ t('deck', 'Card details') }}</ActionButton>
+					<ActionButton icon="icon-settings-dark" @click.stop="modalShow=true">{{ t('deck', 'Move card') }}</ActionButton>
 				</Actions>
 
 			</transition>
+			<modal v-if="modalShow" title="Move card to another board" @close="modalShow=false">
+				<div class="modal__content">
+					<Multiselect v-model="selectedBoard" :options="boards" label="title" />
+					{{ stacksBySelectedBoard }}
+
+				</div>
+			</modal>
+
 		</div>
 		<ul class="labels" @click="openCard">
 			<li v-for="label in card.labels" :key="label.id" :style="labelStyle(label)"><span>{{ label.title }}</span></li>
@@ -51,9 +60,10 @@
 </template>
 
 <script>
-import { PopoverMenu } from 'nextcloud-vue'
+import { Modal } from 'nextcloud-vue/dist/Components/Modal'
 import { Actions } from 'nextcloud-vue/dist/Components/Actions'
 import { ActionButton } from 'nextcloud-vue/dist/Components/ActionButton'
+import { Multiselect } from 'nextcloud-vue/dist/Components/Multiselect'
 import ClickOutside from 'vue-click-outside'
 import { mapState } from 'vuex'
 
@@ -63,7 +73,7 @@ import Color from '../../mixins/color'
 
 export default {
 	name: 'CardItem',
-	components: { PopoverMenu, CardBadges, LabelTag, Actions, ActionButton },
+	components: { Modal, CardBadges, LabelTag, Actions, ActionButton, Multiselect },
 	directives: {
 		ClickOutside
 	},
@@ -78,16 +88,31 @@ export default {
 		return {
 			menuOpened: false,
 			editing: false,
-			copiedCard: ''
+			copiedCard: '',
+			modalShow: false,
+			selectedBoard: '',
+			selectedStack: ''
 		}
 	},
 	computed: {
 		...mapState({
 			compactMode: state => state.compactMode,
-			showArchived: state => state.showArchived
+			showArchived: state => state.showArchived,
+			currentBoard: state => state.currentBoard
 		}),
 		card() {
 			return this.$store.getters.cardById(this.id)
+		},
+		boards() {
+			return this.$store.getters.boards.filter(board => {
+				return board.id !== this.currentBoard.id
+			})
+		},
+		stacksBySelectedBoard() {
+			if (this.selectedBoard === '') {
+				return []
+			}
+			return this.$store.getters.stacksByBoard(this.selectedBoard.id)
 		},
 		menu() {
 			return []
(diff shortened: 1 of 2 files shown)


juliushaertl:
Hm, so either we refactor the store to store all stacks and just return the ones related to the current board id or we just do a temporary api request to fetch the stacks from the foreign board in the modal component when trying to move. I think the second approach would be fine for now, since we do not really need the stack list for any other part of the app.

jakobroehrl:
@juliushaertl 
Tried to implement the API call. 
Do you see a problem with that? I can't get the Stacks to the UI, but the data is there, proved with a console.log, thanks!

jakobroehrl:
@skjnldsv 
Could you check this please?
https://github.com/nextcloud/deck/blob/27b8c79bcab7c871f747ef89488cf9bb3c1196dd/src/components/cards/CardItem.vue#L114-L134

I don't get the stacks on the UI, but a console.log in line 132 shows me the values. 
Do you have an idea?

Thanks!

src/components/cards/CardItem.vue
```diff
@@ -78,17 +91,47 @@ export default {
...
+				return board.id !== this.currentBoard.id
+			})
+		},
+		stacksBySelectedBoard() {
+			if (this.selectedBoard === '') {
+				return []
+			}
+
+			let url = OC.generateUrl('/apps/deck/stacks/' + this.selectedBoard.id)
+			axios.get(url)
```
skjnldsv: Use try...catch instead of chaining .then and .catch :)
With async/await
This will help you figuring things out :)

skjnldsv: Also, you should not want to use a request in a computed.
Either load it once on beforeMount and store the results in a local data, or use the dedicated multiselect https://vue-multiselect.js.org/#sub-asynchronous-select

juliushaertl:
@jakobroehrl Would be cool if you could try assigning labels to represent the state of the pr, that way it is easier to see in the overview if they need some attention for reviewing. :wink:

jakobroehrl:
@juliushaertl Could you review please and merge if everything is fine?

src/components/cards/CardItem.vue
```diff
@@ -37,9 +37,25 @@
 					<ActionButton icon="icon-archive" @click="archiveUnarchiveCard()">{{ t('deck', (showArchived ? 'Unarchive card' : 'Archive card')) }}</ActionButton>
 					<ActionButton v-if="showArchived === false" icon="icon-delete" @click="deleteCard()">{{ t('deck', 'Delete card') }}</ActionButton>
 					<ActionButton icon="icon-settings-dark" @click="openCard">{{ t('deck', 'Card details') }}</ActionButton>
+					<ActionButton icon="icon-settings-dark" @click.stop="modalShow=true">{{ t('deck', 'Move card') }}</ActionButton>
```
juliushaertl: Let's use the icon external for this. Also card details should always be the last item i think

src/components/cards/CardItem.vue
```diff
@@ -37,9 +37,25 @@
...
 				</Actions>
 
 			</transition>
+			<modal v-if="modalShow" title="Move card to another board" @close="modalShow=false">
+				<div class="modal__content">
+					<Multiselect v-model="selectedBoard" :options="boards" label="title"
+						@select="loadStacksFromBoard" />
+					<Multiselect v-model="selectedStack" :options="stacksFromBoard" label="title" />
```
juliushaertl: Please add proper labels to the multiselects, otherwise users don't know what they actually need to do.

src/components/cards/CardItem.vue
```diff
@@ -37,9 +37,25 @@
...
+					<button @click="moveCard">
+						<span class="icon-confirm" />{{ t('deck', 'Move card') }}
```
juliushaertl: Some design input on the buttons:
- Right-align them
- Change the order so that the Mvoe card action is the one to the most right
- Make the move card button primary
- No need for icons here I'd say

src/components/cards/CardItem.vue
```diff
@@ -36,10 +36,25 @@
...
 					<ActionButton icon="icon-archive" @click="archiveUnarchiveCard()">{{ t('deck', (showArchived ? 'Unarchive card' : 'Archive card')) }}</ActionButton>
 					<ActionButton v-if="showArchived === false" icon="icon-delete" @click="deleteCard()">{{ t('deck', 'Delete card') }}</ActionButton>
+					<ActionButton icon="icon-external" @click.stop="modalShow=true">{{ t('deck', 'Move card') }}</ActionButton>
 					<ActionButton icon="icon-settings-dark" @click="openCard">{{ t('deck', 'Card details') }}</ActionButton>
 				</Actions>
 
 			</transition>
+			<modal v-if="modalShow" title="Move card to another board" @close="modalShow=false">
+				<div class="modal__content">
+					<Multiselect :placeholder="t('deck', 'select board')" v-model="selectedBoard" :options="boards"
```
juliushaertl:
```suggestion
					<Multiselect :placeholder="t('deck', 'Select a board')" v-model="selectedBoard" :options="boards"
```

src/components/cards/CardItem.vue
```diff
@@ -36,10 +36,25 @@
...
+						label="title"
+						@select="loadStacksFromBoard" />
+					<Multiselect :placeholder="t('deck', 'select stack')" v-model="selectedStack" :options="stacksFromBoard"
```
juliushaertl:
```suggestion
					<Multiselect :placeholder="t('deck', 'Select a stack')" v-model="selectedStack" :options="stacksFromBoard"
```

src/components/cards/CardItem.vue
```diff
@@ -248,4 +296,13 @@ export default {
 			color: transparent;
 		}
 	}
+
+	.modal__content {
+		width: 25vw;
```
juliushaertl: A proper max-width/max-height would be good for small screen sizes

src/components/cards/CardItem.vue
```diff
@@ -248,4 +296,13 @@ export default {
...
+		text-align: center;
+		margin: 4vw 1vw;
```
juliushaertl:
```suggestion
		margin: 20px 20px 80px 20px;
```

src/components/cards/CardItem.vue
```diff
@@ -248,4 +296,13 @@ export default {
...
+	}
+	.modal__content button{
```
juliushaertl: Maybe add 
```
.multiselect {
    margin-bottom: 10px;
}
```

src/components/cards/CardItem.vue
```diff
@@ -300,9 +300,14 @@ export default {
...
 		width: 25vw;
 		text-align: center;
-		margin: 4vw 1vw;
+		margin: 20px 20px 60px 20px;
 	}
+
 	.modal__content button{
 		float: right;
+
+		.multiselect {
```
juliushaertl: .multiselect is not inside the button ;)

jakobroehrl: done

PR #1243, opened 2019-09-13
Add circles and groups to cards

jakobroehrl:
todo:
Add them to cards, too

### DIFF
--- a/src/components/board/SharingTabSidebard.vue
+++ b/src/components/board/SharingTabSidebard.vue
@@ -87,10 +87,11 @@ export default {
 					sharee.isNoUser = true
 				}
 				if (item.value.shareType === 7) {
-					sharee.icon = 'icon-circle'
+					sharee.icon = 'icon-circles'
 					sharee.isNoUser = true
 				}
 
+				sharee.value = item.value
 				return sharee
 			})
 		},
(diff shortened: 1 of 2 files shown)


jakobroehrl:
@juliushaertl 
At the moment alle users from a board are shown in the Card Sidebar for sharing. They come from here:
https://github.com/nextcloud/deck/blob/vue/src/components/board/Board.vue#L92-L97

Should I add circles and groups to board.users on the BoardController?

juliushaertl:
> Should I add circles and groups to board.users on the BoardController?

I just found that we still miss to add them in the backend. Once we fetch the list of circle members here https://github.com/nextcloud/deck/blob/29f6ae477bdc7f0530cb37d47033705587a8eb4b/lib/Service/PermissionService.php#L255 it should just work™

juliushaertl:
> Should I add circles and groups to board.users on the BoardController?

I just found that we still miss to add them in the backend. Once we fetch the list of circle members here https://github.com/nextcloud/deck/blob/29f6ae477bdc7f0530cb37d47033705587a8eb4b/lib/Service/PermissionService.php#L255 it should just work™

juliushaertl:
Anyway let's fix this in a separate PR.

PR #1257, opened 2019-09-18
Unify reorder results to always be an array

juliushaertl:
@desperateCoder @stefan-niedermann


desperateCoder:
Nice! Thanks! Please tell us as soon as there is an update!

PR #1293, opened 2019-10-15
Share boards and assign cards

jakobroehrl:
* Target version: vue

### Summary
 - Share boards with users, groups and circles
 - Assign cards to shared users and users of shared groups and shared circles

### TODO
- [x] get users of circles, circles have to be improved
- [x] BoardSidebar, sharing: Only the 1st shared element is filtered out of the list
- [x] Sharing option isn't visible, browser problem?
grafik

### DIFF
--- a/lib/Service/PermissionService.php
+++ b/lib/Service/PermissionService.php
@@ -262,6 +262,26 @@ class PermissionService {
 					$users[$user->getUID()] = new User($user);
 				}
 			}
+
+			if ($this->circlesEnabled && $acl->getType() === Acl::PERMISSION_TYPE_CIRCLE) {
+				try {
+					$circle = \OCA\Circles\Api\v1\Circles::detailsCircle($acl->getParticipant(), true);
+					if ($circle === null) {
+						$this->logger->info('No circle found for acl rule ' . $acl->getId());
+						continue;
+					}
+					
+					/* $b = $circle->getMember();
+
+					foreach ($circle['members'] as $user) {
+						$a = $user;
+						//$users[$user->userId()] = new User($user);
+					}  */ 
+					
+				} catch (\Exception $e) {
+					$this->logger->info('Member not found in circle that was accessed. This should not happen.');
+				}
+			}
 		}
 		$this->users[(string) $boardId] = $users;
 		return $this->users[(string) $boardId];
--- a/lib/Service/PermissionService.php
+++ b/lib/Service/PermissionService.php
@@ -271,9 +271,13 @@ class PermissionService {
 						continue;
 					}
 				
-					foreach ($circle->getMembers() as $user) {
-						$iuser = $this->userManager->get($user->getUserId());
-						$users[$user->getUserId()] = new User($iuser);
+					foreach ($circle->getMembers() as $member) {
+						$user = $this->userManager->get($member->getUserId());
+						if ($user === null) {
+							$this->logger->info('No user found for circle member ' . $member->getUserId());
+						} else {
+							$users[$member->getUserId()] = new User($user);
+						}
 					}   
 				} catch (\Exception $e) {
 					$this->logger->info('Member not found in circle that was accessed. This should not happen.');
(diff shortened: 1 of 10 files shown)


juliushaertl:
>  get users of circles, circles have to be improved

Could you give this patch for the circles app a try ? https://gist.github.com/juliushaertl/66026f56a0438d1e1ad96d961cf57adb

lib/Service/PermissionService.php
```diff
@@ -270,14 +270,11 @@ public function findUsers($boardId) {
...
-					/* $b = $circle->getMember();
-
-					foreach ($circle['members'] as $user) {
-						$a = $user;
-						//$users[$user->userId()] = new User($user);
-					}  */ 
-					
+				
+					foreach ($circle->getMembers() as $user) {
+						$iuser = $this->userManager->get($user->getUserId());
```
juliushaertl: Please check the result of the user manager, so we don't add users that do not exist.

juliushaertl: Also $user would be better fitting the general naming convention

lib/Service/PermissionService.php
```diff
@@ -262,6 +262,23 @@ public function findUsers($boardId) {
...
+
+			if ($this->circlesEnabled && $acl->getType() === Acl::PERMISSION_TYPE_CIRCLE) {
+				try {
+					$circle = \OCA\Circles\Api\v1\Circles::detailsCircle($acl->getParticipant(), true);
+					if ($circle === null) {
+						$this->logger->info('No circle found for acl rule ' . $acl->getId());
+						continue;
+					}
+				
+					foreach ($circle->getMembers() as $user) {
```
juliushaertl: maybe $member is more obvious

lib/Service/PermissionService.php
juliushaertl: getUserId() is only valid if the type of the circle member is a user

jakobroehrl: > getUserId() is only valid if the type of the circle member is a user
OK, are there any other types? Do I need to check before?

juliushaertl: https://github.com/nextcloud/circles/blob/00e251705362f9e20c4d2bea14020421b88c9bc3/lib/Model/BaseMember.php#L48-L51 

but deck only supports sharing with users, so that is the only one we need to handle

juliushaertl:
Works nicely, some more comments:

- [ ] When adding/removing acls the list of users that can be assigned to should be updated
- [ ] Groups/circles should not have an avatar but the related icon, similar to the current master: 
image

jakobroehrl:
>     * [ ]  When adding/removing acls the list of users that can be assigned to should be updated
How can I solve this clever? Adding a user is no problem but getting the users of added group/circle on the UI is maybe not so easy. 

>     * [ ]  Groups/circles should not have an avatar but the related icon, similar to the current master:
Do we have better icons for this?
grafik

>     * [ ]  Please rename the component to remove the typo in its name
Where is this typo?

juliushaertl:
> Do we have better icons for this?

Just needs some CSS to make them pretty :wink: You can have a look at the current stable release, on how it is done there.

> How can I solve this clever? Adding a user is no problem but getting the users of added group/circle on the UI is maybe not so easy.

Hm, tricky. Maybe just refetch the board information which includes the list of members for now?

> Where is this typo?

SharingTabSidebar**d**.vue

jakobroehrl:
> > How can I solve this clever? Adding a user is no problem but getting the users of added group/circle on the UI is maybe not so easy.
> 
> Hm, tricky. Maybe just refetch the board information which includes the list of members for now?

How can I access from here https://github.com/nextcloud/deck/blob/vue/src/store/main.js#L211 to here?
https://github.com/nextcloud/deck/blob/vue/src/components/board/Board.vue#L92

jakobroehrl:
Todo
- [ ] When adding/removing acls the list of users that can be assigned to should be updated
Other topics fixed

jakobroehrl:
@juliushaertl: Fix the last point. Now the assignable users will be updated after changing the share board users/groups/circles. Let's merge it! :)

PR #1294, opened 2019-10-16
Collections

jakobroehrl:
* Target version: vue
Tried to get the colletions working under the vue branch

Uncaught Error: Cannot find module './init-collections.js'    -> bootstrap:83
### TODO
- [x] Get Board collections running
- [x] Add Card collections
- [x] URL adjustment, remove !


js/webpack.config.js
```diff
@@ -9,7 +9,7 @@ module.exports = {
 	},
 	entry: {
 		deck: ['./init.js'],
-		collections: ['./init-collections.js']
```
juliushaertl: No need for this, this file is legacy code and will be dropped.

webpack.common.js
```diff
@@ -2,11 +2,14 @@ const path = require('path');
 const { VueLoaderPlugin } = require('vue-loader');
 
 module.exports = {
-	entry: path.join(__dirname, 'src', 'main.js'),
+	entry: {
+		deck: path.join(__dirname, 'src', 'main.js'),
+		collections: ['./src/init-collections.js']
```
juliushaertl: Please adjust to also use path.join

jakobroehrl:
@juliushaertl Changed all your review topics but still get the error: Uncaught Error: Cannot find module './init-collections.js' -> bootstrap:83

juliushaertl:
Seems my commit got lost, please pull again and run the npm build

jakobroehrl:
> 
> Seems my commit got lost, please pull again and run the npm build

git pull
npm run build
Browser reload

Sorry, I still get the error.

lib/Collaboration/Resources/ResourceProviderCard.php
```diff
@@ -0,0 +1,134 @@
...
+	 * @param IResource $resource
+	 * @param IUser|null $user
+	 * @return bool
+	 * @since 16.0.0
+	 */
+	public function canAccessResource(IResource $resource, ?IUser $user): bool {
+		if ($resource->getType() !== self::RESOURCE_TYPE || !$user instanceof IUser) {
+			return false;
+		}
+		$board = $this->getBoard($resource);
```
juliushaertl: The issue here is that you try to fetch the board but you have the card id. So the board is not found and the canAccessResource returns false, so the resource is never added to the project.

jakobroehrl: Thanks for your review. 
The problem is, I can choose a card, but it would not shown
grafik
show project details:
grafik

And if I change the canAccessResource function to "return true;" there is no change

juliushaertl: So it is not just the canAccessResource method but also getResourceRichObject

You need to adjust the implementation to properly fetch the card details. Right now when linking to a card, you add the card id but try to fetch details for the board with that id. It cannot work that way. Please always checkout the nextcloud log in such cases as the error message gives a clear hint on why the request fails:

image

```
"Exception":"Error","Message":"Call to a member function getTitle() on null","Code":0,"Trace":[{"file":"/var/www/html/lib/private/Collaboration/Resources/Manager.php","line":302,"function":"getResourceRichObject","class":"OCA\\Deck\\Collaboration\\Resources\\ResourceProviderCard"
```

Issue #1253, opened 2019-09-16
Use clip instead of folder to represent attachment


stefan-niedermann:
Please use a clip icon instead of a folder to represent attachments.

The android client will do so. BTW this is a feature request by @jancborchardt who said it would be good to do this in the web interface also :)

![grafik](https://user-images.githubusercontent.com/4741199/64985964-5accb480-d8c6-11e9-9f06-53cef71af248.png)

jancborchardt:
To be specific it’s @jenniferpiperek who brought it up, I simply agree with it. :)

jakobroehrl:
@jancborchardt Where can I get the icon?
Didn't found it here: https://docs.nextcloud.com/server/17/developer_manual/design/icons.html

jancborchardt:
I assume @stefan-niedermann has it from Material icons, and it would be findable in his repo at https://github.com/stefan-niedermann/nextcloud-deck ? :)

juliushaertl:
Are those icons compatible to be used with AGPL?

jancborchardt:
> Are those icons compatible to be used with AGPL?

Yep, it’s Apache 2.0 as per https://material.io/resources/icons/?style=baseline and https://github.com/google/material-design-icons :)

jancborchardt:
> Are those icons compatible to be used with AGPL?

Yep, it’s Apache 2.0 as per https://material.io/resources/icons/?style=baseline and https://github.com/google/material-design-icons :)

See specific compatibility info: https://en.wikipedia.org/wiki/Apache_License#Compatibility (TL;DR Apache is more permissive than the copyleft GPL, so it is ok to include Apache stuff into GPL, just not the other way around.)

juliushaertl:
Closing as this was implemented in the vue branch and is now merged to master

PR #1305, opened 2019-10-24
clip icon instead of folder icon

jakobroehrl:
* Resolves: #1253 
* Target version: vue

### Summary
changed icon

grafik


css/icons.scss
```diff
@@ -13,6 +13,10 @@
   background-image: url('../img/add-white.svg');
 }
 
+.icon-attach {
+  background-image: url('../img/attach.svg');
```
juliushaertl: You also need to add an entry to the list here: https://github.com/nextcloud/deck/pull/1305/files#diff-eccb3df06bdb0d0e156bfd782b26d599R58 

That way we can be sure that when using the dark mode the icon is properly inverted.

PR #1315, opened 2019-10-31
list view

jakobroehrl:
Signed-off-by: Jakob <jakob.roehrl@web.de>
* Target version: vue

### TODO
- [ ] setting the orientation is not working: https://github.com/nextcloud/deck/blob/4d0741e6911f877c9424f4be62c1af29dac51fcf/src/components/board/Board.vue#L27
- [ ] change listItem class, actually it's only a copy of carditem
- [ ] use this icon: "icon-toggle-pictures"

### Picture
grafik

Issue #414, opened 2018-02-01
Filter / sorting options for cards [$167]


juliushaertl:
This issue is to track various filter options that should be implemented in the board view:

- Filter by assigned users (from https://github.com/nextcloud/deck/issues/412)
- Filter by due date (from #222)
- Filter by tags (from https://github.com/nextcloud/deck/issues/289)
- Filter by archived state (from https://github.com/nextcloud/deck/issues/1098)
  - it should be possible to show all/only unarchived/only archived cards

Filtering should include the option to:

- Sort the cards by modification/creation/due date

Some general thoughts:
- Filters should be combinable (from #222)
- Filters should be represented in the URL(from https://github.com/nextcloud/deck/issues/412)

We still need a way to integrate this properly into the UI.

GoetheG:
Very good idea!

juliushaertl:
One possible UI implementation could be to show a filter sidebar, similar to how trello does it:
![image](https://user-images.githubusercontent.com/3404133/42732533-b37dd090-8823-11e8-8ec9-b7ff71e68c3b.png)

RobindaRobot:
Hi, what is the status on filtering? Some items are closed here, but I can't see new functionality in my version of Deck

juliushaertl:
@RobindaRobot There is no filtering implemented right now. The other issues were closed since they were duplicates of this one basically.

hgfhj658d4fg64dfg:
Our nonprofit would love the label filtering so much since we have so many people and areas to plan for, we are losing oversight with either too many cards in one deck or to many decks.
+1

dnik-offcut:
Dear juliushaertl, first of all thank you a lot for working on this issue. it will increase the usability of deck amazingly. Could you allready estimate when the feature will be available?

putt1ck:
Would this functionality not resolve #805 ? In which case the bounty goes up.

BKapelari:
Just as an idea:
Maybe the best option would be to have a search syntax and the option to safe the search as a smart list.
I really like the way how you can do it in Jira:
https://confluence.atlassian.com/jirasoftwarecloud/advanced-searching-764478330.html

mxc:
[...]

Issue #15, opened 2017-02-01
Integrate with Nextcloud CalDAV API [$185]


juliushaertl:
This issue is about integrating deck with Nextclouds CalDAV API.

Current plan from the discussion below:

| Deck | CalDAV | Tasks |
|-|-|-|
| Board | `VCALENDAR` | List |
| Stack | `VTODO` without `RELATED-TO` property | Root-Task |
| Card | `VTODO` with `RELATED-TO` (UUID of the Root-Task / Stack) | Subtask |
| Tags | [`CATEGORIES`](https://tools.ietf.org/html/rfc5545#section-3.8.1.2) or custom implementation (Nextcloud tagging system) | Currently uses `CATEGORIES`, integration with Nextcloud tags would be better, see https://github.com/nextcloud/tasks/issues/89 |
| Assigned users | [`ATTENDEE`](https://tools.ietf.org/html/rfc5545#section-3.8.4.1) or custom implementation | [`ATTENDEE`](https://tools.ietf.org/html/rfc5545#section-3.8.4.1) or custom implementation, see https://github.com/nextcloud/tasks/issues/72 |
| Recurring stacks & cards | [`RRULE`](https://tools.ietf.org/html/rfc5545#section-3.8.5.3) | Recurring tasks & subtasks, see https://github.com/nextcloud/tasks/issues/34 |
| Attachments | [`ATTACH`](https://tools.ietf.org/html/rfc5545#section-3.8.1.1) or custom implementation | [`ATTACH`](https://tools.ietf.org/html/rfc5545#section-3.8.1.1) or custom implementation, see https://github.com/nextcloud/tasks/issues/91 |

> Tags, assigning users, attachments and recurrence is not supported by Tasks yet either, but it would make much sense to implement these features. And if we see that it is not really possible to implement this with pure CalDAV I would be open to implement it without syncing it via CalDAV and have it only available in the webinterface. This would still be better than not having it at all in Tasks and would not limit Deck.

1. Expose deck boards as VCALENDAR, try to map as much data from deck to the VTODO (read only access) #1545 
2. Allow updating VCALENDAR entries, try to map as much as possible from vtodo to the deck database and store the rest as plain vtodo data, so we can still push it out over CalDAV (write access) #2399
3. Deck data could be moved to the DAV backend.

joergmschulz:
as sharing the cards is essential, a log of changes to 
# Activities 
would be nice.

poVoq:
[...]

PR #1319, opened 2019-11-02
card collections/dashboard

jakobroehrl:
* Target version: vue

Added some collections, and it's working so far :)
Could you give me your feedback @juliushaertl ?

- There is now redundant code from the card view
- The Router could be better, I think
- Fetching the data is not really fast, should I try to solve this on the server or client side?
- Any ideas how to change the code, if we have filtering and searching in mind?

Thanks a lot! :)

I think I would use this view in #1315, too. 

grafik

### DIFF
--- a/src/components/CardItem.vue
+++ b/src/components/CardItem.vue
@@ -73,11 +73,9 @@
 </template>
 
 <script>
-import { Avatar } from 'nextcloud-vue'
+import { Avatar, Actions, ActionButton } from '@nextcloud/vue'
 import Color from '../mixins/color'
 import AvatarList from './cards/AvatarList'
-import { Actions } from 'nextcloud-vue/dist/Components/Actions'
-import { ActionButton } from 'nextcloud-vue/dist/Components/ActionButton'
 
 export default {
 	name: 'CardItem',
--- a/src/components/CardItem.vue
+++ b/src/components/CardItem.vue
@@ -1,0 +1,188 @@
+<!--
+* @copyright Copyright (c) 2018 Michael Weimann <mail@michael-weimann.eu>
+*
+* @author Michael Weimann <mail@michael-weimann.eu>
+*
+* @license GNU AGPL version 3 or any later version
+*
+* This program is free software: you can redistribute it and/or modify
+* it under the terms of the GNU Affero General Public License as
+* published by the Free Software Foundation, either version 3 of the
+* License, or (at your option) any later version.
+*
+* This program is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+* GNU Affero General Public License for more details.
+*
+* You should have received a copy of the GNU Affero General Public License
+* along with this program. If not, see <http://www.gnu.org/licenses/>.
+*
+-->
+
+<template>
+	<!-- <router-link :id="`board-${board.id}`"
+		:title="board.title"
+		:to="routeTo" class="board-list-row" tag="div">
+		<div class="board-list-bullet-cell">
+			<div :style="{ 'background-color': `#${board.color}` }" class="board-list-bullet" />
+		</div>
+		<div class="board-list-title-cell">{{ board.title }}</div>
+		<div class="board-list-avatars-cell">
+			<avatar :user="board.owner.uid" class="board-list-avatar" />
+			<avatar v-for="user in limitedAcl" :key="user.id" :user="user.participant.uid"
+				class="board-list-avatar" />
+			<div v-tooltip="otherAcl" v-if="board.acl.length > 5" class="avatardiv popovermenu-wrapper board-list-avatar icon-more" />
+		</div>
+		<div class="board-list-actions-cell" />
+	</router-link> -->
+
+	<div class="card-list-row">
+		<div class="card-list-bullet-cell">
+			<input id="card-status" type="checkbox" class="checkbox">
+			<label for="card-status" />
+		</div>
+		<div class="card-list-bullet-cell">
+			<div :style="{ 'background-color': `#990` }" class="card-list-bullet" />
+		</div>
+		<div class="card-list-title-cell">{{ card.title }}</div>
+
+		<div>
+			<ul class="labels card-list-title-cell">
+				<li v-for="label in card.labels" :key="label.id" :style="labelStyle(label)"><span>{{ label.title }}</span></li>
+			</ul>
+		</div>
+
+		<div>
+			<div v-if="card.attachments" class="card-files icon icon-files-dark" />
+			<div v-if="card.duedate"><span>{{ dueTime(card.duetime) }}</span></div>
+		</div>
+
+		<avatar-list :users="card.assignedUsers" class="card-list-avatars-cell" />
+		<div class="card-list-actions-cell">
+
+			<Actions v-if="!editing" @click.stop.prevent>
+				<ActionButton icon="icon-user">{{ t('deck', 'Assign to me') }}</ActionButton>
+
+				<ActionButton icon="icon-delete">{{ t('deck', 'Delete card') }}</ActionButton>
+				<ActionButton icon="icon-external">{{ t('deck', 'Move card') }}</ActionButton>
+				<ActionButton icon="icon-settings-dark">{{ t('deck', 'Card details') }}</ActionButton>
+			</Actions>
+		</div>
+	</div>
+</template>
+
+<script>
+import { Avatar } from 'nextcloud-vue'
+import Color from '../mixins/color'
+import AvatarList from './cards/AvatarList'
+import { Actions } from 'nextcloud-vue/dist/Components/Actions'
+import { ActionButton } from 'nextcloud-vue/dist/Components/ActionButton'
+
+export default {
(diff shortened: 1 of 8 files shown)


jancborchardt:
Nice! :) Some design feedback:
- [ ] The counters of "Due soon" and "Assigned to you" should use the counter on the right of the row as defined in the component
- [ ] No categorization and indentation is needed. Simply do a flat list with
  > - Due soon
  > - Assigned to you (we always use "you", never "me" in the interface)
  > - All cards
  > `space`
  > - Board 1
  > - Board 2
  > - …

ardarandir:
Hi, we would love this collections/dashboard feature to go live for productive use. I don't quite get when we can expect that to happen. Would you be so kind as to give me any kind of estimation? We are using Deck 0.7.0 on Nextcloud 17 (stable). Thanks in advance!

jakobroehrl:
> 
> Hi, we would love this collections/dashboard feature to go live for productive use. I don't quite get when we can expect that to happen. Would you be so kind as to give me any kind of estimation? We are using Deck 0.7.0 on Nextcloud 17 (stable). Thanks in advance!

@ardarandir 
We have to finish the vue migration first. There is no progress it the last weeks because of other priorities. Sorry, can't give a an estimation. Would love to see this feature, too :)

juliushaertl:
In addition to the comments by @jancborchardt I think we should try to keep the main content view stay in a similar card based layout as the regular board view has. 

I don't see a good use case for an all cards view, as this would basically be just a long endless list without any proper filtering. We should aim to provide lists that give users a good overview or help them to find the relevant cards more easily.

### Due date view

Frame 5

The assigned to me list could also be grouped by duedates, as this is most likely what the priorization is done by. In addition we could allow users to apply custom filters on those lists with https://github.com/nextcloud/deck/issues/414

cc @nextcloud/designers

juliushaertl:
Rebased and changed base branch to master

alexanderdd:
Not sure how you resolve conflicting feedback around here?
Personally, I agree with https://github.com/nextcloud/deck/pull/1319#issuecomment-559928631 and while I think https://github.com/nextcloud/deck/pull/1319#issuecomment-577861069 has a good point, I think this should not stand in the way of merging this pull request.

Or is the only thing missing now the review (one post above)?

Thanks.

juliushaertl:
> Or is the only thing missing now the "Review required"?

No, we definitely should address the mentioned points before merging. :wink:

juliushaertl:
> Not sure how you resolve conflicting feedback around here?

Let's see if @jancborchardt and @nextcloud/designers have any comments on my additions.

jancborchardt:
Agree with everything said by @juliushaertl in https://github.com/nextcloud/deck/pull/1319#issuecomment-577861069 :)

Only thing is that I’m not sure the list layout should use the card style, as that’s too many containers for a single-row layout. I tend towards doing it like @jakobroehrl above or as in the Tasks app.

cc @raimund-schluessler as this now gets very close into the same territory as Tasks and we should coordinate how best to move forward together. :) As @jakobroehrl mentioned, the other tools all go in the direction of just having views for the "to dos", instead of different apps: See Microsoft Tasks in Teams, Asana, Ticktick, Todoist, Ora, Zenkit, …

raimund-schluessler:
> cc @raimund-schluessler as this now gets very close into the same territory as Tasks and we should coordinate how best to move forward together. :) As @jakobroehrl mentioned, the other tools all go in the direction of just having views for the "to dos", instead of different apps: See Microsoft Tasks in Teams, Asana, Ticktick, Todoist, Ora, Zenkit, …

The easiest would be if Deck would just use CalDAV 😉 Then the user could decide which view/app to use and re-implementing the same view for every app would not be necessary. A card view for the Tasks app was also requested once https://github.com/nextcloud/tasks/issues/100, but since we have Deck, that's not really reasonable.

However, exposing the Deck data as a CalDAV endpoint as described in https://github.com/nextcloud/deck/issues/15 by @juliushaertl wouldn't really do the trick. Tasks could then work with the Deck data, but Deck wouldn't be able to work with the Tasks data.

juliushaertl:
> The easiest would be if Deck would just use CalDAV wink Then the user could decide which view/app to use and re-implementing the same view for every app would not be necessary. A card view for the Tasks app was also requested once nextcloud/tasks#100, but since we have Deck, that's not really reasonable.

> However, exposing the Deck data as a CalDAV endpoint as described in #15 by @juliushaertl wouldn't really do the trick. Tasks could then work with the Deck data, but Deck wouldn't be able to work with the Tasks data.

I also considered using CalDAV as a data source in the past, but this limits us somehow to represent the card data in the properties defined in the vevent/vtodo rfc. One big concern when discussing this with @georgehrke were tags and the actual stack a card is in. There is the category field which would work for tags, but then we don't have a way to represent the stack. Also the category is basically a free text field, so when changing a task over CalDAV there is no way in checking if such a stack actually exists on the board and if a card can be moved to that one. Other card data that might be hard to represent would be assigned users and attachments. 

I'm open for ideas on how we could solve this, as obviously having a shared data backend is what would make the most sense here, it's just that I don't see a proper way of achieving it.

georgehrke:
Another big problem that would come up from taking CalDAV and `VTODO`s as a source for Deck would be how to handle recurring `VTODO`s. There is no concept of recurring cards in Deck.

I totally agree that from a Design POV, Deck should work on the same data as Tasks and just be an alternative view for it, but from a technical POV, this is introducing to many problems, which we have still no answers to.

juliushaertl:
Back to the original topic of the pull request, I would suggest we go for a card based approach as well for now, as once boards are exposed as calendars with https://github.com/nextcloud/deck/pull/1545 we could redirect the user to the tasks app showing the cards in a list as a first starting point.

BKapelari:
INFO: We did a quick transitional solution for our company by adding 3 widgets to the dashboard.
is not a really useful solution but it does the job. if you click on the link the card shows in detail view in deck.

Bildschirmfoto 2020-02-26 um 18 02 42

jakobroehrl:
There are a lot of the same functionality in Deck and Tasks eg:
- assign to user
- share with user, group, circle
- comments
- attachments
- tags
- activity
- due dates
- filter, sort
- collections (dueToday, assignToMe, ...)

I think it's easier to interate a list/table view in Deck than duplicate and maintain the hole functionality in two apps.

BKapelari:
> There are a lot of the same functionality in Deck and Tasks eg:
> 
>     * assign to user
> 
>     * share with user, group, circle
> 
>     * comments
> 
>     * attachments
> 
>     * tags
> 
>     * activity
> 
>     * due dates
> 
>     * filter, sort
> 
>     * collections (dueToday, assignToMe, ...)
> 
> 
> I think it's easier to interate a list/table view in Deck than duplicate and maintain the hole functionality in two apps.

how can you assing tasks to users in the Tasks App?

jakobroehrl:
> 
> how can you assing tasks to users in the Tasks App?

You can't. But it's an function both tools should have.

raimund-schluessler:
I continued our off-topic discussion in the correct issue: https://github.com/nextcloud/deck/issues/15#issuecomment-592582216

alexanderdd:
So what will happen to this pull request? Merge or decline (if thats the right word)?

Path1
@juliushaertl proposed to use a card layout instead  https://github.com/nextcloud/deck/pull/1319#issuecomment-577861069

Path2
@jancborchardt said he likes the row style as proposed in this pull request, but some minor improvements are needed (these https://github.com/nextcloud/deck/pull/1319#issuecomment-577861069)

Which path will we continue? (I am never sure how decisions are made here...)

juliushaertl:
> So what will happen to this pull request? Merge or decline (if thats the right word)?

Well, the discussion drove a bit into details that are actually part of #15 but are of course related to the lists. My comment from https://github.com/nextcloud/deck/pull/1319#issuecomment-591394936 still stands. So with that and the plan we came up with in https://github.com/nextcloud/deck/issues/15 I think reusing the tasks app for any list view would make the most sense here.

> Which path will we continue? (I am never sure how decisions are made here...)

We usually try to find a solution that is somehow common sense in the end. Let me quote our code of conduct here:

> Be pragmatic
> Nextcloud is a pragmatic community. We value tangible results over having the last word in a discussion. We defend our core values like freedom and respectful collaboration, but we don't let arguments about minor issues get in the way of achieving more important results. We are open to suggestions and welcome solutions regardless of their origin. When in doubt support a solution which helps getting things done over one which has theoretical merits, but isn't being worked on. Use the tools and methods which help getting the job done. Let decisions be taken by those who do the work.
https://nextcloud.com/contribute/code-of-conduct/

juliushaertl:
I think we drifted away from what this pull request was actually about - the dashboard to give quick access to important cards.

Let me summarize what I think the consent was:
- Due dates - A column view makes most sense here as we can have the due date categorization as different columns here. 

- All cards - does not have a real use case -> Let's remove that one

## Open questions: 

### Assigned to me

>> Only thing is that I’m not sure the list layout should use the card style, as that’s too many containers for a single-row layout. I tend towards doing it like @jakobroehrl above or as in the Tasks app.

I was not arguing for a single row layout but to still provide some sort of grouping for the dashboard overview, like this:

> The assigned to me list could also be grouped by duedates, as this is most likely what the priorization is done by.

juliushaertl:
Continued in https://github.com/nextcloud/deck/pull/1934

Issue #1275, opened 2019-10-02
Rest API: assigned users without read permissions


desperateCoder:
**Describe the bug**
A friend of mine has an account on my nextcloud instance. I made a board for him and shared it with him. Now my friend creates a board (using Deck Android app, therefore the Rest-API) and assigns a card to me.
-> I get a notification via NC Files app
-> I still can't see the board, neither the card.

Is this working as designed?

juliushaertl:
As discussed you should not be able to assign users to a card which are not part of the board. How do you get the list of users in the android app for assigning to the card?

desperateCoder:
Select * from users where accountId = ? -> no acl check, not depending on boards. We plan to include ACL as well, but shouldn't the server validate the rest call? I mean... You know...

desperateCoder:
BTW: The users in the DB are all users I know about for a specific account.

desperateCoder:
Just looked it up. The users are either owners of boards, or the user behind the account, or already assigned to at least one card on any board of the account

juliushaertl:
The `GET /boards - Get a list of boards` endpoint should return a list of users per board that can be assigned to the card. However I'll look into limiting the assignment on the server side as well.

juliushaertl:
![image](https://user-images.githubusercontent.com/3404133/66220254-fd5da380-e6cc-11e9-9557-ef5b0b26a1e2.png)

juliushaertl:
Fixed with https://github.com/nextcloud/deck/commit/6fa7295b420dc7fececcbc0352ba07e685982024

PR #1395, opened 2019-12-19
Limit card assignment to users who are participants of the board

juliushaertl:
@desperateCoder as discussed in https://github.com/nextcloud/deck/issues/1275 the server will now report an error if you try to assing a user that is not part of the board.


desperateCoder:
Nice! We already check this in the ui now and hide the buttons for triggering unpermitted operations. And thanks to this PR a mean attacker won't be able to make this call, too!

Issue #1233, opened 2019-09-09
Case insensitive searching


rprpx:
Could we make the SEARCH field in the upper right case-INsensitive? 

Having to be sure I search for the same CASE may lead me to miss a card that's actually present.

Additionally, case-INsensitive searching in this field would then be consistent with typing/adding a TAG to an item/card, which narrows your tag selections as you type regardless of case used in the actual TAG. 

Thank you!

matchish:
A Search engine for the query can be different for different applications. For example, if an application uses Elasticsearch then results can be sorted by a score(relevancy) so then you will get all cards but cards with the wrong case will be at the end of the list.

So I think it not the best idea to lowercase query in the field, but I agree that usually, it's more useful to have a case-insensitive search. In what application you don't have case-insensitive search. I mean in `files` and `contacts` for example the search is case-insensitive.

PR #1434, opened 2020-01-10
Merge pull request #1434 from 4workers/insensitive-search

matchish:
* Resolves: https://github.com/nextcloud/deck/issues/1233
* Target version: master

### DIFF
--- a/js/filters/cardSearchFilter.js
+++ b/js/filters/cardSearchFilter.js
@@ -32,6 +32,7 @@ app.filter('cardSearchFilter', function() {
 			var _card = card;
 			Object.keys(rules).some(function(rule) {
 				var pattern = rules[rule];
+				// eslint-disable-next-line detect-non-literal-regexp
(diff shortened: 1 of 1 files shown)


juliushaertl:
> Found non-literal argument to RegExp Constructor

Can be ignored since we are running this in the frontend so the user would just be able to trigger a DoS on their own browser tab.

matchish:
> > Found non-literal argument to RegExp Constructor
> 
> Can be ignored since we are running this in the frontend so the user would just be able to trigger a DoS on their own browser tab.

Could you tell me the rule name? I can't find the rule name to ignore the error.

matchish:
cc @juliushaertl

juliushaertl:
That should be the correct rule, but I'm not sure how codacy is checking that. Fine to merge from my side and dismis the warning in codacy. Also our own eslint run is fine: https://drone.nextcloud.com/nextcloud/deck/2714/5/2

juliushaertl:
Thanks a lot for the contribution :rocket:

PR #1482, opened 2020-01-25
Bump version and fix wording (fixes #1088)

### LINKED ISSUE #1088: Possible name and terminology change
We discussed in the past that our current naming might need revision.
It’s currently **Deck**, Board, Stack, Card (app name in bold).

Alternative ideas @juliushaertl and I had:
- **Boards**, Board, Stack, Card
- **Plans**, Plan, Stack, Card
- "Projects" is already taken by the Projects feature in core, so we can not use that.

Just researched how others call it:
- **Trello**, Board, List, Card
- **GitHub Projects**, Project, Column, Card
- **Wekan**, Board, List, Card
- **Asana**, Board, Column, Task
- **Monday.com Kanban**, Board, Column?, Pulse?
- This [blog post](https://zapier.com/blog/best-kanban-apps/) also says: Board, List, Card

So after that, it’s pretty clear that "Card" is the one thing almost everyone uses. Then "List" instead of "Stack/Column" seems to be good and more understandable. And "Board" is also basically standard.

That leaves the app name. So these are our options:
- **Nextcloud Deck**, Board, List, Card
- **Nextcloud Boards**, Board, List, Card
- **Nextcloud Plan/Plans**, Board, List, Card
- Or another app name idea?

What do you think @nextcloud/designers @nextcloud/deck?

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/75358875-possible-name-and-terminology-change?utm_campaign=plugin&utm_content=tracker%2F52758431&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F52758431&utm_medium=issues&utm_source=github).
</bountysource-plugin>

juliushaertl:
Master is now heading to 1.0.0 and we'll rename stacks to lists.

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -3,7 +3,7 @@
       xsi:noNamespaceSchemaLocation="https://apps.nextcloud.com/schema/apps/info.xsd">
     <id>deck</id>
     <name>Deck</name>
-    <summary>A kanban style project and personal management tool for Nextcloud</summary>
+    <summary>Personal planning and team project organization</summary>
     <description>Deck is a kanban style organization tool aimed at personal planning and project organization for teams integrated with Nextcloud.
 
 
@@ -17,7 +17,7 @@
 - 🚀 Get your project organized
 
 </description>
-    <version>0.8.0</version>
+    <version>1.0.0-alpha1</version>
     <licence>agpl</licence>
     <author>Julius Härtl</author>
     <namespace>Deck</namespace>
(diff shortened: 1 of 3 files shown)


juliushaertl:
>  We should change here stack to list: /src/components/board/Stack.vue@master#L44

Ah, thanks. Rebased and fixed.

PR #1492, opened 2020-01-26
Merge pull request #1492 from nextcloud/enh/application-cleanup

juliushaertl:
- Make sure register methods are only called once
- Create default board also if first request is though the API


jakobroehrl:
@juliushaertl a check isn't passed, cloud you have a look please?

juliushaertl:
Let's ignore the warnings. There is no non-static function for the addScript method and the parameter is inherited from the parent class, we just don't need it :wink:

PR #1505, opened 2020-01-28
use color picker from nextcloud vue

jakobroehrl:
* Target version: master 

It's working with the ColorPicker from nextcloud/vue but it don't look nice:
grafik
grafik
grafik

@jancborchardt Any ideas to make this look nice?


juliushaertl:
I'd say we move the indicator for the color picker to the same line as the text field, basically something like this:
image

jakobroehrl:
> 
> 
> I'd say we move the indicator for the color picker to the same line as the text field, basically something like this:
> image

Great idea, let's give it a try

jakobroehrl:
@juliushaertl It's implemented on the 3 relevant positions. 
Only some CSS magic is missing, could you do that?

PR #1507, opened 2020-01-29
Board filter

jakobroehrl:
* Target version: master 

Todo
- [x] due filter reset on click again (@change is not reacting)

grafik

### DIFF
--- a/img/filter.svg
+++ b/img/filter.svg
@@ -1,0 +1,78 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   width="16"
+   height="16"
+   viewBox="0 0 4.2333332 4.2333335"
+   version="1.1"
+   id="svg4524"
+   inkscape:version="0.92.4 5da689c313, 2019-01-14"
+   sodipodi:docname="filter.svg">
+  <defs
+     id="defs4518" />
+  <sodipodi:namedview
+     id="base"
+     pagecolor="#ffffff"
+     bordercolor="#666666"
+     borderopacity="1.0"
+     inkscape:pageopacity="0.0"
+     inkscape:pageshadow="2"
+     inkscape:zoom="11.2"
+     inkscape:cx="-13.015771"
+     inkscape:cy="15.433087"
+     inkscape:document-units="mm"
+     inkscape:current-layer="layer1"
+     showgrid="false"
+     units="px"
+     showguides="true"
+     inkscape:guide-bbox="true"
+     inkscape:window-width="1920"
+     inkscape:window-height="1017"
+     inkscape:window-x="0"
+     inkscape:window-y="26"
+     inkscape:window-maximized="1">
+    <sodipodi:guide
+       position="3.1773623,1.9016928"
+       orientation="0,1"
+       id="guide5088"
+       inkscape:locked="false" />
+  </sodipodi:namedview>
+  <metadata
+     id="metadata4521">
+    <rdf:RDF>
+      <cc:Work
+         rdf:about="">
+        <dc:format>image/svg+xml</dc:format>
+        <dc:type
+           rdf:resource="http://purl.org/dc/dcmitype/StillImage" />
+        <dc:title></dc:title>
+      </cc:Work>
+    </rdf:RDF>
+  </metadata>
+  <g
+     inkscape:label="Ebene 1"
+     inkscape:groupmode="layer"
+     id="layer1"
+     transform="translate(0,-292.76665)">
+    <path
+       style="opacity:1;fill:#000000;fill-opacity:1;stroke:none;stroke-width:0.09337848;stroke-linecap:round;stroke-linejoin:miter;stroke-miterlimit:4;stroke-dasharray:none;stroke-opacity:1;paint-order:stroke fill markers"
+       d="M 0.51971728,293.23203 H 3.8033853 l -1.1728849,1.45285 H 1.6418341 Z"
+       id="rect5069"
+       inkscape:connector-curvature="0"
+       sodipodi:nodetypes="ccccc" />
+    <path
+       style="opacity:1;fill:#000000;fill-opacity:1;stroke:none;stroke-width:0.05817544;stroke-linecap:round;stroke-linejoin:miter;stroke-miterlimit:4;stroke-dasharray:none;stroke-opacity:1;paint-order:stroke fill markers"
+       d="m 1.6418341,294.68488 h 0.9921874 v 1.86627 L 1.637658,296.09596 Z"
+       id="rect5069-7"
+       inkscape:connector-curvature="0"
+       sodipodi:nodetypes="ccccc" />
+  </g>
+</svg>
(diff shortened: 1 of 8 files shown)


src/components/board/TagsTabSidebar.vue
```diff
@@ -60,7 +66,7 @@
 
 import { mapGetters } from 'vuex'
 import Color from '../../mixins/color'
-import ColorPicker from '../ColorPicker'
+import { ColorPicker } from '@nextcloud/vue/dist/Components/ColorPicker'
```
juliushaertl:
```suggestion
import { ColorPicker } from '@nextcloud/vue'
```

Since we use a lot of the components, please use the global import. That way we save quite some bytes in the bundle size.

juliushaertl: Ah nevermind this was from the old color picker branch ;)

juliushaertl:
@jakobroehrl I've rebased the branch to current master, please make sure to reset your local branch to the origin one before continuing.

src/components/Controls.vue
```diff
@@ -69,7 +147,7 @@
 				</Actions>
 				<!-- FIXME: ActionRouter currently doesn't work as an inline action -->
 				<Actions>
-					<ActionButton icon="icon-share" @click="toggleDetailsView" />
+					<ActionButton icon="icon-settings" @click="toggleDetailsView" />
```
juliushaertl: Please keep such changes separate. Ideally we should use the new sidebar icon here:
image

src/components/Controls.vue
```diff
@@ -79,11 +157,13 @@
 <script>
 import { mapState, mapGetters } from 'vuex'
 import { Actions, ActionButton } from '@nextcloud/vue'
+import { Popover } from '@nextcloud/vue/dist/Components/Popover'
```
juliushaertl: Since we use a lot of the components, please use the global import. That way we save quite some bytes in the bundle size.
```suggestion
import { Popover } from '@nextcloud/vue'
```

src/store/card.js
```diff
@@ -30,7 +30,8 @@ export default {
 		cards: [],
 	},
 	getters: {
-		cardsByStack: state => (id) => {
+		cardsByStack: (state, getters) => (id) => {
+			console.log(getters.getFilter)
```
juliushaertl: :fire:

src/store/main.js
```diff
@@ -61,8 +61,12 @@ export default new Vuex.Store({
 		boardFilter: BOARD_FILTERS.ALL,
 		activity: [],
 		activityLoadMore: true,
+		filter: { tags: [], user: [], due: [] },
 	},
 	getters: {
+		getFilter: state => {
```
juliushaertl: We could just use mapState in the component that uses it so there is less code in the store.

juliushaertl:
@jakobroehrl Pushed some fixes, feel free to take over again now :wink:

src/store/main.js
jakobroehrl: removed getFilter

src/store/card.js
```diff
@@ -30,8 +30,87 @@ export default {
...
+					if (due !== '') {
+						const datediffHour = ((new Date() - new Date(card.duedate)) / 3600000)
+						switch (due) {
+						case 'overdue':
+							if (card.overdue === 0) {
+								return false
+							} else {
+								return true
+							}
+							break
```
juliushaertl: As eslint states, we don't need to break here if we return earlier

src/store/card.js
```diff
@@ -30,8 +30,87 @@ export default {
...
+						if (!allUsersMatch) {
+							return false
+						}
+					}
+
+					if (due !== '') {
+						const datediffHour = ((new Date() - new Date(card.duedate)) / 3600000)
+						switch (due) {
+						case 'overdue':
+							if (card.overdue === 0) {
```
juliushaertl: Can be simplified as `return card.overdue > 0`

jakobroehrl: no, that's not working

juliushaertl:
Lint is not happy, also as mentioned please rebase instead of creating merge commits, that allows us to maintain clean git log in the master branch.

img/filter.svg
```diff
@@ -0,0 +1,78 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
juliushaertl: We should minify this one with https://github.com/nextcloud/server/blob/master/build/image-optimization.sh#L10

src/components/Controls.vue
```diff
@@ -47,6 +47,102 @@
...
+							<div class="filter--item">
+								<input
+									id="dueMonth"
+									v-model="filter.due"
+									type="radio"
+									class="radio"
+									value="dueMonth"
+									@change="setFilter"
+									@click="beforeSetFilter">
+								<label for="dueMonth">{{ t('deck', 'Next 30 Days') }}</label>
```
juliushaertl:
```suggestion
								<label for="dueMonth">{{ t('deck', 'Next 30 days') }}</label>
```

src/components/Controls.vue
```diff
@@ -47,6 +47,102 @@
...
+							<div class="filter--item">
+								<input
+									id="dueWeek"
+									v-model="filter.due"
+									type="radio"
+									class="radio"
+									value="dueWeek"
+									@change="setFilter"
+									@click="beforeSetFilter">
+								<label for="dueWeek">{{ t('deck', 'Next 7 Days') }}</label>
```
juliushaertl:
```suggestion
								<label for="dueWeek">{{ t('deck', 'Next 7 days') }}</label>
```

src/components/Controls.vue
```diff
@@ -47,6 +47,102 @@
...
+							<div class="filter--item">
+								<input
+									id="noDue"
+									v-model="filter.due"
+									type="radio"
+									class="radio"
+									value="noDue"
+									@change="setFilter"
+									@click="beforeSetFilter">
+								<label for="noDue">{{ t('deck', 'No due') }}</label>
```
juliushaertl:
```suggestion
								<label for="noDue">{{ t('deck', 'No due date') }}</label>
```

src/store/card.js
```diff
@@ -30,13 +30,75 @@ export default {
...
-		cardsByStack: (state, getters) => (id) => {
-			return state.cards.filter((card) => card.stackId === id && (getters.getSearchQuery === '' || (card.title.toLowerCase().includes(getters.getSearchQuery.toLowerCase()) || card.description.toLowerCase().includes(getters.getSearchQuery.toLowerCase())))
-			).sort((a, b) => a.order - b.order)
-		},
-		cardById: state => (id) => {
-			return state.cards.find((card) => card.id === id)
-		},
+		cardsByStack: (state, getters, rootState) => (id) => {
+			return state.cards
+				.filter((card) => card.stackId === id && (getters.getSearchQuery === '' || (card.title.toLowerCase().includes(getters.getSearchQuery.toLowerCase()) || card.description.toLowerCase().includes(getters.getSearchQuery.toLowerCase())))
```
juliushaertl: We should move this filter rule below the other one, so that the string filter operations are performed on the filtered subset already.

src/store/card.js
```diff
@@ -30,13 +30,75 @@ export default {
...
+						if (!allUsersMatch) {
+							return false
+						}
+					}
+
+					if (due !== '') {
+						const datediffHour = ((new Date(card.duedate) - new Date()) / 3600000)
+						switch (due) {
+						case 'noDue':
+							if (card.duedate === null) {
```
juliushaertl:
```suggestion
							return (card.duedate === null)
```

src/store/card.js
```diff
@@ -30,13 +30,75 @@ export default {
...
+								return true
+							} else {
+								return false
+							}
+						case 'overdue':
+							if (card.overdue === 3) {
```
juliushaertl:
```suggestion
							return (card.overdue === 3)
```

juliushaertl: Same for the others :wink:

img/filter.svg
jakobroehrl: ok, what is exactly to do?

juliushaertl:
:rocket:

ilippert:
Would it be possible to apply this filter across boards? I.e. at apps/deck/#!/ apply the filter and get the results? That would be very helpful.

Issue #1059, opened 2019-05-21
Comments: Need Error Message on Length Exceeding a 1000 Characters


blu3knight:
**Describe the bug**
No Error Message is displayed in the comments it just keeps spinning, had to go in to logs to figure out that this was because Length is > 1000 Characters

**To Reproduce**
Have more then a 1000 Characters message

**Expected behavior**
Error message of some kind to appear

**Screenshots**

**Client details:**
 - OS: MAC OSX
 - Browser Safari
 - Version Version 12.1
 - Device: Laptop

<details>
<summary>Server details</summary>

**Operating system**:
Linux / Gentoo

**Web server:**
Apache

**Database:**
MySQL

**PHP version:**
7.1

**Nextcloud version:** (see Nextcloud admin page)

**Where did you install Nextcloud from:**

**Signing status:**

```
Login as admin user into your Nextcloud and access
http://example.com/index.php/settings/integrity/failed
paste the results here.
```

**List of activated apps:**

```
If you have access to your command line run e.g.:
sudo -u www-data php occ app:list
from within your Nextcloud installation folder
```

**Nextcloud configuration:**

```
If you have access to your command line run e.g.:
sudo -u www-data php occ config:list system
from within your Nextcloud installation folder

or

Insert your config.php content here
Make sure to remove all sensitive content such as passwords. (e.g. database password, passwordsalt, secret, smtp password, …)
```

**Are you using an external user-backend, if yes which one:** LDAP/ActiveDirectory/Webdav/...

</details>

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
```
Insert your Nextcloud log here
```

#### Browser log
```
Insert your browser log here, this could for example include:

a) The javascript console log
b) The network log
c) ...
```

</details>

juliushaertl:
Fixed in #1508

smehrbrodt:
I just tried to add a comment which has about 480 chars, but contains a number of links.
I still got the error message, maybe because the links were expanded to html?

Is there a technical reason for this limitation?

PR #1508, opened 2020-01-29
Comments

juliushaertl:
## Feature parity

- [x] Move to vue-at since it seems a lot easier than prosemirror if we just use if for mentions
- [x] Fetch comments in steps
- [x] Infinite scrolling
- [x] Load new comments once posting
- [x] Display UserBubble in the comment list
- [x] Use vue-at when editing comments
- [x] Change fetch limit to something useful
- [x] Show indicator for unread comments
- [x] Mark comments as read
- [x] Check if comment is empty or exceeds 1000 chars https://github.com/nextcloud/deck/issues/1059
- [x] comments only shown if there are more the one posted already
- [x] enter should save comments


jakobroehrl:
test results
- possible to save empty comments
- enter should save comments
- comments only shown if there are more the one  posted already
- mention is working

juliushaertl:
Still in development, I've added your points to the todolist ;)

PR #1511, opened 2020-01-30
search cards (title and description)

jakobroehrl:
* Target version: master 

- [x] case insensitive?
- [ ] search over all boards?


src/main.js
```diff
@@ -44,6 +44,8 @@ Vue.mixin({
 	},
 })
 
+Vue.prototype.$OC = OC
```
juliushaertl: We don't need that one.

src/main.js
```diff
@@ -58,5 +60,16 @@ new Vue({
 	el: '#content',
 	router,
 	store,
+	mounted: function() {
+		this.$OC.Search = new OCA.Search(this.filter, this.cleanSearch)
```
juliushaertl: no need to assing it anywhere for now, just call `new OCA.Search()`

src/store/card.js
```diff
@@ -30,8 +30,12 @@ export default {
 		cards: [],
 	},
 	getters: {
-		cardsByStack: state => (id) => {
-			return state.cards.filter((card) => card.stackId === id).sort((a, b) => a.order - b.order)
+		cardsByStack: (state, getters) => (id) => {
+			return state.cards.filter(
+				(card) => card.stackId === id
+				&& (card.title.includes(getters.getSearchQuery) || card.description.includes(getters.getSearchQuery))
```
juliushaertl: Indentation looks a bit odd here, also it would be nicer if we don't do the include calls at all if no search query is set.

juliushaertl:
> case insensitive?

yes, please :wink: 

> search over all boards?

Needs backend, so that is something for a different pr.

src/main.js
jakobroehrl: If I change 
this.$OC.Search = new OCA.Search(this.filter, this.cleanSearch)
to
this.$OC.Search = new OCA.Search()
the search field disappears.

src/main.js
jakobroehrl: If I remove it, I get: Error in mounted hook: "TypeError: this.$OC is undefined"

src/main.js
juliushaertl: Sorry, you got me wrong there. You can just remove the assignment to `this.$OC.Search` and call `new OCA.Search(this.filter, this.cleanSearch)` without assigning the result.

src/main.js
juliushaertl: Should be fine once the assignment is gone :wink:

juliushaertl:
eslint is still not happy :wink:

jakobroehrl:
>
> eslint is still not happy 😉

/var/www/html/server/apps/deck/src/main.js
  64:3  error  Do not use 'new' for side effects  no-new

/var/www/html/server/apps/deck/src/store/card.js
  35:5  error  Unexpected mix of '&&' and '||'  no-mixed-operators
  36:5  error  Unexpected mix of '&&' and '||'  no-mixed-operators

Mhh, could we merge nevertheless?

juliushaertl:
> Mhh, could we merge nevertheless?

No, if there are false positive errors then we can add an ignore statement, but those two can be handled:

https://eslint.org/docs/rules/no-mixed-operators
https://eslint.org/docs/rules/no-new

PR #1513, opened 2020-01-30
create several cards if the new title contains a linebreak

jakobroehrl:

Todo
 - [ ] Let Textarea look nice 

Function is implemented, tested copy from textfile and excel. It's working. 😀

grafik

### DIFF
--- a/src/components/board/Stack.vue
+++ b/src/components/board/Stack.vue
@@ -59,8 +59,7 @@
 				type="text"
 				class="no-close"
 				placeholder="Add a new card"
-				required>
-			</textarea>
+				required />
 
 			<input class="icon-confirm"
 				type="submit"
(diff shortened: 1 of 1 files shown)


src/components/board/Stack.vue
```diff
@@ -161,14 +162,36 @@ export default {
...
+				match.forEach(() => (line) => {
+				
+					let newCard = {
+					title: line,
+					stackId: this.stack.id,
+					boardId: this.stack.boardId,
+					}
+				
+					this.$store.dispatch('addCard', newCard)
+					this.newCardTitle = ''
```
juliushaertl: this should be moved out of the loop

src/components/board/Stack.vue
```diff
@@ -161,14 +162,36 @@ export default {
...
 		clickAddCard() {
-			const newCard = {
-				title: this.newCardTitle,
-				stackId: this.stack.id,
-				boardId: this.stack.boardId,
+
+			let match = /\r|\n/.exec(this.newCardTitle);
+			if (match) {
+				
+				match.forEach(() => (line) => {
```
juliushaertl: This is why it is not working, there is a a callback returned by your forEach callback which means that it will never be executed.

```suggestion
				match.forEach((line) => {
```

juliushaertl:
> Changed the input field with a textfield. Could we make this look nice, too?

Sure, that should be possible, similar to the styles in the comments tab.

> Do you see an error here? It's not creating the cards if an linebreak was found

See inline comment.

If this is added, pressing enter should still send the comment. We can allow newlines when Shift+Enter is pressed. 

Please run `npm run lint:fix` before committing, since otherwise the lint annotations in the diff make it really hard to review.

juliushaertl:
Could you also set the labels and milestones for your new pull requests?

jakobroehrl:
> 
> Could you also set the labels and milestones for your new pull requests?

Sure, sorry about that.

juliushaertl:
Let's close this as discussed. The behavior is not self-explaining and users might be confused why a newline is not part of the card title and in contrast creating multiple cards. 

cc @jancborchardt

PR #1514, opened 2020-01-30
add calendar link

jakobroehrl:
* Target version: master 

grafik

1st try, doesn't work


jakobroehrl:
@juliushaertl 
Do we need this link or do you have another idea to link to the calendar?
If we need it, is this the right approach?

PR #1524, opened 2020-02-04
Attachments

jakobroehrl:
* Target version: master 

### Todo
- [x] if attachment is already there, ask if update is required
- [x] restore removed attachment 
- [x] Clicking the attachment should download it
- [x] Styles (https://github.com/nextcloud/deck/blob/stable0.8/css/style.scss#L867-L957)
- [ ] drag and drop

### DIFF
--- a/lib/Exceptions/ConflictException.php
+++ b/lib/Exceptions/ConflictException.php
@@ -1,0 +1,36 @@
+<?php
+/**
+ * @copyright Copyright (c) 2020 Jakob Röhrl <jakob.roehrl@web.de>
+ *
+ * @author Jakob Röhrl <jakob.roehrl@web.de>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ *  This program is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU Affero General Public License as
+ *  published by the Free Software Foundation, either version 3 of the
+ *  License, or (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU Affero General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Affero General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\Deck;
+
+
+class ConflictException extends \Exception {
+
+	public function __construct($message) {
+		parent::__construct($message);
+	}
+
+	public function getStatus() {
+		return 409;
+	}
+}
--- a/lib/Exceptions/ConflictException.php
+++ b/lib/Exceptions/ConflictException.php
@@ -21,16 +21,24 @@
  *
  */
 
-namespace OCA\Deck;
+namespace OCA\Deck\Exceptions;
 
+use OCA\Deck\StatusException;
 
-class ConflictException extends \Exception {
+class ConflictException extends StatusException {
 
-	public function __construct($message) {
+	private $data;
+
+	public function __construct($message, $data = null) {
 		parent::__construct($message);
+		$this->data = $data;
 	}
 
 	public function getStatus() {
 		return 409;
+	}
+
+	public function getData() {
+		return $this->data;
 	}
 }
--- a/lib/Service/FileService.php
+++ b/lib/Service/FileService.php
@@ -26,6 +26,7 @@ namespace OCA\Deck\Service;
 use OC\Security\CSP\ContentSecurityPolicyManager;
 use OCA\Deck\Db\Attachment;
 use OCA\Deck\StatusException;
+use OCA\Deck\Exceptions\ConflictException;
 use OCP\AppFramework\Http\ContentSecurityPolicy;
 use OCP\AppFramework\Http\EmptyContentSecurityPolicy;
 use OCP\AppFramework\Http\FileDisplayResponse;
@@ -42,7 +43,7 @@ use OCP\IConfig;
 use OCP\IL10N;
 use OCP\ILogger;
 use OCP\IRequest;
-use OCP\Deck\ConflictException;
+
 
 class FileService implements IAttachmentService {
 
@@ -154,7 +155,7 @@ class FileService implements IAttachmentService {
 		$folder = $this->getFolder($attachment);
 		$fileName = $file['name'];
 		if ($folder->fileExists($fileName)) {
-			throw new ConflictException('File already exists.');
+			throw new ConflictException('File already exists.', $attachment);
 		}
 
 		$target = $folder->newFile($fileName);
--- a/lib/Service/FileService.php
+++ b/lib/Service/FileService.php
@@ -42,6 +42,7 @@ use OCP\IConfig;
 use OCP\IL10N;
 use OCP\ILogger;
 use OCP\IRequest;
+use OCP\Deck\ConflictException;
 
 class FileService implements IAttachmentService {
 
@@ -146,13 +147,14 @@ class FileService implements IAttachmentService {
 	 * @param Attachment $attachment
 	 * @throws NotPermittedException
 	 * @throws StatusException
+	 * @throws ConflictException
 	 */
 	public function create(Attachment $attachment) {
 		$file = $this->getUploadedFile();
 		$folder = $this->getFolder($attachment);
 		$fileName = $file['name'];
 		if ($folder->fileExists($fileName)) {
-			throw new StatusException('File already exists.');
+			throw new ConflictException('File already exists.');
 		}
 
 		$target = $folder->newFile($fileName);
--- a/lib/Service/FileService.php
+++ b/lib/Service/FileService.php
@@ -25,6 +25,7 @@ namespace OCA\Deck\Service;
 
 use OC\Security\CSP\ContentSecurityPolicyManager;
 use OCA\Deck\Db\Attachment;
+use OCA\Deck\Db\AttachmentMapper;
 use OCA\Deck\StatusException;
 use OCA\Deck\Exceptions\ConflictException;
 use OCP\AppFramework\Http\ContentSecurityPolicy;
@@ -53,6 +54,7 @@ class FileService implements IAttachmentService {
 	private $logger;
 	private $rootFolder;
 	private $config;
+	private $attachmentMapper;
 
 	public function __construct(
 		IL10N $l10n,
@@ -60,7 +62,8 @@ class FileService implements IAttachmentService {
 		IRequest $request,
 		ILogger $logger,
 		IRootFolder $rootFolder,
-		IConfig $config
+		IConfig $config,
+		AttachmentMapper $attachmentMapper
 	) {
 		$this->l10n = $l10n;
 		$this->appData = $appData;
@@ -68,6 +71,7 @@ class FileService implements IAttachmentService {
 		$this->logger = $logger;
 		$this->rootFolder = $rootFolder;
 		$this->config = $config;
+		$this->attachmentMapper = $attachmentMapper;
 	}
 
 	/**
@@ -155,6 +159,7 @@ class FileService implements IAttachmentService {
 		$folder = $this->getFolder($attachment);
 		$fileName = $file['name'];
 		if ($folder->fileExists($fileName)) {
+			$attachment = $this->attachmentMapper->findByData($attachment->getCardId(), $fileName);
 			throw new ConflictException('File already exists.', $attachment);
 		}
 
--- a/src/components/card/CardSidebarTabAttachments.vue
+++ b/src/components/card/CardSidebarTabAttachments.vue
@@ -23,7 +23,7 @@
 <template>
 	<div>
 		<button class="icon-upload" @click="clickAddNewAttachmment()">
-			{{ t('deck', 'Upload attachment') }}
+			{{ t('settings', 'Upload attachment') }}
 		</button>
 		<input ref="localAttachments"
 			type="file"
@@ -59,22 +59,20 @@
 			</ul>
 		</div>
 
-		<Modal v-if="modalShow" title="File already exists" @close="modalShow=false">
+		<Modal v-if="modalShow" :title="t('deck', 'File already exists')" @close="modalShow=false">
 			<div class="modal__content">
 				<h2>{{ t('deck', 'File already exists') }}</h2>
 				<p>
-					{{ t('deck', 'A file with the name') }}
-					{{ file.name }}
-					{{ t('deck', 'already exists.') }}
+					{{ t('deck', 'A file with the name {filename} already exists.', {filename: file.name}) }}
 				</p>
 				<p>
 					{{ t('deck', 'Do you want to overwrite it?') }}
 				</p>
 				<button class="primary" @click="overrideAttachment">
-					{{ t('deck', 'Yes') }}
+					{{ t('deck', 'Overwrite file') }}
 				</button>
 				<button @click="modalShow=false">
-					{{ t('deck', 'No') }}
+					{{ t('deck', 'Keep existing file') }}
 				</button>
 			</div>
 		</Modal>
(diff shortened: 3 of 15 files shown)


src/components/card/CardSidebarTabAttachments.vue
```diff
@@ -38,11 +45,30 @@ export default {
 			default: null,
 		},
 	},
+	data() {
+		return {
+		}
+  },
+  computed: {
+		attachments() {
+		  return this.$store.getters.attachmentsByCard(this.cardId)
```
juliushaertl: cardId is undefined here. This is why your getter doesn't return anything

src/services/AttachmentApi.js
```diff
@@ -0,0 +1,68 @@
...
+
+	async getAll(cardId) {
+		const response = await axios({
+			method: 'GET',
+			url: this.url(`/cards/${cardId}/attachments`),
+		})
+		return response.data
+	}
+
+	async createAttachment(data) {
```
juliushaertl:
```suggestion
	async createAttachment({ cardId, formData}) {
```

Makes it a bit easier to actually get what is required here. Make sure to adjust the usages below as well then.

src/store/attachment.js
```diff
@@ -0,0 +1,80 @@
...
+
+const apiClient = new AttachmentApi()
+
+export default {
+	state: {
+		attachments: {},
+	},
+	getters: {
+		attachmentsByCard: state => (cardId) => {
+			if (state.attachments[cardId] === undefined) {
```
juliushaertl: Please use a typeof check

src/store/attachment.js
```diff
@@ -0,0 +1,80 @@
...
+				return []
+			}
+			return state.attachments[cardId]
+		},
+
+	},
+	mutations: {
+		createAttachment(state, { cardId, attachments }) {
+			if (state.attachments[cardId] === undefined) {
```
juliushaertl: same here

src/components/card/CardSidebarTabAttachments.vue
```diff
@@ -22,30 +22,99 @@
...
 			default: null,
 		},
 	},
+	data() {
+		return {
+		}
+	},
+	computed: {
+		attachments() {
+		  return this.$store.getters.attachmentsByCard(this.card.id)
```
juliushaertl: Use tabs not spaces :wink:

src/store/attachment.js
```diff
@@ -0,0 +1,80 @@
...
+		deleteAttachment(state, deletedAttachment) {
+			const existingIndex = state.attachments[deletedAttachment.cardId].findIndex(a => a.id === deletedAttachment.id)
+			if (existingIndex !== -1) {
+				state.attachments[deletedAttachment.cardId].splice(existingIndex, 1)
+			}
+		},
+
+	},
+	actions: {
+		async getAllAttachments({ commit }, cardId) {
```
juliushaertl: I would prefer a naming that is more like a action, maybe fetchAttachments, as I think get might lead to confusion that it could be a getter

src/store/card.js
```diff
@@ -159,6 +159,20 @@ export default {
 				existingCard.labels = card.labels
 			}
 		},
+		cardIncreaseAttachmentCount(state, cardId) {
+			const existingIndex = state.cards.findIndex(_card => _card.id === cardId)
+			if (existingIndex !== -1) {
+				const existingCard = state.cards.find(_card => _card.id === cardId)
+				existingCard.attachmentCount++
```
juliushaertl: Is that actually updating instantly then when triggering the mutation. We might need to use Vue.set

src/components/card/CardSidebarTabAttachments.vue
```diff
@@ -22,30 +22,99 @@
...
+				<span class="filesize">{{ attachment.extendedData.filesize }}</span>
+				<span class="filedate">{{ attachment.createdAt }}</span>
+				<span class="filedate">{{ attachment.createdBy }}</span>
+			</div>
+			<Actions>
+				<ActionButton icon="icon-delete" @click="deleteAttachment(attachment)">
+					{{ t('deck', 'Delete Attachment') }}
+				</ActionButton>
+
+				<ActionButton v-if="attachment.extendedData.mimetype.includes('image')" icon="icon-picture" @click="coverToggle(attachment)">
```
juliushaertl: SVGs should not be considered to be savely shown in a webpage if uploaded by a user. And as discussed, I'd prefer this to be in a separte PR so we can make sure we use previews instead of the full files when showing the cover.

src/components/card/CardSidebarTabAttachments.vue
```diff
@@ -22,30 +22,99 @@
...
+					{{ t('deck', 'Use as cover') }}
+				</ActionButton>
+			</Actions>
+		</li>
 		<button class="icon-upload" @click="clickAddNewAttachmment()">
 			{{ t('deck', 'Upload attachment') }}
 		</button>
+		<input ref="localAttachments"
+			type="file"
+			multiple
```
juliushaertl: Multiple is not handled by the backend properly yet,

src/components/card/CardSidebarTabAttachments.vue
```diff
@@ -22,30 +22,99 @@
...
 	methods: {
 		clickAddNewAttachmment() {
-
+			this.$refs.localAttachments.click()
+		},
+		onLocalAttachmentSelected(e) {
+			const bodyFormData = new FormData()
+			bodyFormData.append('cardId', this.card.id)
+			bodyFormData.append('type', 'deck_file')
+			bodyFormData.append('file', e.target.files[0])
```
juliushaertl: Related to the previous multiple comment, this just uploads the first file. Either the backend needs to support uploading multiple files in one API call or you need to do one for each file in the frontend. If done in the frontend we need some proper progress display for this and queue them somehow.

src/components/card/CardSidebarTabAttachments.vue
```diff
@@ -22,30 +22,99 @@
...
+		attachmentUrl(attachment) {
+			return OC.generateUrl(`/apps/deck/cards/${attachment.cardId}/attachment/${attachment.id}`)
 		},
 	},
 }
 </script>
 
 <style scoped>
-
+.fileicon {
```
juliushaertl: Indentation looks odd

src/store/attachment.js
```diff
@@ -0,0 +1,80 @@
+/*
+ * @copyright Copyright (c) 2018 Julius Härtl <jus@bitgrid.net>
```
juliushaertl: You created those files, make sure to add yourself to the copyright and author header :wink:

src/store/attachment.js
```diff
@@ -0,0 +1,80 @@
...
+
+	},
+	mutations: {
+		createAttachment(state, { cardId, attachments }) {
+			if (state.attachments[cardId] === undefined) {
+				Vue.set(state.attachments, cardId, attachments)
+			} else {
+				state.attachments[cardId].push(attachments)
+			}
+			// let card = getters.cardById(cardId)
```
juliushaertl: :fire:

src/store/attachment.js
jakobroehrl: Like this? if (typeof (state.attachments[cardId]) === 'undefined') {
What is the advantage?

juliushaertl: The current way will throw an error, while when checking with the typeof operator it won't:
image

juliushaertl:
Uploading an attachment fails if it is the first one:
```
vue.esm.js:1897 TypeError: Cannot read property 'mimetype' of undefined
    at VueComponent.mimetypeForAttachment (CardSidebarTabAttachments.vue:94)
    at CardSidebarTabAttachments.vue?f369:15
    at Proxy.renderList (vue.esm.js:2660)
    at Proxy.render (CardSidebarTabAttachments.vue?f369:8)
    at VueComponent.Vue._render (vue.esm.js:3557)
    at VueComponent.updateComponent (vue.esm.js:4075)
    at Watcher.get (vue.esm.js:4488)
    at Watcher.run (vue.esm.js:4563)
    at flushSchedulerQueue (vue.esm.js:4319)
    at Array.<anonymous> (vue.esm.js:1989)
```

- [ ] Clicking the attachment should download it
- [ ] Fix styling to match the one from the previous version
image
- [ ] Card badge icon to big and overlaying the counter 
image

src/services/AttachmentApi.js
```diff
@@ -0,0 +1,70 @@
...
+ */
+
+import axios from '@nextcloud/axios'
+
+export class AttachmentApi {
+
+	url(url) {
+		url = `/apps/deck${url}`
+		// eslint-disable-next-line
+		return OC.generateUrl(url)
```
juliushaertl: please use generateUrl from `@nextcloud/router`

src/services/AttachmentApi.js
```diff
@@ -0,0 +1,70 @@
...
+	async createAttachment({ cardId, formData }) {
+		try {
+			const response = await axios({
+				method: 'POST',
+				url: this.url(`/cards/${cardId}/attachment`),
+				data: formData,
+			})
+			return response.data
+		} catch (e) {
+			// eslint-disable-next-line
```
juliushaertl: No need to igore this line, console.error is ok ;)

src/services/AttachmentApi.js
```diff
@@ -0,0 +1,70 @@
...
+
+	async createAttachment({ cardId, formData }) {
+		try {
+			const response = await axios({
+				method: 'POST',
+				url: this.url(`/cards/${cardId}/attachment`),
+				data: formData,
+			})
+			return response.data
+		} catch (e) {
```
juliushaertl: I would prefer to have this catch being thrown to the vue component that calls the vuex action, so we can show a proper warning in the frontend that the upload has failed.

src/store/attachment.js
```diff
@@ -0,0 +1,77 @@
...
+		attachmentsByCard: state => (cardId) => {
+			if (typeof state.attachments[cardId] === 'undefined') {
+				return []
+			}
+			return state.attachments[cardId]
+		},
+
+	},
+	mutations: {
+		createAttachment(state, { cardId, attachments }) {
```
juliushaertl: Naming it attachment would be more obvious since it is just about one single attachment.

juliushaertl:
I'm fine to handle those in a separate PR but the remaining would be good to have before merging this PR:

> drag and drop
> if attachment is already there, ask if update is required
> restore removed attachment

src/services/AttachmentApi.js
jakobroehrl: Make sense, how can I do it?

src/store/attachment.js
jakobroehrl: This mutation is called to generate all attachment entries on card load, too.

src/services/AttachmentApi.js
juliushaertl: If you want to keep the logging in the AttachmentApi just add a `throw e` after the log entry. That makes sure that the error continues to propagate out of the function.

src/store/attachment.js
juliushaertl: Then we should split the mutation into two I'd say. We could get into some weired state with the current implementation if you call createAttachment for some reason with an array of attachments twice.

lib/Exceptions/ConflictException.php
```diff
@@ -0,0 +1,36 @@
...
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU Affero General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Affero General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\Deck;
```
juliushaertl:
```suggestion
namespace OCA\Deck\Exceptions;
```

lib/Exceptions/ConflictException.php
```diff
@@ -0,0 +1,36 @@
...
+
+
+class ConflictException extends \Exception {
```
juliushaertl:
```suggestion
class ConflictException extends StatusException {
```

lib/Service/FileService.php
```diff
@@ -42,6 +42,7 @@
 use OCP\IL10N;
 use OCP\ILogger;
 use OCP\IRequest;
+use OCP\Deck\ConflictException;
```
juliushaertl:
```suggestion
use OCA\Deck\Exceptions\ConflictException;
```

lib/Exceptions/ConflictException.php
juliushaertl: + use statement ;)

ostasevych:
Please, add an opportunity to attach documents from the storage.

jakobroehrl:
> 
> Please, add an option to attach documents from the storage.

It's planned with NC19

src/components/card/CardSidebarTabAttachments.vue
```diff
@@ -88,7 +88,7 @@ export default {
 		return {
 			modalShow: false,
 			file: '',
-			overrideError: null
+			OverrideAttachment: null
```
juliushaertl:
```suggestion
			overwriteAttachment: null
```

src/components/card/CardSidebarTabAttachments.vue
```diff
@@ -22,30 +22,254 @@
...
+					</Actions>
+				</li>
+			</ul>
+		</div>
+
+		<Modal v-if="modalShow" title="File already exists" @close="modalShow=false">
+			<div class="modal__content">
+				<h2>{{ t('deck', 'File already exists') }}</h2>
+				<p>
+					{{ t('deck', 'A file with the name') }}
```
juliushaertl: Constructing language strings like this doesn't work in all languages. Please use a placeholder like 
```suggestion
					{{ t('deck', 'A file with the name {filename} already exists.', {filename: file.name}) }}
```

src/components/card/CardSidebarTabAttachments.vue
```diff
@@ -22,30 +22,254 @@
...
+
+						<ActionButton v-else icon="icon-history" @click="restoreAttachment(attachment)">
+							{{ t('deck', 'Restore Attachment') }}
+						</ActionButton>
+					</Actions>
+				</li>
+			</ul>
+		</div>
+
+		<Modal v-if="modalShow" title="File already exists" @close="modalShow=false">
```
juliushaertl: Please translate that as well

src/components/card/CardSidebarTabAttachments.vue
```diff
@@ -22,30 +22,254 @@
...
+				<p>
+					{{ t('deck', 'A file with the name') }}
+					{{ file.name }}
+					{{ t('deck', 'already exists.') }}
+				</p>
+				<p>
+					{{ t('deck', 'Do you want to overwrite it?') }}
+				</p>
+				<button class="primary" @click="overrideAttachment">
+					{{ t('deck', 'Yes') }}
```
juliushaertl: Yes/no actions are quite bad from usability perspective, please use something like "Overwrite file/Keep existing file"

src/components/cards/CardBadges.vue
```diff
@@ -99,9 +101,10 @@ export default {
 
 		.icon {
 			opacity: 0.5;
-			padding: 12px 3px;
+			padding: 12px 14px;
```
juliushaertl: That change should only be applied to the attachment counter, otherwise the space usage gets to wide for the other icons.

src/store/attachment.js
```diff
@@ -0,0 +1,111 @@
...
+			const existingIndex = state.attachments[attachment.cardId].findIndex(a => a.id === attachment.id)
+			if (existingIndex !== -1) {
+				Vue.set(state.attachments[cardId], existingIndex, attachment)
+			}
+		},
+
+		deleteAttachment(state, deletedAttachment) {
+			const existingIndex = state.attachments[deletedAttachment.cardId].findIndex(a => a.id === deletedAttachment.id)
+			if (existingIndex !== -1) {
+				// state.attachments[deletedAttachment.cardId].splice(existingIndex, 1)
```
juliushaertl: Please remove if no longer needed ;)

src/store/attachment.js
```diff
@@ -0,0 +1,111 @@
...
+				return []
+			}
+			return state.attachments[cardId]
+		},
+
+	},
+	mutations: {
+		createAttachment(state, { cardId, attachment }) {
+			if (typeof state.attachments[cardId] === 'undefined') {
+				Vue.set(state.attachments, cardId, attachment)
```
juliushaertl: This will cause the attachments array be no array any longer. For applying full list on load we have createAttachments.

juliushaertl:
- [ ] Icon in upload button is overlaying the text
image

src/store/attachment.js
jakobroehrl: So, I can remove the lines 43 - 45, right?

juliushaertl: No, we need one muation to add an array of attachments and one to add a single attachment. But the single attachment one (L42) should also handle that the array doesn't exist yet in the store, which is what this if statement is used for.

PR #1526, opened 2020-02-05
Finish activity

juliushaertl:
- Move activities out of the vuex store for now, so we can more easily move them to a dedicated library
- Use vue-richtext for rendering the subject
- Fix filtering for board activites
- Add infinite loading
- Fix issues when switching between cards


juliushaertl:
Hm, worked for me, but I've pushed some code hardening. Could you give it another try?

jakobroehrl:
> 
> 
> Hm, worked for me, but I've pushed some code hardening. Could you give it another try?

Yes, sure. 
Now I get no content back, but please merge it. Sometimes it's only not working on my machine.

PR #1528, opened 2020-02-06
reduce card spacing

jakobroehrl:
* Target version: master 

**Before**
grafik

**After**
grafik

### DIFF
--- a/src/components/board/Board.vue
+++ b/src/components/board/Board.vue
@@ -162,8 +162,7 @@ export default {
 					flex-grow: 1;
 					display: flex;
 					flex-direction: column;
-					padding: 3px;
-					padding-top: 0;
+					padding: 0;
 					/**
 					 * Use this to scroll each stack individually
 					 * This currenly has the issue that the popover menu will be cut off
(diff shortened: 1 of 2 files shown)


juliushaertl:
I think the spacing actually helps here to distinguish the individual tasks, maybe we can just reduce the space if we are in compact mode?

bpcurse:
> I think the spacing actually helps here to distinguish the individual tasks, maybe we can just reduce the space if we are in compact mode?

I wanted to propose this spacing reduction for a long time myself, but somehow always forgot about it.
Of course spacing is needed to distinguish individual tasks, but the actual spacing is a bit too large. This helps a little to better use the available space.

jakobroehrl:
I would discuss about reducing the space even more. 
Here some shots, there is a lot less space:

grafik
grafik
grafik

@jancborchardt 
What are you thinking about this? I personal like it to see more content and have less space

juliushaertl:
Thinking about this a bit more ... actually I changed my mind a bit :see_no_evil: I've pushed another commit to fix the small offset on the left/right.

PR #1530, opened 2020-02-07
Check for board membership on comments entity event

juliushaertl:
I somehow was only checking for the existence of a card but not if the card is actually accessible by the user.


jakobroehrl:
How excatly could I test it?

juliushaertl:
backported with 8accc8a020bd3969b2f8d1c1e5e640a79ac646d4

juliushaertl:
stable-0.6 912f9ab2

PR #1537, opened 2020-02-11
Comments reply

juliushaertl:
### TODO

- [x] Frontend for selecting a comment to reply to
- [x] Backend to fetch nested comments (1 level depth)
- [x] Create endpoint to set parentId on comment
- [x] Move endpoints to OCS route
- [x] API documentation

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -3,6 +3,6 @@ module.exports = {
 		'nextcloud'
 	],
 	rules: {
-		'valid-jsdoc': ['warn'],
+		'valid-jsdoc': ['off'],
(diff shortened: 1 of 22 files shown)


juliushaertl:
@jancborchardt Could you have a look if you have some design feedback on this one? This is the current state:

image

juliushaertl:
Let's merge this so we can continue with the other pull requests that base on this. 

@jancborchardt Your feedback would still be appreciated :wink:

sudwhiwdh:
It has already been merged and is still on the roadmap for 1.5?

PR #1545, opened 2020-02-16
Expose boards as calendars

juliushaertl:
Implements https://github.com/nextcloud/deck/issues/15

- [x] Requires https://github.com/nextcloud/server/pull/19196 so 19+ only change
- [x] We could switch directly to VTODO entries if https://github.com/nextcloud/calendar/pull/1798 is ready
- [ ] Open deck from the calendar app https://github.com/nextcloud/calendar/pull/2478

### DIFF
--- a/lib/DAV/Calendar.php
+++ b/lib/DAV/Calendar.php
@@ -27,7 +27,9 @@ use OCA\DAV\CalDAV\Plugin;
 use OCA\DAV\DAV\Sharing\IShareable;
 use OCA\Deck\Db\Board;
 use OCA\Deck\Db\Card;
+use OCA\Deck\Db\Stack;
 use OCA\Deck\Service\CardService;
+use OCA\Deck\Service\StackService;
 use Sabre\CalDAV\Xml\Property\SupportedCalendarComponentSet;
 use Sabre\DAV\PropPatch;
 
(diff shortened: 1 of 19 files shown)


raimund-schluessler:
Will Deck then also move to the CalDAV interface / become a CalDAV client?

juliushaertl:
No, until now this is only for exposing cards with duedates to the calendar/tasks app, related to the discussion in  https://github.com/nextcloud/deck/pull/1319#issuecomment-591388534

lib/DAV/Calendar.php
```diff
@@ -0,0 +1,199 @@
...
+				switch ($key) {
+					case '{DAV:}displayname':
+						if (mb_substr($value, 0, strlen('Deck: '))) {
+							$value = mb_substr($value, strlen('Deck: '));
+						}
+						$this->board->setTitle($value);
+						break;
+					case '{http://apple.com/ns/ical/}calendar-color':
+						$this->board->setColor(substr($value, 1));
+						break;
```
juliushaertl, line 181: @georgehrke Is there any standard where the format of this is specified? Are there any other values than hex colors that could occur here?

juliushaertl: ping :wink:

georgehrke: `{http://apple.com/ns/ical/}calendar-color` always contains a hex color. Usually it's `#rrggbb`. macOS calendar is sending `#rrggbbaa` though, so you would have to filter that out.

jakobroehrl (commented):
It's working fine :)
But I saw the due hour is +1h in the calendar. A 12:00 due card is in the calendar show as a 13:00 due event

hitam4450:
Maybe timezone problem ?

juliushaertl:
Hm, I didn't have this when testing, but I pushed a commit that should force UTC interpretation for the duedate so that the clients can take care of the timezone. Could you check if that fixes it for you?

jakobroehrl:
> Hm, I didn't have this when testing, but I pushed a commit that should force UTC interpretation for the duedate so that the clients can take care of the timezone. Could you check if that fixes it for you?

Now it's working.

juliushaertl:
Let's get this in for the next beta :rocket:

Issue #864, opened 2019-01-29
Board jumps up and down when the Title is edited


der-Daniel:
**Describe the bug**
This is and UI issue. The board jumps up and down when the title is edited. This is very distracting as you are currently typing a new name.

**To Reproduce**
1. Create a new board
2. Edit the name of the board.
3. It will jump up and down depending on the alphabetical ordering.

**Expected behavior**
Keep the text field of the board on the same spot, until the edit is over. Then sort it.

**Client details:**
 - OS: Windows
 - Browser: Chrome
 - Version: 0.5.2
 - Device: desktop

juliushaertl:
@jakobroehrl Not yet fixed.

![Peek 2020-01-25 11-03](https://user-images.githubusercontent.com/3404133/73119456-6b6ef480-3f62-11ea-9836-0d8a36c3fdac.gif)

PR #1576, opened 2020-03-07
Fix card sizing when editing

juliushaertl:
Fixes https://github.com/nextcloud/deck/issues/864


juliushaertl:
Thanks for the notice, pushed a fix for that as well.

Issue #1597, opened 2020-03-12
Attachments - missing pieces


juliushaertl:
- [x] Drag upload #1601
- [x] Check for file size before sending the upload request #1600
- [x] Drag upload on individual cards (Ideally we should move this handling into a mixin and the overlay into a separate component, since then we can reuse this for the CardItem where dragging files to should also work.) #1610 
- [x] Progress indicator and queue for uploads

PR #1601, opened 2020-03-12
drag and drop on attachment tab

jakobroehrl:
* Target version: master


src/components/card/CardSidebarTabAttachments.vue
```diff
@@ -59,6 +62,20 @@
...
 
+		<transition name="slide" mode="out-in">
+			<div
+				v-show="isDraggingOver"
+				class="dragover">
+				<div class="drop-hint">
+					<div class="drop-hint__icon icon-upload" />
+					<h2
+						class="drop-hint__text">
+						text
```
juliushaertl: :wink:

src/components/card/CardSidebarTabAttachments.vue
```diff
@@ -136,15 +154,24 @@ export default {
...
+		},
+		handleUploadFile(event) {
+			this.onLocalAttachmentSelected(event.target.files[0])
+			event.target.value = ''
+		},
 		clickAddNewAttachmment() {
 			this.$refs.localAttachments.click()
 		},
-		async onLocalAttachmentSelected(event) {
+		async onLocalAttachmentSelected(file) {
```
juliushaertl: Please also add a check if the user has edit permissions on a board here, that is a leftover from the first attachments pr.

src/components/card/CardSidebarTabAttachments.vue
```diff
@@ -136,15 +154,24 @@ export default {
 		this.$store.dispatch('fetchAttachments', this.card.id)
 	},
 	methods: {
+		handleDropFiles(event) {
```
juliushaertl: Ideally we should move this handling into a mixin and the overlay into a separate component, since then we can reuse this for the `CardItem` where dragging files to should also work. But feel free to just finish this for the sidebar first and then we do the card handling in another pr.

jakobroehrl: let's do it in another PR.
Could you do a quick review, please? And merge if it's ok.

juliushaertl: Added to the list of https://github.com/nextcloud/deck/issues/1597

juliushaertl:
Pushed a small cleanup commit that also replaced the slide in transition with a simple fade and reduce opacity flickering since that seems a bit to much to me.

PR #1609, opened 2020-03-15
French translation update

KevID:
A small correction of the French translation.


juliushaertl:
Thank you very much for your pull request. We handle translations at https://www.transifex.com/nextcloud/nextcloud/dashboard/ so even if we merge your changes they would be overwritten automatically by the translations from transifex. Could you submit them over there?

PR #1652, opened 2020-04-01
Progress indicator for attachment uploads

jakobroehrl:
* Resolves: #1597 
* Target version: master

### DIFF
--- a/src/services/AttachmentApi.js
+++ b/src/services/AttachmentApi.js
@@ -38,18 +38,10 @@ export class AttachmentApi {
 	}
 
 	async createAttachment({ cardId, formData, onUploadProgress }) {
-		const config = {
-			onUploadProgress: function(progressEvent) {
-			  const percentCompleted = Math.round((progressEvent.loaded * 100) / progressEvent.total)
-			  console.log(percentCompleted)
-			},
-		}
-
 		const response = await axios({
 			method: 'POST',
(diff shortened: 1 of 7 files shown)


jakobroehrl:
@juliushaertl 
This doesn't work so far. 
Do you see an error?

src/services/AttachmentApi.js
```diff
@@ -38,10 +38,18 @@ export class AttachmentApi {
...
+			  const percentCompleted = Math.round((progressEvent.loaded * 100) / progressEvent.total)
+			  console.log(percentCompleted)
+			},
+		}
+
 		const response = await axios({
 			method: 'POST',
 			url: this.url(`/cards/${cardId}/attachment`),
 			data: formData,
+			config: config,
```
juliushaertl, line 52:
```suggestion
			...config,
```

As the first parameter to axios() is the config already.

jakobroehrl: Thanks, now I get a console log result. 
How can I access this value from the view components?

juliushaertl: Well the easiest would be to pass a callback to the service method and then being able to access it directly in the component.

jakobroehrl:
@juliushaertl 
Could you check this please, thanks.

jakobroehrl:
@juliushaertl 
Looks good, let's merge this in

Issue #1561, opened 2020-02-26
Integrate Nextcloud Text as a markdown editor


juliushaertl:
As discussed with @jakobroehrl we should make use of the capabilities that Nextcloud Text offers as a nicer way of markdown editing.

To achieve this it would be great if we can extract the actual editor parts of text into a dedicated library that can then be used by apps. As a first step this could be done without collaborative editing, since that would either require a custom backend in deck or extending the text app to support other sources than files.

~The basic entry point would be some refactored version of the [EditorWrapper](https://github.com/nextcloud/text/blob/master/src/components/EditorWrapper.vue), but we need to make sure that the [collaboration plugin](https://github.com/nextcloud/text/blob/master/src/components/EditorWrapper.vue#L311-L323) and [sync service handling](https://github.com/nextcloud/text/blob/master/src/components/EditorWrapper.vue#L263-L279) can be injected from outside (by the text app) if needed.~

Ongoing effort on moving the editor parts into a shared npm library: https://github.com/nextcloud/text/pull/2278

ShogunWeb:
As a simple user, I would greatly appreciate to have the kind of WYSIWYG interface that Nextcloud Text offers. Well, any kind of "styling" menu would be great to me, as I'm not very familiar to markdown notation...

putt1ck:
V1 (currently in beta) provides a link to Markdown notation, which is really easy once you've used it a few times. Ends up here: https://commonmark.org/help/

joreiff:
Currently, we have the ability to toggle between the fully rendered and the plain text version (with some markup) of the description. This has the advantage that it is easy to copy/paste the description to/from other apps supporting markdown, e.g. Tasks. Would Nextcloud Text integration remove this separation? From https://github.com/nextcloud/text/issues/439 it seems like Text will not have a source view anytime soon. Or would it still be possible to have a raw view?

mgielda:
I second @joreiff 's question - e.g. the files_markdown app is actually very cool in that it has this source editing with preview setup, and the lack of such capability in Text is a huge blocker for our use case. We believe there is huge value in plain text (ability to diff using simple tools, version in git, simple interop with many other tools), and collaborative editing of plain text would be ideal, as it transcends many use cases (including not only simple documents but also e.g. technical documentations). So integrating the Nextcloud Text approach (with the "prosemirror cannot into plaintext" limitation) into more apps actually breaks them from our point of view...

mgielda:
[...]

PR #1663, opened 2020-04-06
Introduce @nextcloud/text as editor

juliushaertl:
This will replace the easymde editor with a variant of Nextcloud Text, currently being developed as a separate library in https://github.com/juliushaertl/nextcloud-text

As this is not yet released you will need to link the library manually for running this PR.


jakobroehrl:
https://github.com/nextcloud/deck/issues/1561

Issue #1344, opened 2019-11-23
API board creation allowed without title


tcassaert:
**Describe the bug**
When creating a board via the API, it is possible to create one, without providing a title, with an empty string. This is not possible in the UI. 

**To Reproduce**
1. Make a Post request with an empty title
2. Board got created, with a blank title

**Expected behavior**
I'd expect an error, when there is an empty string provided

stefan-niedermann:
@juliushaertl can you please explain this a little more?

I saw `<length>255</length>` and `const TITLE_MAX_LENGTH = 100;`. Did the max title length change or not? If yes: Beginning with which version? (Also: The `title` max length should be documented in the REST-API since the `message` max length is also documented there to be `1000`).

PR #1729, opened 2020-04-20
Add length documentation for title attributes in the API documentation

juliushaertl:
https://github.com/nextcloud/deck/issues/1344#issuecomment-616396716

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -117,7 +117,7 @@ Returns an array of board items
 
 | Parameter | Type   | Description                                          |
 | --------- | ------ | ---------------------------------------------------- |
-| title     | String | The title of the new board                           |
+| title     | String | The title of the new board, maximum length is limited to 100 characters |
 | color     | String | The hexadecimal color of the new board (e.g. FF0000) |
 
 ```json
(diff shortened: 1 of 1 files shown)


stefan-niedermann:
Okay, so

- Titles of Stacks and Boards is now limited to `100` characters, that's fine
- Titles of Cards are now extended to `255` characters, that's a breaking change

If a client sends a title with `255` characters to an older version of the Deck API, it will probably break with some kind of SQL-error, because it was previously limited to `100` characters, won't it?

juliushaertl:
> Titles of Cards are now extended to 255 characters, that's a breaking change

Well kind of, but with the old release exceeding the character limit would always cause an exception, that is why i considered it a bugfix. :wink:

stefan-niedermann:
Bugfix: maybe. But this doesn't change that **this is a breaking change**. You only have to support one API version at your frontend, 3rd-party-clients must be able to handle multiple different versions.

For example here two problematic szenarios:

1. New version gets installed. Android client doesn't change anything, user enters a title > `100` characters via web interface
Best case: Title gets truncated at next sync to 100 characters, worst case: somewhere a SQL-exception get's thrown, because a client has a `VARCHAR(100)` type for this column. damnation. doom. death.

2. Old version stays installed. Android client changes the title limit to `255`, user enters a title > `100` characters via android app
Server responds, that the title cannot be accepted

We have to make a differentiation of the available server API in any case.

**A breaking change is fine and absolutely valid for a new major version. But we need to know about breaking changes so we can *handle* them. Otherwise it would cause strange errors which are badly reproducible and to mad users asking for support :wink: .**

juliushaertl:
Agreed, yes I'll try to improve communication in that regards. 

In theory we also have an API version in the URL which is currently hardcoded to 1.0, so we could also use that to push breaking changes, but I'd generally prefer to just make this depend on the app major/minor release version and put the changes in a proper log. Would that work for you?

stefan-niedermann:
Thank you, yes, i agree (we already have the server app version logic and i would like to avoid to handle another version.

So, which server version will contain this change? 1.0.0?

juliushaertl:
Yes, 1.0.0 will be the first stable release to contain this (it can already be tested with the beta releases). I#ve also pushed a commit to list the two changes that could affect you for now. I'll keep that updated then and mention you if new changes come up.

Issue #1673, opened 2020-04-10
v1.0.0-beta1 Bug report: No images in Description Box 


tobealive:
Bug as titled. I See is the Markdown code, but no image is displayed in the card details descption as it was before.

Issue #1709, opened 2020-04-17
markdown formatting (bold, italic, check boxes) in card description gone with 1.0.0 beta


benjamindahl:
**Describe the bug**
Before the deck app was updated last week we had formatted check boxes in the card description. Now it appears only as markdown syntax.

**To Reproduce**
1. Go to ' Deck app'
2. Click on 'Any Card'
3. View 'Description with markdown syntax'

**Expected behavior**
See markdown formatted Text

**Screenshots**

![Bildschirmfoto 2020-04-17 um 09 44 35](https://user-images.githubusercontent.com/2698024/79545067-5e106480-8090-11ea-8021-1fd34e0ddca8.png)

**Client details:**
 - OS: macOS 10.12
 - Browser  Chrome

**Server details:**
 - nextcloud 18.0.4 RC1 (snap version)

juliushaertl:
cc @jakobroehrl 

We actually should adapt https://github.com/nextcloud/notes/pull/303 / https://github.com/nextcloud/notes/pull/419 until #1663 is ready

juliushaertl:
For the first issue, yes we will always show markdown syntax for now so users don't need to switch between a preview and the editor. The barrier of using markdown will be solved by #1663

benjamindahl:
I think writing in markdown is ok but it should then switch to preview automatically.
To use nextcloud text editor is a good idea.

cheers

b

PR #1731, opened 2020-04-20
Use separate preview for description and allow toggling checkboxes

juliushaertl:
image

Fixes https://github.com/nextcloud/deck/issues/1673 and parts of https://github.com/nextcloud/deck/issues/1709

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3857,7 +3857,6 @@
 			"version": "1.0.10",
 			"resolved": "https://registry.npmjs.org/argparse/-/argparse-1.0.10.tgz",
 			"integrity": "sha512-o5Roy6tNG4SL/FOkCAN6RzjiakZS25RLYFrcMttJqbdd8BWrnA+fGz57iN5Pb06pvBGvl5gQ0B48dJlslXvoTg==",
-			"dev": true,
 			"requires": {
 				"sprintf-js": "~1.0.2"
 			}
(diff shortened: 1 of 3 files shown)


jakobroehrl:
@juliushaertl what is the markdown code to generate a checkbox?

putt1ck:
The use of 2 square brackets with a space between is how the old editor did it (and also I think github MD) [ ] `[ ]`

jakobroehrl:
> 
> 
> The use of 2 square brackets with a space between is how the old editor did it (and also I think github MD) `[ ]`

Tried it so, doesn't work for me.

stefan-niedermann:
`- [ ]`

try `minus space bracket space bracket`

putt1ck:
> > The use of 2 square brackets with a space between is how the old editor did it (and also I think github MD) `[ ]`
> 
> Tried it so, doesn't work for me.

Testing on v1b1, the card view shows a check list when I add something like

[ ] an item
[ ] another item

(at this point the card shows a :white_check_mark: 0/2)

But the view of description continues to show just `[ ] an item` and the only way to check one is replace the space with an x

putt1ck:
> `- [ ]`
> 
> try `minus space bracket space bracket`

I tried that, which makes it autolist while editting (useful :) ) but still doesn't render in a way that allows a user to mark the check box by clicking it.

juliushaertl:
> I tried that, which makes it autolist while editting (useful :) ) but still doesn't render in a way that allows a user to mark the check box by clicking it.

This is what this pull request implements for the preview mode. However there is still a small issue, since the library we use for checkmarks in markdown only supports list style checkboxes `- [x]` not independent ones `[x]`, so we need to adjust the regex for the card badges as I don't see a way to support both right now.

juliushaertl:
Peek 2020-04-21 08-35

jakobroehrl:
OK, this is the way it works:
- [ ] x
- [x] x

PR #1739, opened 2020-04-21
Optimize filter icon

juliushaertl:


jakobroehrl:
It still looks like before

juliushaertl:
Yeah, just reducing the size ;)

juliushaertl:
Is that an approval then? :wink:

Issue #1736, opened 2020-04-20
Button to add attachment to the description is missing


juliushaertl:
This has been lost during the rewrite:

![](https://user-images.githubusercontent.com/12276250/56885038-bdcec380-6a6b-11e9-9d42-5f8ad9ecd73d.png)

@jakobroehrl Do you have time to look into that?

Related issue with a bug from the old version https://github.com/nextcloud/deck/issues/907

PR #1740, opened 2020-04-21
add attachments to desc

jakobroehrl:

* Resolves: https://github.com/nextcloud/deck/issues/1736
* Target version: master 

### TODO
- [x] CSS
- [x] "this.markdownEditor is undefined" error on adding the attachment

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13962,7 +13962,7 @@
 		},
 		"os-homedir": {
 			"version": "1.0.2",
-			"resolved": "https://registry.npmjs.org/os-homedir/-/os-homedir-1.0.2.tgz",
+			"resolved": "http://registry.npmjs.org/os-homedir/-/os-homedir-1.0.2.tgz",
 			"integrity": "sha1-/7xJiDNuDoM94MFox+8VISGqf7M=",
 			"dev": true
 		},
@@ -13979,7 +13979,7 @@
 		},
 		"os-tmpdir": {
 			"version": "1.0.2",
-			"resolved": "https://registry.npmjs.org/os-tmpdir/-/os-tmpdir-1.0.2.tgz",
+			"resolved": "http://registry.npmjs.org/os-tmpdir/-/os-tmpdir-1.0.2.tgz",
 			"integrity": "sha1-u+Z0BseaqFxc/sdm/lc0VV36EnQ=",
 			"dev": true
 		},
@@ -14181,7 +14181,7 @@
 		},
 		"path-is-absolute": {
 			"version": "1.0.1",
-			"resolved": "https://registry.npmjs.org/path-is-absolute/-/path-is-absolute-1.0.1.tgz",
+			"resolved": "http://registry.npmjs.org/path-is-absolute/-/path-is-absolute-1.0.1.tgz",
 			"integrity": "sha1-F0uSaHNVNP+8es5r9TpanhtcX18="
 		},
 		"path-is-inside": {
@@ -17293,7 +17293,7 @@
 		},
 		"string_decoder": {
 			"version": "1.1.1",
-			"resolved": "https://registry.npmjs.org/string_decoder/-/string_decoder-1.1.1.tgz",
+			"resolved": "http://registry.npmjs.org/string_decoder/-/string_decoder-1.1.1.tgz",
 			"integrity": "sha512-n/ShnvDi6FHbbVfviro+WojiFzv+s8MPMHBczVePfUpDJLwoLT0ht1l4YwBCbi8pJAveEEdnkHyPyTP/mzRfwg==",
 			"requires": {
 				"safe-buffer": "~5.1.0"
--- a/src/components/card/AttachmentList.vue
+++ b/src/components/card/AttachmentList.vue
@@ -37,9 +37,18 @@
 						<span class="filedate">{{ attachment.createdBy }}</span>
 					</a>
 				</div>
-				<Actions>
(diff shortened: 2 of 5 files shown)


src/components/card/CardSidebar.vue
```diff
@@ -170,11 +175,46 @@
...
+						{{ t('deck', 'Cancel') }}
+					</ActionButton>
+				</Actions>
+
+				<div class="attachment-list">
+					<ul>
+						<li v-for="attachment in attachments"
+							:key="attachment.id"
+							class="attachment">
+							<a class="fileicon" :style="mimetypeForAttachment(attachment.extendedData.mimetype)" :href="attachmentUrl(attachment)" />
```
juliushaertl, line 192: Please move that into a separate component so it can be reused. Right now this is just duplicating logic we already have in the attachments sidebar.

src/components/card/CardSidebar.vue
```diff
@@ -330,6 +390,14 @@ export default {
 		hideEditor() {
 			this.descriptionEditing = false
 		},
+		showAttachmentModal() {
+			this.modalShow = true
+		},
+		addAttachment(attachment) {
+			const string = '![📎 ' + attachment.data + '](' + this.attachmentUrl(attachment) + ')'
+			this.markdownEditor.easymde.value = string
```
juliushaertl, line 398:
```suggestion
			this.$refs.markdownEditor.easymde.value(string)
```

src/components/card/CardSidebar.vue
```diff
@@ -330,6 +390,14 @@ export default {
 		hideEditor() {
 			this.descriptionEditing = false
 		},
+		showAttachmentModal() {
+			this.modalShow = true
+		},
+		addAttachment(attachment) {
+			const string = '![📎 ' + attachment.data + '](' + this.attachmentUrl(attachment) + ')'
```
juliushaertl, line 397: Append this to the existing description with a newline before inserting the attachment. IIrc we also had a check for the mimetype in the old version and just added a link (no embedded image) for other filetypes.

src/components/card/CardSidebar.vue
```diff
@@ -170,11 +175,46 @@
 			icon="icon-activity">
 			<CardSidebarTabActivity :card="currentCard" />
 		</AppSidebarTab>
+		<Modal v-if="modalShow" :title="t('deck', 'Choose attachment')" @close="modalShow=false">
```
juliushaertl, line 178: This could actually also be a separate componen as the CardSidebar is already quite huge.

src/components/card/CardSidebar.vue
```diff
@@ -127,6 +127,11 @@
 					href="https://deck.readthedocs.io/en/latest/Markdown/"
 					target="_blank"
 					class="icon icon-info" />
+				<Actions v-if="canEdit">
+					<ActionButton v-if="descriptionEditing" icon="icon-attach" @click="showAttachmentModal()">
```
juliushaertl, line 131: Something is wrong with the sizing of that icon make sure that the background iamge is only 16x16px.

src/components/card/AttachmentList.vue
```diff
@@ -0,0 +1,223 @@
...
+		Actions,
+		ActionButton,
+	},
+	mixins: [relativeDate],
+	props: {
+		cardId: {
+			type: Number,
+			required: true,
+		},
+		editor: {
```
juliushaertl, line 76: Hm, I'm not a big fan of that appraoch as this bundles the AttachmentList with the implementation of the editor. Having an event would be much nicer here

src/components/card/CardSidebar.vue
```diff
@@ -542,4 +584,101 @@ export default {
...
+		text-align: center;
+		margin: 20px 20px 60px 20px;
+	}
+
+	.modal__content button {
+		float: right;
+		margin: 40px 3px 3px 0;
+	}
+
+	.attachment-list {
```
juliushaertl, line 600: Can be removed now as this is in the AttachmentList component.

src/components/card/CardSidebarTabAttachments.vue
```diff
@@ -172,19 +135,6 @@ export default {
...
-		text-align: center;
-		margin: 20px 20px 60px 20px;
-	}
-
-	.modal__content button {
-		float: right;
-		margin: 40px 3px 3px 0;
-	}
-
 	.attachment-list {
```
juliushaertl, line 138: This can also be removed.

jakobroehrl: Are you sure? There is still an attachment-list entry: 
https://github.com/nextcloud/deck/blob/a7473e9ec9bca51c435a36bfd4255e17e5291f22/src/components/card/CardSidebarTabAttachments.vue#L33

juliushaertl: Ah, I thought that would already use the new AttachmentList component. Then OK, we can also migrate that later on, maybe you can add a TODO comment for that here.

package-lock.json
```diff
@@ -13068,7 +13068,7 @@
 		},
 		"os-homedir": {
 			"version": "1.0.2",
-			"resolved": "https://registry.npmjs.org/os-homedir/-/os-homedir-1.0.2.tgz",
```
juliushaertl, line 13071: Please don't commit such changes :wink: I assume you ran `npm install` instead of `npm ci`.

PR #1751, opened 2020-04-22
Add animations and pin create card input

juliushaertl:
Peek 2020-04-22 18-03

- Make card input sticky at the top
- Keep card input open when submitting
- Open card directly after creation
- Adding some small animations for loading board, adding cards, showing card input, adding/removing labels/duedate

cc @nextcloud/designers

### DIFF
--- a/src/components/board/Board.vue
+++ b/src/components/board/Board.vue
@@ -23,26 +23,29 @@
 <template>
 	<div class="board-wrapper">
 		<Controls :board="board" />
-		<div v-if="loading" class="emptycontent">
-			<div class="icon icon-loading" />
-			<h2>{{ t('deck', 'Loading board') }}</h2>
-			<p />
-		</div>
-		<div v-else-if="board" class="board">
-			<Container lock-axix="y"
-				orientation="horizontal"
-				:drag-handle-selector="dragHandleSelector"
-				@drop="onDropStack">
-				<Draggable v-for="stack in stacksByBoard" :key="stack.id">
-					<Stack :stack="stack" />
-				</Draggable>
-			</Container>
-		</div>
-		<div v-else class="emptycontent">
-			<div class="icon icon-deck" />
-			<h2>{{ t('deck', 'Board not found') }}</h2>
-			<p />
-		</div>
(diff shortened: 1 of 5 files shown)


src/components/board/Board.vue
```diff
@@ -23,26 +23,29 @@
...
-				</Draggable>
-			</Container>
-		</div>
-		<div v-else class="emptycontent">
-			<div class="icon icon-deck" />
-			<h2>{{ t('deck', 'Board not found') }}</h2>
-			<p />
-		</div>
+		<transition name="fade" mode="out-in">
+			<div v-if="loading" class="emptycontent" key="loading">
```
skjnldsv, line 27: key? Are they not only needed in transition group?

src/components/board/Stack.vue
```diff
@@ -78,7 +82,11 @@
 			@should-accept-drop="canEdit"
 			@drop="($event) => onDropCard(stack.id, $event)">
 			<Draggable v-for="card in cardsByStack" :key="card.id">
-				<CardItem v-if="card" :id="card.id" />
+				<transition :appear="animate && !card.animated && (card.animated=true)"
```
skjnldsv, line 85: you can do assignments inline? I remember looking at it and couldn't see any proper way to do so. Are vuejs guidelines ok with this? :thinking:

src/components/board/Board.vue
juliushaertl: > When toggling between elements that have the same tag name, you must tell Vue that they are distinct elements by giving them unique key attributes

https://vuejs.org/v2/guide/transitions.html#Transitioning-Between-Elements

src/components/board/Stack.vue
juliushaertl: Yeah, not the nicest but i've seen this in quite some places. Also this part of code is not really the nicest, but I cannot tell the transition to only run when a card if freshly added and not on sorting/initial load. :woman_shrugging:

skjnldsv: Then I'm closing my eyes :see_no_evil: :stuck_out_tongue_closed_eyes:

PR #1752, opened 2020-04-22
Add php-cs config, ci check and fix code style

juliushaertl:


.php_cs.dist
```diff
@@ -0,0 +1,18 @@
...
+declare(strict_types=1);
+
+require_once './vendor/autoload.php';
+
+use Nextcloud\CodingStandard\Config;
+
+$config = new Config();
+$config
+	->getFinder()
+//	->ignoreVCSIgnored(true)
```
juliushaertl, line 12: That one fails for me locally when using the phar, did you also encounter an issue with that @ChristophWurst

ChristophWurst: No, I have not. But you won't need it in an app. I added it so server so external apps are excluded. Bur it doesn't work anyway :)

PR #1774, opened 2020-04-27
Remove deprecated global API calls

juliushaertl:
Some cleanup to make eslint complain less and have cleaner code :wink:

### DIFF
--- a/src/BoardSelector.vue
+++ b/src/BoardSelector.vue
@@ -40,6 +40,55 @@
 		</div>
 	</Modal>
 </template>
+<script>
+import Modal from '@nextcloud/vue/dist/Components/Modal'
+import axios from '@nextcloud/axios'
+import { generateUrl } from '@nextcloud/router'
+
+export default {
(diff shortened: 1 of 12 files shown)


src/components/cards/CardBadges.vue
```diff
@@ -145,7 +147,7 @@ export default {
 			this.$store.dispatch('assignCardToUser', {
 				card: this.copiedCard,
 				assignee: {
-					userId: OC.getCurrentUser().uid,
+					userId: getCurrentUser().uid,
```
ChristophWurst, line 150: :warning: the replacement returns `null` if there is no user and the object with UID only when there is one

juliushaertl: Thanks, although that should be fine for now as the whole frontend currently assumes that the user is logged in.

ChristophWurst: Makes sense. You can also use `getCurrentUser()?.uid` to be on the safe side :)

Issue #1776, opened 2020-04-28
clone a stack


jakobroehrl:

jakobroehrl:
Unassigned because PR is ready

juliushaertl:
@jakobroehrl Please avoid comments like that as this causes a notification to everyone subscribed to the issue

TheWatcher:
Are there any plans to implement the stack clone process, even if it's restricted to cloning within a given deck? The lack of this feature is a blocker to the adoption of NextCloud Dock for an organisation I am helping with digital sovereignty work.

Issue #119, opened 2017-05-04
Clone / Copy boards [$150]


JSoko:
Imho it would be very helpful to be able to copy boards with and without their content

So you don't have to create the Markers new for each board and can split boards monthly or weekly

Btw: Thanks for your great work!

juliushaertl:
Makes sense. Would also be nice to have for cards/stacks as well.

plrunner:
+1

monsieurhannes:
We have the same case, similar rutines and way too many tags to write them every time. I don't know if that was discussed already, but it would be also great to have a new tag automatically added to the board when someone writes that in the blank tag field. 
**Anyway, great app!**

rhersel:
We make much use of templates in Kanban boards. If you create a new board, you don't want to design it again an again.

Energiegenossenschaft:
We use nextcloud in our cooperative. We would like to say goodbye to Trello and change completely on deck. But the function to copy boards is missing. We would welcome this extension and are willing to pay an amount of CHF 150. Thank you for your efforts so far!

novadeviator:
Duplicating boards (and stacks and tasks) seems to be a really crucial feature in order for Deck to be really useful for any medium sized project management.

juliushaertl:
@Energiegenossenschaft You can always put a bounty on it at https://www.bountysource.com/issues/44830048-clone-copy-boards That way community contributors who solve the issue will get the money then.

niftylouie:
This feature is absolutely needed. Add templates or Copy or both, and it will become a web based project management solution overnight. I'm an architect licensed in two states who plans to use this system for collaborative project management - one-stop-shop.

charlessteiner:
Agreed. It would be cool. 
What is a rough figure for a bounty for this? Maybe a few of us can contribute.

pieter-groeneweg:
mmm.. yes.. or a way to create templates for recurring projects types.
when creating a new website 90% of the activities is similar...
when clicking new board it would be nice to be able from a list of present templates.

frankmaszinek:
Exactly!

Sent from ProtonMail mobile

-------- Original Message --------

Energiegenossenschaft:
We will push this issue with additional CHF 150. Which means we are now at bounty CHF 300.-. We would extremely welcome a rapid implementation!

juliushaertl:
@Energiegenossenschaft That is awesome :rocket: 

[...]

PR #1777, opened 2020-04-29
clone stacks

jakobroehrl:
* Resolves: #1776 
* Target version: master 

**Todo / to clarify**
- [ ] Ideally we get the number of stacks in total for the board and move it to the end.
- [ ] We could use the card clone implementation and use it here, too: #119 (comment)
- [ ] labels and assigned users clone
- [x] I think it should also work to just operate on $card directly so you just overwrite the properties that need to be changed and then passed to insert afterwards.

---

 1. Labels are only shown after page reload
 2. User assignment doesn't work: "The user is not part of the board"
https://github.com/nextcloud/deck/blob/0dd38f5f2e067ef78aa7dbf3cfc8cb61c6d2d6ca/lib/Service/StackService.php#L451-L455


jakobroehrl:
@juliushaertl 
We could use the card clone implementation and use it here, too: https://github.com/nextcloud/deck/issues/119#issuecomment-578393767
What do you think?

stefan-niedermann:
Please have a look at https://github.com/nextcloud/deck/issues/1832 before you make exact clones of the cards @jakobroehrl

lib/Service/StackService.php
```diff
@@ -362,4 +368,93 @@ public function reorder($id, $order) {
...
+	 * @return Stack
+	 * @throws StatusException
+	 * @throws BadRequestException
+	 */
+	public function clone($id, $boardId) {
+		if (is_numeric($id) === false) {
+			throw new BadRequestException('stack id must be a number');
+		}
+
+		$this->permissionService->checkPermission(null, $boardId, Acl::PERMISSION_MANAGE);
```
juliushaertl, line 384: Please check if the user has read permission on the stack as well

lib/Service/StackService.php
```diff
@@ -362,4 +368,93 @@ public function reorder($id, $order) {
...
+
+	/**
+	 * @param $id
+	 * @param $boardId
+	 * @return Stack
+	 * @throws StatusException
+	 * @throws BadRequestException
+	 */
+	public function clone($id, $boardId) {
+		if (is_numeric($id) === false) {
```
juliushaertl, line 380: Check for a valid board id is missing.

lib/Service/StackService.php
```diff
@@ -362,4 +368,93 @@ public function reorder($id, $order) {
...
+		if ($this->boardService->isArchived(null, $boardId)) {
+			throw new StatusException('Operation not allowed. This board is archived.');
+		}
+
+		$stack = $this->stackMapper->find($id);
+
+		$newStack = new Stack();
+		$newStack->setTitle($stack->getTitle() . ' (' . $this->l10n->t('copy') . ')');
+		$newStack->setBoardId($boardId);
+		$newStack->setOrder($stack->getOrder() +1);
```
juliushaertl, line 394: Ideally we get the number of stacks in total for the board and move it to the end.

lib/Service/StackService.php
```diff
@@ -362,4 +368,93 @@ public function reorder($id, $order) {
...
+
+		$this->activityManager->triggerEvent(
+			ActivityManager::DECK_OBJECT_BOARD, $newStack, ActivityManager::SUBJECT_STACK_CREATE
+		);
+		$this->changeHelper->boardChanged($boardId);
+
+		$this->eventDispatcher->dispatch(
+			'\OCA\Deck\Stack::onCreate',
+			new GenericEvent(null, ['id' => $newStack->getId(), 'stack' => $newStack])
+		);
```
juliushaertl, line 452: Please trigger this right after creating the stack and before the card creating so it is still emitted even if something goes wrong during card creation.

src/components/board/Stack.vue
```diff
@@ -43,6 +43,9 @@
 				<ActionButton icon="icon-delete" @click="deleteStack(stack)">
 					{{ t('deck', 'Delete list') }}
 				</ActionButton>
+				<ActionButton icon="icon-clone" @click="cloneStack(stack)">
```
juliushaertl, line 46: Let's make delete the last option

lib/Service/StackService.php
```diff
@@ -362,4 +368,93 @@ public function reorder($id, $order) {
...
+		$newStack->setTitle($stack->getTitle() . ' (' . $this->l10n->t('copy') . ')');
+		$newStack->setBoardId($boardId);
+		$newStack->setOrder($stack->getOrder() +1);
+		$newStack = $this->stackMapper->insert($newStack);
+
+		$cards = $this->cardMapper->findAll($id);
+		foreach ($cards as $card) {
+			$newCard = new Card();
+			$newCard->setTitle($card->getTitle());
+			$newCard->setStackId($card->getStackId());
```
juliushaertl, line 401: You need to add the cards to the new stack instead of the old one

lib/Service/StackService.php
```diff
@@ -362,4 +368,93 @@ public function reorder($id, $order) {
...
+			$newCard->setType($card->getType());
+			$newCard->setOrder($card->getOrder());
+			$newCard->setOwner($card->getOwner());
```
juliushaertl, line 404: The owner should be the current user rather than the one of the old card.

lib/Service/StackService.php
```diff
@@ -362,4 +368,93 @@ public function reorder($id, $order) {
...
+
+		$newStack = new Stack();
+		$newStack->setTitle($stack->getTitle() . ' (' . $this->l10n->t('copy') . ')');
+		$newStack->setBoardId($boardId);
+		$newStack->setOrder($stack->getOrder() +1);
+		$newStack = $this->stackMapper->insert($newStack);
+
+		$cards = $this->cardMapper->findAll($id);
+		foreach ($cards as $card) {
+			$newCard = new Card();
```
juliushaertl, line 399: I think it should also work to just operate on $card directly so you just overwrite the properties that need to be changed and then passed to insert afterwards.

lib/Service/StackService.php
jakobroehrl: OK. This way I'll get the corresponding board:
$board = $this->boardMapper->find($boardId);
How can I get the stack number of this board?

lib/Db/AssignedLabels.php
```diff
@@ -0,0 +1,38 @@
...
+
+class AssignedLabels extends RelationalEntity implements JsonSerializable {
+	public $id;
+	protected $label;
+	protected $cardId;
+
+	public function __construct() {
+		$this->addType('id', 'integer');
+		$this->addType('cardId', 'integer');
+		$this->addResolvable('label');
```
juliushaertl, line 36: The database returns a field that is called `label_id` so we need a property for that one as well.

lib/Db/AssignedLabelsMapper.php
```diff
@@ -0,0 +1,76 @@
...
+class AssignedLabelsMapper extends DeckMapper  {
+	private $cardMapper;
+	private $userManager;
+	/**
+	 * @var IGroupManager
+	 */
+	private $groupManager;
+
+	public function __construct(IDBConnection $db, CardMapper $cardMapper, IUserManager $userManager, IGroupManager $groupManager) {
+		parent::__construct($db, 'deck_assigned_labels', AssignedLabels::class);
```
juliushaertl, line 41: I think the easiest would be to just reuse the Label class here, since it should already work for the assignment as well. So you could do that and drop the `AssignedLabels` class for now.

lib/Db/AssignedLabelsMapper.php
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+
+	/**
+	 *
+	 * @param Entity $entity
+	 * @return null|Entity
+	 */
+	public function insert(Entity $entity) {
+		$origin = $this->getOrigin($entity);
```
juliushaertl, line 68: No need for that, you should be able to just remove the whole insert method and it will use the parent one then.

ricstg:
ok 


Riccardo De Servi 
Managing Director 

Stellini S.r.l. 
Via Manzoni 54 - 20020 Magnago MI - Italy 
Tel.: +39 0331 072 5 47 
[ http://www.stellinigroup.com/ | www.stellinigroup.com ] 

P Go Green. Think before you print 


Da: "Julius Härtl" <notifications@github.com> 
A: "nextcloud/deck" <deck@noreply.github.com> 
Cc: "Subscribed" <subscribed@noreply.github.com> 
Inviato: Mercoledì, 10 giugno 2020 11:06:17 
Oggetto: Re: [nextcloud/deck] clone stacks (#1777) 





@juliushaertl commented on this pull request. 


In [ https://github.com/nextcloud/deck/pull/1777#discussion_r437975029 | lib/Db/AssignedLabelsMapper.php ] : 
> +		$sql = 'SELECT l.*,card_id FROM `*PREFIX*deck_assigned_labels` as al ' . 
+			'INNER JOIN *PREFIX*deck_labels as l ON l.id = al.label_id ' . 
+			'WHERE `card_id` = ?';
+
+		$labels = $this->findEntities($sql, [$cardId]);
+		return $labels;
+	}
+
+
+	/**
+	 *
+	 * @param Entity $entity
+	 * @return null|Entity
+	 */
+	public function insert(Entity $entity) {
+		$origin = $this->getOrigin($entity); 


No need for that, you should be able to just remove the whole insert method and it will use the parent one then.

jakobroehrl:
Works great! 
Let's merge this in, so I can work on the board clone and card clone topics:
 - https://github.com/nextcloud/deck/issues/119#issuecomment-645178609
 - https://github.com/nextcloud/deck/issues/1813

lib/Service/StackService.php
```diff
@@ -362,4 +379,110 @@ public function reorder($id, $order) {
...
+
+	/**
+	 * @param $id
+	 * @param $boardId
+	 * @param $userId
+	 * @return Stack
+	 * @throws StatusException
+	 * @throws BadRequestException
+	 */
+	public function clone($id, $boardId, $userId) {
```
juliushaertl, line 391: Actually let's remove the userId as a parameter and just inject the current one through the constructor

lib/Service/StackService.php
juliushaertl: Feel free to set the order to 999 which is what we do for new stacks and cards anyways. They will then end up at the end of the stack list sorted by their creation time as a second sorting parameter.

lib/Service/StackService.php
```diff
@@ -362,4 +379,110 @@ public function reorder($id, $order) {
...
+				$l = [];
+				foreach ($assignedLabels as $assignedLabel) {
+
+					$assignment = new AssignedLabels();
+					$assignment->setCardId($newCard->getId());
+					$assignment->setLabelId($assignedLabel->getLabelId());
+					$assignment = $this->assignedLabelsMapper->insert($assignment);
+
+
+					// $assignment = new AssignedLabels();
```
juliushaertl, line 460: :broom: Not needed anymore

lib/Service/StackService.php
```diff
@@ -362,4 +379,110 @@ public function reorder($id, $order) {
...
+			$this->eventDispatcher->dispatch('\OCA\Deck\Card::onCreate',
+				new FTSEvent(
+					null, ['id' => $newCard->getId(), 'card' => $newCard, 'userId' => $owner, 'stackId' => $stackId]
+				)
+			);
+
+			if ($boardId === $stack->getBoardId()) {
+
+				$assignedLabels = $this->assignedLabelsMapper->find($card->getId());
+				$l = [];
```
juliushaertl, line 451: Please rename `$l` to something more meaningful.

lib/Service/StackService.php
```diff
@@ -362,4 +379,110 @@ public function reorder($id, $order) {
...
+					// $assignment = new AssignedLabels();
+					// $assignment->setCardId($newCard->getId());
+					// $assignment->setLabel($assignedLabel);
+					// $assignment = $this->assignedLabelsMapper->insert($assignment);
+					$l[] = $assignment;
+				}
+				$newCard->setLabels($l);
+
+				$assignedUsers = $this->assignedUsersMapper->find($card->getId());
+				$u = [];
```
juliushaertl, line 469: Please give that a better name as well :)

lib/Service/StackService.php
```diff
@@ -362,4 +379,110 @@ public function reorder($id, $order) {
...
+				foreach ($assignedUsers as $assignedUser) {
+					$assignment = new AssignedUsers();
+					$assignment->setCardId($newCard->getId());
+					$assignment->setParticipant($assignedUser->getParticipant());
+					$assignment->setType($assignedUser->getType());
+					$assignment = $this->assignedUsersMapper->insert($assignment);
+					$u[] = $assignment;
+				}
+				$newCard->setAssignedUsers($u);
+				$c[] = $newCard;
```
juliushaertl, line 479: Same for `$c`

lib/Service/StackService.php
```diff
@@ -362,4 +379,110 @@ public function reorder($id, $order) {
...
+	 */
+	public function clone($id, $boardId, $userId) {
+		if (is_numeric($id) === false) {
+			throw new BadRequestException('stack id must be a number');
+		}
+		if (is_numeric($boardId) === false) {
+			throw new BadRequestException('board id must be a number');
+		}
+
+		$this->permissionService->checkPermission(null, $boardId, Acl::PERMISSION_MANAGE);
```
juliushaertl, line 399: Any specific reason for requiring manage permissions and not just write permissions?

src/components/board/Stack.vue
```diff
@@ -40,6 +40,9 @@
 				</form>
 			</transition>
 			<Actions v-if="canManage" :force-menu="true">
+				<ActionButton icon="icon-clone" @click="cloneStack(stack)">
```
juliushaertl, line 43: Depending if we switch to write permissions, this needs to be adjusted as well

src/store/stack.js
```diff
@@ -58,6 +58,9 @@ export default {
 				state.stacks.splice(existingIndex, 1)
 			}
 		},
+		/* cloneStack(state, stack) {
```
juliushaertl, line 61: :broom:

lib/Controller/StackController.php
```diff
@@ -107,4 +107,13 @@ public function delete($stackId) {
...
 		return $this->stackService->fetchDeleted($boardId);
 	}
+
+	/**
+	 * @NoAdminRequired
+	 * @param $boardId
+	 * @return \OCP\Deck\DB\Board
+	 */
+	public function clone($stackId, $boardId) {
+		return $this->stackService->clone($stackId, $boardId, $this->userId);
```
juliushaertl, line 117: Please don't forget our friends from the andorid app and provide a rest api as well ;)

lib/Service/StackService.php
jakobroehrl: @juliushaertl 
You mean "PERMISSION_EDIT", right?

lib/Service/StackService.php
```diff
@@ -362,4 +379,106 @@ public function reorder($id, $order) {
...
+		$this->permissionService->checkPermission(null, $boardId, Acl::PERMISSION_READ);
```
juliushaertl, line 400: Read permission is implied by the manage permisison

lib/Service/StackService.php
```diff
@@ -362,4 +379,106 @@ public function reorder($id, $order) {
...
+		if ($this->boardService->isArchived(null, $boardId)) {
+			throw new StatusException('Operation not allowed. This board is archived.');
+		}
+
```
juliushaertl, line 404: :warning: We also need to verify that the stack is actually part of the board, otherwise this would be a security issue where an attacker could clone a stack of another user to their own board.

juliushaertl: Please do not resolve this comment, I'll do that once verified that this case is covered. :wink:

lib/Service/StackService.php
```diff
@@ -362,4 +379,106 @@ public function reorder($id, $order) {
...
+				foreach ($assignedUsers as $assignedUser) {
+					$assignment = new AssignedUsers();
+					$assignment->setCardId($newCard->getId());
+					$assignment->setParticipant($assignedUser->getParticipant());
+					$assignment->setType($assignedUser->getType());
+					$assignment = $this->assignedUsersMapper->insert($assignment);
+					$newUserArray[] = $assignment;
+				}
+				$newCard->setAssignedUsers($newUserArray);
+				$newCardArray[] = $newCard;
```
juliushaertl, line 475: there is nothing happening with this array

src/components/board/Stack.vue
```diff
@@ -178,6 +181,9 @@ export default {
 		deleteStack(stack) {
 			this.$store.dispatch('deleteStack', stack)
 		},
+		cloneStack(stack) {
+			this.$store.dispatch('cloneStack', stack)
```
juliushaertl, line 185: Some error handling would be great here.

lib/Service/StackService.php
```diff
@@ -48,11 +55,15 @@ class StackService {
...
 	private $assignedUsersMapper;
+	private $assignedLabelsMapper;
 	private $attachmentService;
+
 	private $activityManager;
 	/** @var EventDispatcherInterface */
 	private $eventDispatcher;
 	private $changeHelper;
+	private $l10n;
+	private $assignmentService;
```
juliushaertl, line 66: This is never used.

appinfo/routes.php
```diff
@@ -99,6 +100,7 @@
 		['name' => 'stack_api#create', 'url' => '/api/v1.0/boards/{boardId}/stacks', 'verb' => 'POST'],
 		['name' => 'stack_api#update', 'url' => '/api/v1.0/boards/{boardId}/stacks/{stackId}', 'verb' => 'PUT'],
 		['name' => 'stack_api#delete', 'url' => '/api/v1.0/boards/{boardId}/stacks/{stackId}', 'verb' => 'DELETE'],
+		['name' => 'stack_api#clone', 'url' => '/api/v1.0/boards/{boardId}/stacks/{stackId}/clone', 'verb' => 'POST'],
```
juliushaertl, line 103: Documentation needs an update for this

jakobroehrl: How and where to do this?

src/components/board/Stack.vue
jakobroehrl: ok, what would you like to do here?

lib/Service/StackService.php
jakobroehrl: How to do this?

appinfo/routes.php
juliushaertl: https://github.com/nextcloud/deck/blob/master/docs/API.md

src/components/board/Stack.vue
juliushaertl: So cloning a stack is an operation that everyone with write permissions should be able to do, so this is not the proper place since it is only shown with canManage.

lib/Service/StackService.php
juliushaertl: https://docs.nextcloud.com/server/latest/developer_manual/basics/dependency_injection.html for general details about dependency injection. You'll need to add userId to the constructor

lib/Service/StackService.php
juliushaertl: Well this is not fixed yet. We will should switch this to edit permission.

lib/Service/StackService.php
juliushaertl: Could you also address that?

lib/Service/StackService.php
juliushaertl: Ah wait, we indeed require manage permissions for all stack operations, so this is ok then.

lib/Service/StackService.php
juliushaertl: @jakobroehrl This is not checked yet.

src/components/board/Stack.vue
```diff
@@ -42,11 +42,14 @@
...
 				</form>
 			</transition>
-			<Actions v-if="canManage && !isArchived" :force-menu="true">
-				<ActionButton icon="icon-archive" @click="modalArchivAllCardsShow=true">
+			<Actions v-if="!isArchived" :force-menu="true">
+				<ActionButton v-if="canManage" icon="icon-archive" @click="modalArchivAllCardsShow=true">
 					{{ t('deck', 'Archive all cards') }}
 				</ActionButton>
-				<ActionButton icon="icon-delete" @click="deleteStack(stack)">
+				<ActionButton v-if="canEdit" icon="icon-clone" @click="cloneStack(stack)">
```
juliushaertl, line 49:
```suggestion
				<ActionButton v-if="canManage" icon="icon-clone" @click="cloneStack(stack)">
```

Creating new stacks requires manage permissions so we should do the same with cloning (since this is also what is checked in the backend)

src/components/board/Stack.vue
```diff
@@ -42,11 +42,14 @@
 						value="">
 				</form>
 			</transition>
-			<Actions v-if="canManage && !isArchived" :force-menu="true">
-				<ActionButton icon="icon-archive" @click="modalArchivAllCardsShow=true">
+			<Actions v-if="!isArchived" :force-menu="true">
+				<ActionButton v-if="canManage" icon="icon-archive" @click="modalArchivAllCardsShow=true">
```
juliushaertl, line 46: Archiving all cards is something that would only require edit permissions, maybe you can fix this in the same run.

lib/Service/StackService.php
```diff
@@ -360,4 +376,103 @@ public function reorder($id, $order) {
...
+			$newCard->setStackId($newStack->getId());
+			$newCard->setType($card->getType());
+			$newCard->setOrder($card->getOrder());
+			$newCard->setOwner($this->userId);
+			$newCard->setDescription($card->getDescription());
+			$newCard->setDuedate($card->getDuedate());
+
+			$newCard = $this->cardMapper->insert($newCard);
+
+			$this->activityManager->triggerEvent(ActivityManager::DECK_OBJECT_CARD, $newCard, ActivityManager::SUBJECT_CARD_CREATE);
```
juliushaertl, line 435: I'm not entirely sure if we should have one activity entry for each of the cards or one activity entry that the stack has been cloned, but we can improve that in a later step.

lib/Service/StackService.php
```diff
@@ -360,4 +376,103 @@ public function reorder($id, $order) {
...
+		);
+		$this->changeHelper->boardChanged($boardId);
+
+		$this->eventDispatcher->dispatch(
+			'\OCA\Deck\Stack::onCreate',
+			new GenericEvent(null, ['id' => $newStack->getId(), 'stack' => $newStack])
+		);
+
+		$cards = $this->cardMapper->findAll($id);
+		$newCardArray = [];
```
juliushaertl, line 421: This is never used.

docs/API.md
```diff
@@ -492,6 +492,19 @@ The board list endpoint supports setting an `If-Modified-Since` header to limit
...
+#### Request parameters
+
+| Parameter | Type    | Description                              |
+| --------- | ------- | ---------------------------------------- |
+| boardId   | Integer | The id of the board the stack belongs to |
+| stackId   | Integer | The id of the stack                      |
+
+#### Response
+
+##### 200 Success
```
juliushaertl, line 506: Let's also describe the response here a bit that it will return an object of the new stack containing the new cards as well.

Ideally please add some example JSON of the response.

lib/Db/AssignedLabelsMapper.php
```diff
@@ -0,0 +1,59 @@
...
+		$this->groupManager = $groupManager;
+	}
+
+	/**
+	 *
+	 * @param $cardId
+	 * @return array|Entity
+	 */
+	public function find($cardId) {
+		$sql = 'SELECT * from `*PREFIX*deck_assigned_labels` where `card_id` = ?';
```
juliushaertl, line 53: Could be switched to QBMapper right away to get rid of the deprecated DeckMapper. See https://docs.nextcloud.com/server/18/developer_manual/app/storage/database.html?highlight=querybuilder#mappers for an example

juliushaertl: Feel free to do that in a follow up, but it would be good to not introduce new code that uses legacy functions right away :wink:

docs/API.md
```diff
@@ -502,6 +502,19 @@ The board list endpoint supports setting an `If-Modified-Since` header to limit
 | stackId   | Integer | The id of the stack                      |
 
 #### Response
+It will return an object of the new stack containing the new cards as well.
+
+```json
+  {"title":"l1 (copy)",
```
juliushaertl, line 508: Please pass this into https://jsonformatter.curiousconcept.com/

juliushaertl:
Tests fail :wink:

docs/API.md
```diff
@@ -492,6 +492,93 @@ The board list endpoint supports setting an `If-Modified-Since` header to limit
 
 ##### 200 Success
 
+### POST /boards/{boardId}/stacks/{stackId}/clone - Clone a stack
+
+#### Request parameters
+
+| Parameter | Type    | Description                              |
+| --------- | ------- | ---------------------------------------- |
+| boardId   | Integer | The id of the board the stack belongs to |
```
juliushaertl, line 501:
```suggestion
| boardId   | Integer | The id of the board where the stack should be cloned to |
```

lib/Service/StackService.php
juliushaertl: Nope, as mentioned we also need to check if the original ***stack can be read**

Otherwise a user could basically clone any stack from other users boards to their own ones.

lib/Service/StackService.php
```diff
@@ -365,4 +381,103 @@ public function reorder($id, $order) {
...
+		$stack = $this->stackMapper->find($id);
+		$board = $this->boardMapper->find($boardId);
+
+		if ($stack->getBoardId() !== $board->getId()) {
+			throw new StatusException('Operation not allowed. Stack is not part of this board');
+		}
+
+
+		$newStack = new Stack();
+		$newStack->setTitle($stack->getTitle() . ' (' . $this->l10n->t('copy') . ')');
```
juliushaertl, line 416: We should only add the (copy) if the clone is going to the same board I guess, or ideally if there is another stack with the same title on the destination board.

jakobroehrl: currently cloing is only possible on the same board

doodhout:
There was no reason given why this was never merged and then closed.

Is it possible to pick this up again, by the original committer or by someone else?

Issue #1711, opened 2020-04-17
How to archive all tasks in a column?


patrick-motard:
**Is your feature request related to a problem? Please describe.**
I use three columns; 'todo', 'doing', 'done'. Over time, the done column fills up. In order to archive the issues in the column, I have to archive each issue individually. I think a better user experience would be a button that allows me to archive all issues in a column. Bonus points if it only archives issues older than x date. Super bonus points if it can happen automatically daily.

juliushaertl:
Makes sense to have an archive all button imo. #1721 is also somehow related

jakobroehrl:
basis function will be available in the next version

juliushaertl:
Let's keep this open until the pr is actually merged :wink:

jakobroehrl:
unassign me because PR is ready

stefan-niedermann:
Note: This feature is available in the Android-App since version [`1.5.0`](https://github.com/stefan-niedermann/nextcloud-deck/issues/495)

PR #1853, opened 2020-05-11
archive all cards from stack

jakobroehrl:
* Resolves: #1711
* Target version: master 

grafik


juliushaertl:
@jakobroehrl Enhancements should go to the next minor release so this will need to wait for 1.1.0. :wink:

src/components/board/Stack.vue
```diff
@@ -178,6 +181,11 @@ export default {
 		deleteStack(stack) {
 			this.$store.dispatch('deleteStack', stack)
 		},
+		archiveAllCardsFromStack(stack) {
+			this.cardsByStack.forEach(card => {
```
juliushaertl, line 185: Not fully convinced about this, we should somehow have a progress shown about this and make sure that the changes are not done until all the archive operations have been performed.

stefan-niedermann:
I would switch the entries, like in the boards menu:

grafik

In general we should put the most destructive operations (delete in this case) always at the bottom (Files app does this, too).

jakobroehrl:
> In general we should put the most destructive operations (delete in this case) always at the bottom (Files app does this, too).

done

src/components/board/Stack.vue
```diff
@@ -123,6 +140,8 @@ export default {
 			showAddCard: false,
 			stateCardCreating: false,
 			animate: false,
+			modalShow: false,
+			archiveAllCardsProgress: '',
```
juliushaertl, line 144: Please use a sane default like `null`

src/components/board/Stack.vue
```diff
@@ -51,6 +54,19 @@
 			</Actions>
 		</div>
 
+		<Modal v-if="modalShow" title="Archive all cards in this list" @close="modalShow=false">
+			<div class="modal__content">
+				<h3>Archive all cards in this list</h3>
+				<progress :value="archiveAllCardsProgress" :max="cardsByStack.length" />
```
juliushaertl, line 60: cardsByStack will decrease during the process, so ideally you store the max in a local data field before starting the process

src/components/board/Stack.vue
```diff
@@ -51,6 +54,19 @@
 			</Actions>
 		</div>
 
+		<Modal v-if="modalShow" title="Archive all cards in this list" @close="modalShow=false">
+			<div class="modal__content">
+				<h3>Archive all cards in this list</h3>
```
juliushaertl, line 59: Should be translated

src/components/board/Stack.vue
```diff
@@ -51,6 +54,19 @@
 			</Actions>
 		</div>
 
+		<Modal v-if="modalShow" title="Archive all cards in this list" @close="modalShow=false">
```
juliushaertl, line 57: No need for a title if we have a header in the modal

src/components/board/Stack.vue
```diff
@@ -123,6 +140,9 @@ export default {
 			showAddCard: false,
 			stateCardCreating: false,
 			animate: false,
+			modalShow: false,
```
juliushaertl, line 143: Please give it a more obvious name on what kind of modal this is

src/components/board/Stack.vue
```diff
@@ -123,6 +140,9 @@ export default {
...
+			archiveAllCardsProgress: null,
+			stackLen: 0,
```
juliushaertl, line 145: lets group those into an object:

```
stackTransfer: {
  total: 0,
  current: null,
}
```

src/components/board/Stack.vue
```diff
@@ -299,4 +328,25 @@ export default {
...
 	}
 
+	.modal__content {
+		width: 25vw;
+		min-width: 250px;
+		min-height: 100px;
+		text-align: center;
+		margin: 20px 20px 20px 20px;
+
+		.multiselect {
```
juliushaertl, line 338: Why?

juliushaertl: This still needs to be removed, there is no multiselect in here.

src/components/board/Stack.vue
juliushaertl: @jakobroehrl What about this?

Issue #1778, opened 2020-04-29
Columns not aligned when a title is too long


Clementine46:
When one list has a title that takes more than one line, the list ends up beginning lower than others, leaving the columns unaligned. 

<img width="1140" alt="Capture d’écran 2020-04-29 à 18 56 53" src="https://user-images.githubusercontent.com/50834537/80624002-5915c280-8a4b-11ea-87b2-3be2ad032524.png">

I believe the board would look much cleaner if the columns stayed aligned no matter what.

Here are my two suggestions : 

1. The space for titles could be restricted to a defined number of characters.
Cons : limiting the user's freedom (then again, are there a lot of people using long titles on the deck ?)

2. The columns should stay aligned even if some titles are longer than others.
Cons : Some lists migh end up having a big space between the title and the first card.

Nyco:
3. Let the user type a long column title, but only show one line, show the whole title when hovered

Maybe a title is not enough: a sub-title or level 2 title, or a description. This second text to be shown smaller, which allows to display more text. For example, that could fill one of the key principles of Kanban, which is to explicit rules. Deck (or Trello) missing that feature, results are that in some teams, unmovable cards are kept manually at the top of each column (consequences: introduce erroneous data in reports).

PR #1854, opened 2020-05-11
stack title doesn't break up

jakobroehrl:
* Resolves: #1778
* Target version: master 

**before:**
grafik

**after:**
grafik

### DIFF
--- a/src/components/board/Stack.vue
+++ b/src/components/board/Stack.vue
@@ -28,7 +28,10 @@
 				<h3 v-if="!canManage">
 					{{ stack.title }}
 				</h3>
-				<h3 v-else-if="!editing" @click="startEditing(stack)">
+				<h3 v-else-if="!editing"
+					v-tooltip="stack.title"
+					class="stack--title"
+					@click="startEditing(stack)">
 					{{ stack.title }}
 				</h3>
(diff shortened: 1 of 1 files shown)


src/components/board/Stack.vue
```diff
@@ -29,7 +29,9 @@
 					{{ stack.title }}
 				</h3>
 				<h3 v-else-if="!editing" @click="startEditing(stack)">
-					{{ stack.title }}
+					<p class="stack--title" :title="stack.title">
```
juliushaertl, line 32: please use v-tooltip for this

juliushaertl: Also why is this wrapped in another p tag ? :wink:

src/components/board/Stack.vue
```diff
@@ -247,6 +250,13 @@ export default {
 		}
 	}
 
+	.stack--title {
+		white-space: nowrap;
+		overflow: hidden;
+		text-overflow: ellipsis;
+		max-width: 200px;
```
juliushaertl, line 257: Please use calc to get the maximum width based on `$stack-width` so it continues to work if we change the global stack width at some point.

Issue #1806, opened 2020-05-06
Undefined index: debug


simonspa:
**Describe the bug**
I had a previous version of Deck installed with two boards created. After upgrading I deleted both boards and they are gone from the sidebar but still shown in the main view whenever I open the app.

In the logs I find entries like this:
```
{"reqId":"ZFr3DECCADvL1bVDBDu3","level":3,"time":"2020-05-06T15:34:53+02:00","remoteAddr":"<IP>","user":"me","app":"PHP","method":"GET","url":"/apps/deck/","message":"Undefined index: debug at /var/www/nextcloud/apps/deck/templates/main.php#31","userAgent":"Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:75.0) Gecko/20100101 Firefox/75.0","version":"18.0.4.2","id":"5eb2bd6a1a5d7"}
```

**To Reproduce**
1. Go to Deck
2. Delete all boards
3. Reload app

**Expected behavior**
The boards should really be gone and I wouldn't expect errors in the log.

**Screenshots**
![image](https://user-images.githubusercontent.com/1677436/81184453-b7452700-8fb0-11ea-9929-8996612b97ad.png)

**Client details:**
 - OS: Ubuntu
 - Browser Firefox
 - Version 75
 - Device: desktop

<details>
<summary>Server details</summary>

**Operating system**: Debian Stretch

**Web server:** nginx 1.14.2

**Database:** Mysql

**PHP version:** 7.3

**Nextcloud version:** 18.0.4

**Where did you install Nextcloud from:** manually

</details>

simonspa:
Interestingly even some content still seems to be there:

![image](https://user-images.githubusercontent.com/1677436/81184749-10ad5600-8fb1-11ea-8034-2ad609ee020c.png)

PR #1860, opened 2020-05-11
Handle deleted boards better

juliushaertl:
- [x] Remove unused debug field
- [x] Hide deleted boards from the main content list
- [ ] Make sure that deleted boards are marked as deleted even before the undo timeout

Fix https://github.com/nextcloud/deck/issues/1806


juliushaertl:
>  Make sure that deleted boards are marked as deleted even before the undo timeout


Lets get this in for now and do the remaining point in a follow up to have the already existing stuff in 1.0.1

Issue #1819, opened 2020-05-07
Do not allow user to attempt cloning a board they can't clone


ArnY:
**Describe the bug**
The UI of deck will let the user attempt cloning a boad they do not own. Deck will go as far as showing a new board with a new name, but the board will actually not exist and an error will be displayed "Board not found" when trying to access it.

**To Reproduce**
1. User1 create a board and shares it with User2
2. User2 clones the board. 
3. A new board is being shown under the original board but this board is broken, unusable since User2 does not own the original board. The original board is also broken until the user reload the page.

**Expected behavior**
Board that are not clonable should not be allowed to be cloned, duh.

**Versions:**
Nextcloud 18.0.4
Deck 1.0

PR #1866, opened 2020-05-12
user can only clone a board on canManage permissions

jakobroehrl:
* Resolves: https://github.com/nextcloud/deck/issues/1819
* Target version: master

### DIFF
--- a/lib/Controller/BoardController.php
+++ b/lib/Controller/BoardController.php
@@ -153,6 +153,6 @@ class BoardController extends ApiController {
 	 * @return \OCP\Deck\DB\Board
 	 */
 	public function clone($boardId) {
-		return $this->boardService->clone($boardId);
+		return $this->boardService->clone($boardId, $this->userId);
 	}
 }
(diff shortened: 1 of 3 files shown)


juliushaertl:
The backend actually should allow cloning with read permission:

https://github.com/nextcloud/deck/blob/52febb396cc79fd162efdd87550947f35d3e3f84/lib/Service/BoardService.php#L635

I'd say that is also what makes most sense here, so the question is rather why it does not work with that one already.

jakobroehrl:
The problem is, that the user is copied from the previous board: 
https://github.com/nextcloud/deck/blob/52febb396cc79fd162efdd87550947f35d3e3f84/lib/Service/BoardService.php#L640

@juliushaertl I have to transfer the userId here, right?
https://github.com/nextcloud/deck/blob/52febb396cc79fd162efdd87550947f35d3e3f84/lib/Service/BoardService.php#L630

juliushaertl:
> The problem is, that the user is copied from the previous board

Ah :see_no_evil: Yes, then fixing the owner to use the current user id should do the trick here when cloning.

Issue #1826, opened 2020-05-07
1.0.0: Calendar formatting does not allow to choose next month


jacotec:
**Describe the bug**
Simple. Tell me how to select to the next month ...

<img width="390" alt="grafik" src="https://user-images.githubusercontent.com/3321655/81315208-4d994b80-908a-11ea-8f4c-5ce06b04ee4d.png">

fm-sys:
I have the same Issue. Hope that will be fixed quickly.

Until then following workaround to come to the next month:
![2020-05-08 11_42_52-Präsentation1 - PowerPoint](https://user-images.githubusercontent.com/64581222/81393905-686acf00-9121-11ea-83dc-09920deee905.png)
Click at one of the days of next month and then go back to date select. That's it!

BKapelari:
Same issue here

juliushaertl:
@BKapelari Please use the [Github Reactions](https://github.com/blog/2119-add-reactions-to-pull-requests-issues-and-comments) feature instead of commenting without providing additional information.

d0-wN:
at the moment, the easiest workaround (since @fm-sys solution does not quite work for me) is to install the"custom css plugin" ([store-link](https://apps.nextcloud.com/apps/theming_customcss)) and add the following code:
`.mx-datepicker[data-v-e1f55ce] .mx-calendar-header, .mx-datepicker[data-v-e1f55ce] .mx-time-header{
          width: auto !important;
}`

the design will be a little bit messed up, but all buttons you need to pick a date/time will be visible then. (let me know if this works for you guys)

BKapelari:
> at the moment, the easiest workaround (since @fm-sys solution does not quite work for me) is to install the"custom css plugin" ([store-link](https://apps.nextcloud.com/apps/theming_customcss)) and add the following code:
> `.mx-datepicker[data-v-e1f55ce] .mx-calendar-header, .mx-datepicker[data-v-e1f55ce] .mx-time-header{ width: auto !important; }`
> 
> the design will be a little bit messed up, but all buttons you need to pick a date/time will be visible then. (let me know if this works for you guys)

Thank you for this solution. For me the following CSS-Code worked the best:
```
.mx-calendar-header {
    display: flex !important;
}
.mx-calendar-header-label {
    overflow: hidden;
}
.mx-btn.mx-btn-text.mx-btn-current-year {
  width: 3.2em;
  overflow: hidden;
}
```
[...]

PR #1868, opened 2020-05-12
try to change calender format

jakobroehrl:
* Resolves: #1826 
* Target version: master 

It doesn't work. 
Avoid mutating an injected value directly since the changes will be overwritten whenever the provided component re-renders. injection being mutated: "t"


juliushaertl:
I'll have a look as discussed.

juliushaertl:
closing in favor of #1945

PR #1870, opened 2020-05-12
fix modal imports

jakobroehrl:
* Resolves: #1822
* Target version: master

### DIFF
--- a/src/BoardSelector.vue
+++ b/src/BoardSelector.vue
@@ -80,7 +80,7 @@
 
 </style>
 <script>
-import { Modal } from '@nextcloud/vue/dist/Components/Modal'
+import Modal from '@nextcloud/vue/dist/Components/Modal'
 import axios from '@nextcloud/axios'
 
 export default {
--- a/src/CardSelector.vue
+++ b/src/CardSelector.vue
@@ -45,8 +45,8 @@
 </template>
 
 <script>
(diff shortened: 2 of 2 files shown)


juliushaertl:
So the partial imports here are actually wanted, because both the BoardSelector and CardSelector are used in the separate init-collections.js bundle. Since we don't use any other components there, it helps to keep the bundle size small.

src/BoardSelector.vue
```diff
@@ -80,7 +80,7 @@
 
 </style>
 <script>
-import { Modal } from '@nextcloud/vue/dist/Components/Modal'
```
juliushaertl, line 83: `import Modal from '@nextcloud/vue/dist/Components/Modal'`

src/CardSelector.vue
```diff
@@ -45,8 +45,7 @@
 </template>
 
 <script>
-import { Modal } from '@nextcloud/vue/dist/Components/Modal'
-import { Multiselect } from '@nextcloud/vue/dist/Components/Multiselect'
```
juliushaertl, line 49:
```
import Modal from '@nextcloud/vue/dist/Components/Modal'
import Multiselect from '@nextcloud/vue/dist/Components/Multiselect'
```

Should also work and keep the bundle size small

Issue #1862, opened 2020-05-11
Parameters icon's placement on cards is not ideal


Clementine46:
The parameters icon on cards is placed on the bottom-right corner.
According to me, this raises three issues : 

1) The placement is unnatural, parameters icons are usually found on top-right corners so this is unusual for the user.

2) Because of that placement, the bottom part of a card ends up being pretty big. There is a lot of empty space that seems unnecessary.

<img width="291" alt="Capture d’écran 2020-05-11 à 17 44 14" src="https://user-images.githubusercontent.com/50834537/81581651-4edeb700-93af-11ea-9c1a-914eacd9d1d5.png">

3) But most importantly, the card's parameters aren't accessible in compact mode, which is unpractical.

<img width="291" alt="Capture d’écran 2020-05-11 à 17 44 23" src="https://user-images.githubusercontent.com/50834537/81581545-2bb40780-93af-11ea-9afb-b9c24dd5e0b4.png">

By placing the parameters icon in the top right corner of a card, you can solve these three issues at once. 

![Deck_Parameters](https://user-images.githubusercontent.com/50834537/81581333-ded03100-93ae-11ea-8144-6a894307a220.png)

Here, you can see that there is no empty space anymore and the parameters are still accessible in compact mode.

What do you think ?

juliushaertl:
Yes, we should definitely fix that. I would actually assume that a bug :wink:

jakobroehrl:
Same topic as in here: https://github.com/nextcloud/deck/issues/799

Issue #799, opened 2018-12-16
The action menu of cards should be available in compact mode.


RobertZenz:
The compact mode is a very nice feature, but using the compact mode will hide the action menu. It would be nice if compact would would preserve the action menu and move it into the single line item.

Issue #1869, opened 2020-05-12
style all modals identical


jakobroehrl:
- [ ] 4 times "add to project"
- [ ] move card
- [ ] add attachment to desc

juliushaertl:
@jakobroehrl I'm not sure what this is exactly about, please be a bit more verbose about what an issue is about ;) Is it fixed already? :grin:

jakobroehrl:
It's open. The idea was to see all modals and do an identical styling.

juliushaertl:
OK, maybe we can collect some screenshots then so we can discuss the layout changes upfront.

joshtrichards:
I guess this was in response to this thread: https://github.com/nextcloud/deck/pull/1875#discussion_r424977973

Still a thing? :)

juliushaertl:
Let's close this, not entirely sure what this was about.

PR #1875, opened 2020-05-13
show menu in compact mode

jakobroehrl:
* Resolves: #1862 #799
* Target version: master 

compact:
grafik

not compact:
grafik

### DIFF
--- a/src/components/cards/CardItem.vue
+++ b/src/components/cards/CardItem.vue
@@ -54,6 +54,8 @@
 						</div>
 					</transition>
 				</div>
+
+				<CardMenu v-if="!editing && compactMode" :id="id" class="right" />
 			</div>
 			<transition-group name="zoom"
 				tag="ul"
@@ -78,10 +80,11 @@ import CardBadges from './CardBadges'
 import Color from '../../mixins/color'
 import labelStyle from '../../mixins/labelStyle'
 import AttachmentDragAndDrop from '../AttachmentDragAndDrop'
+import CardMenu from './CardMenu'
 
 export default {
 	name: 'CardItem',
-	components: { CardBadges, AttachmentDragAndDrop },
+	components: { CardBadges, AttachmentDragAndDrop, CardMenu },
 	directives: {
 		ClickOutside,
 	},
--- a/src/components/cards/CardMenu.vue
+++ b/src/components/cards/CardMenu.vue
@@ -1,0 +1,171 @@
+<!--
+  - @copyright Copyright (c) 2018 Julius Härtl <jus@bitgrid.net>
+  -
+  - @author Julius Härtl <jus@bitgrid.net>
+  -
+  - @license GNU AGPL version 3 or any later version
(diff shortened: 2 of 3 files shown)


src/components/cards/CardItem.vue
```diff
@@ -54,6 +54,11 @@
 						</div>
 					</transition>
 				</div>
+				<div v-if="!editing && compactMode" class="right">
+					<transition name="zoom">
```
juliushaertl, line 58: Let's get rid of the transition since we don't have that anywhere in nextcloud for a menu and it is not working anyways ;)

juliushaertl: Then you can also get rid of the wrapping div here.

src/components/cards/CardMenu.vue
```diff
@@ -0,0 +1,168 @@
...
+		},
+		moveCard() {
+			this.copiedCard = Object.assign({}, this.card)
+			this.copiedCard.stackId = this.selectedStack.id
+			this.$store.dispatch('moveCard', this.copiedCard)
+			this.modalShow = false
+		},
+		async loadStacksFromBoard(board) {
+			try {
+				console.debug(board)
```
juliushaertl, line 140: Let's get rid of that since we are on it ;)

src/components/cards/CardMenu.vue
```diff
@@ -0,0 +1,168 @@
...
+			} catch (err) {
+				return err
+			}
+		},
+	},
+}
+</script>
+
+<style lang="scss" scoped>
+	.modal__content {
```
juliushaertl, line 153: I think we use a styling like that in quite some places. Ideally we can move it to a dedicated css file and import it.

jakobroehrl: I'll do this in here: https://github.com/nextcloud/deck/issues/1869
OK?

juliushaertl: Sure :+1:

jakobroehrl: Then, we can merge? 😬

reox:
I'm not sure if this violates any design guides or would be too weird, but would it not make sense to show the three dots vertically in compact mode instead of horizontal?
Just thinking, that it would save a few pixels ;)

juliushaertl:
> I'm not sure if this violates any design guides or would be too weird, but would it not make sense to show the three dots vertically in compact mode instead of horizontal?
> Just thinking, that it would save a few pixels ;)

the button still needs a click target of 44x44px so we won't actually save something :wink: I also prefer to have the same icon for the menu everywhere.

juliushaertl:
Rebased conflicts

reox:
> the button still needs a click target of 44x44px so we won't actually save something wink I also prefer to have the same icon for the menu everywhere.

I see, that makes sense. It was more loud thinking ;)
But very good to have the menu in the compact view though!

Issue #1712, opened 2020-04-17
Detect iOS/Android browser and show a top "banner" to download respective apps


jakobroehrl:

juliushaertl:
:+1: For now of course just for android ;)

PR #1879, opened 2020-05-13
banner to download android app

jakobroehrl:
* Resolves: #1712
* Target version: master 

- [ ] Can't test if this works on an android browser. 
- [ ] Any ideas how to style it nice? 

grafik

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -24,6 +24,9 @@
 	<div id="content" :class="{ 'nav-hidden': !navShown, 'sidebar-hidden': !sidebarRouterView }">
 		<AppNavigation />
 		<div id="app-content">
+			<div v-if="isAndroid">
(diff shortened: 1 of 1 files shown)


stefan-niedermann:
Lol, your text :laughing:  "There is an app. Consider downloading it, stupid!".

I think the best would be to place a well-known badge somewhere with a nice little close-icon (❎) next to it, so it won't bother the user again.

grafik

One can often see something like this:

grafik

It looks a bit better integrated.

juliushaertl:
Ideally we should the app in the manifest file as described in https://developers.google.com/web/fundamentals/app-install-banners/native 

But I have no idea if we can properly expose that as a Nextcloud app.

mokkin:
And if there ist the Google badge there should be also the F-Droid one :)

Issue #1880, opened 2020-05-13
UI/UX: Hovering should highlight item


simonspa:
Hovering an entry on the board list as well as a card on a board should highlight the item by changing the background color. This helps to spot the item about to be selected.

On Nextcloud files, hovering an entry on the files list looks like this:

![image](https://user-images.githubusercontent.com/1677436/81842298-12948d80-954c-11ea-842e-28520b99d9ab.png)

while no highlight is added to the list item on Deck:

![image](https://user-images.githubusercontent.com/1677436/81842285-0d374300-954c-11ea-8703-015c578e6d4b.png)

(I'm sorry the mouse pointer is not visible in screenshots :))

Optimally also cards could be highlighted when hovering them - the interaction would feel a lot more fluid.

stefan-niedermann:
I agree with the hover effect on the boards list, but i think the background color of a card should stay the same when one is hovering it. I think we can improve the haptic on other points: for example blur the shadow around a dragged card:

![grafik](https://user-images.githubusercontent.com/4741199/81847321-b9305c80-9553-11ea-969b-b7e49b1b4e4c.png)

Instead of changing the color, we should signal the user, that (s)he can actually grab the card by using `cursor: grab;` / `cursor: grabbing;`.

![grafik](https://user-images.githubusercontent.com/4741199/81847513-fc8acb00-9553-11ea-9bb8-39a074ace3ce.png)

simonspa:
I fully agree - coloring the whole card is too much area and will just flicker and distract. Also, the buttons themselves would not show properly. I like your suggestion with the shadow, that looks quite good.

Concerning the cursor: do you think using `grab` over cards would confuse users and distract from the possibility of clicking it for opening the sidebar?

stefan-niedermann:
We can also think about using `cursor: pointer` - i think both variants are better than the usual `cursor: default` because imply that there is some kind of action possible and this is not a readonly view.

Personally i think, that the "click" action is usually discoverable and one of the first things that users will try, so `cursor: grab` should be fine.

Though when the card is actually *dragged*, the cursor `grabbing` should definitely be used.

cc @nextcloud/designers

juliushaertl:
@simonspa Using it for some time now I actually think we should switch the gray shades of the hover and current card state so that the current card is highlighted stronger than the hover focus. Hover should rather be a decent indicator imo. What do you think?

juliushaertl:
[...]

PR #1882, opened 2020-05-13
Add hover effect to board list

simonspa:
This MR adds a mouse-hover effect to the board list, just as seen in Nextcloud files:

image

This is part of #1880


juliushaertl:
Nice one, thanks a lot :+1:

Issue #1804, opened 2020-05-06
Indicate filter is in use


Dubidubiduu:
**Is your feature request related to a problem? Please describe.**
Maybe I forget I use a filter. I don't get a hint it is in use.

**Describe the solution you'd like**
Maybe add an alarming color-background circle to the filter icon, when it is in use.

jakobroehrl:
There is a change it the opacity when a filter was set:
whithout filter:
![grafik](https://user-images.githubusercontent.com/24530680/81545765-a19d7c00-9379-11ea-8bd1-6184aba1f660.png)

with filter:
![grafik](https://user-images.githubusercontent.com/24530680/81545809-b8dc6980-9379-11ea-87fd-00310ea1ced4.png)

@nextcloud/designers Do you have a hint for me? I like it for exampe how it is done in the android deck app: 
![Screenshot_20200511-112449](https://user-images.githubusercontent.com/24530680/81546094-25efff00-937a-11ea-8e27-8eb7016154cc.jpg)

Dubidubiduu:
Desktop:
I don't know if it is accessible, but I question it's obvious.

If I already know this filter could be activated, yeah, it's a helpful fast check. So imo this would work for heavy users. But if I just wonder why I don't see all cards this wouldn't come into my focus.

juliushaertl:
We could also show some additional context to make it clear that the current view is filtered (if space allows) like trello shows the number of results:
![image](https://user-images.githubusercontent.com/3404133/81567521-b3444b00-939c-11ea-94ae-1b69c424364f.png)

PR #1887, opened 2020-05-14
new filter icon

jakobroehrl:
* Resolves: #1804
* Target version: master 

not set:
grafik

set:
grafik

### DIFF
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -54,8 +54,9 @@
 			</div>
 			<div class="board-action-buttons">
 				<Popover>
-					<Actions slot="trigger" :style="filterOpacity" :title="t('deck', 'Apply filter')">
-						<ActionButton icon="icon-filter" />
+					<Actions slot="trigger" :title="t('deck', 'Apply filter')">
+						<ActionButton v-if="isFilterActive" icon="icon-filter_set" />
+						<ActionButton v-else icon="icon-filter" />
 					</Actions>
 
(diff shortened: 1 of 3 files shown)


src/components/Controls.vue
```diff
@@ -54,8 +54,9 @@
 			</div>
 			<div class="board-action-buttons">
 				<Popover>
-					<Actions slot="trigger" :style="filterOpacity" :title="t('deck', 'Apply filter')">
-						<ActionButton icon="icon-filter" />
+					<Actions slot="trigger" :title="t('deck', 'Apply filter')">
+						<ActionButton v-if="isFilterActive" icon="icon-filter_set" />
```
stefan-niedermann, line 58: I don't know `vue` that good, but won't this call and execute the method on each single change detection?

Wouldn't it be smarter in this case to use a boolean property to check against and set actively change that boolean property when a user adds tags, users or due to the list of filters (or, analog, removes them)?

juliushaertl: isFilterActive is a computed property so the result is cached. [This line](https://github.com/nextcloud/deck/pull/1887/files#diff-3c5f6c58f8c054bd6d60375762579633L222) is basically evaulated every time the filter is updated but the v-if condition is only triggered if the actual result of `isFilterActive` is changing.

stefan-niedermann: Well then: forget it and thanks for the explanation :)

PR #1888, opened 2020-05-14
Improve hovering response in board

simonspa:
Add "grab" cursor for stack titles.

I wanted to add the same for cards and a "grabbing" while moving, but I'm still fighting with scoped CSS.

Ref #1880

### DIFF
--- a/src/components/board/Stack.vue
+++ b/src/components/board/Stack.vue
@@ -237,7 +237,7 @@ export default {
 		margin: 3px -3px;
 		margin-right: -10px;
 		margin-top: 0;
-		margin-bottom: 0;
+		margin-bottom: 3px;
 		background-color: var(--color-main-background-translucent);
 		cursor: grab;
 
--- a/src/components/board/Stack.vue
+++ b/src/components/board/Stack.vue
@@ -239,10 +239,12 @@ export default {
 		margin-top: 0;
 		margin-bottom: 0;
 		background-color: var(--color-main-background-translucent);
+		cursor: grab;
 
 		h3, form {
 			flex-grow: 1;
 			display: flex;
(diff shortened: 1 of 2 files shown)


simonspa:
Now also added slightly stronger drop shadow when hovering a card:

image

simonspa:
@jancborchardt not sure if there is a unified way in NC design for this - but to me it looks unobstrusive but still improves user interaction.

stefan-niedermann:
The border looks too dark. I wouldn't increase it but spread it wider instead.

simonspa:
I played with that too, but I liked the darker border better. The problem with increasing the spread is that the rounding of the corners becomes very different from the card's corners - which then "pierce" through the shadow:

image


Also, in order to be noticeable enough the spread needs to be around 10px - which partially overlaps with adjacent elements already.

juliushaertl:
I agree with @stefan-niedermann that the dark shadow looks a bit to strong, especially for a hover feedback. 

cc @Clementine46 @Nyco :wink: Maybe you have some input on that as well

simonspa:
Okay, third option would be to combine a bit more `blur` with a smaller dark border as a way inbetween?

image

juliushaertl:
Looks better to me :+1: 

For the overlapping issue with the header background I think it would be fine to decrease the height of the list header a bit and increase the spacing between header and first card so that the shadow is not cut anymore.

simonspa:
Alright, did that - just increased the margin from 0 to 3px. Now looks like this:

image

juliushaertl:
Let's get this in :+1:

juliushaertl:
Frontend tests happy and no backend changes ... merging

Issue #1848, opened 2020-05-11
Pre-upgrade card won't show card details


ampoz:
**Describe the bug**
Recently upgraded to Nextcloud 18 + Deck 1.0.0. Since then, the deck boards have become unusable as we cannot see a card's details view on the right side of the page. When clicking on a card, nothing happens but the browsers' console log shows:

"uncaught exception: Object"

and

"Error: "Input data should be a String"
    parse index.js:518
    render index.js:543
    renderedDescription CardSidebar.vue:328
    VueJS 3
    Ur CardSidebar.vue:1
    VueJS 7
vue.esm.js:1897:12
"

This is the case for most of the cards. Every once in a while, a card will open the details side view; these are always the same cards -- previously not working card still don't work or show card details after clicking on a card that "works properly".

I've also noticed when creating a new board and adding cards, these all work fine; it seems to only be caused by pre-1.0.0 cards (most of them, but not all). I've also noticed that all the cards that do work seem to have a description, where as card that don't work seem to have an empty description (checked this by using the Android app as we cannot currently see card details for those cards on the web).

**To Reproduce**
1. Go to a Deck board
2. Click on a card
3. Side view doesn't show

**Expected behavior**
Clicking on a card show open the card details right-side view so the card can be edited.

**Client details:**
Tried on:
 - Linux and Windows
 - Chrome and Firefox

<details>
<summary>Server details</summary>

**Nextcloud version:** 18.0.4

```
Login as admin user into your Nextcloud and access
http://example.com/index.php/settings/integrity/failed
paste the results here.

No errors have been found.
```

#### Browser log
```
Insert your browser log here, this could for example include:

"uncaught exception: Object"

and

"Error: "Input data should be a String"
    parse index.js:518
    render index.js:543
    renderedDescription CardSidebar.vue:328
    VueJS 3
    Ur CardSidebar.vue:1
    VueJS 7
vue.esm.js:1897:12
"
```

</details>

putt1ck:
OK, can reproduce but not sure it only depends on the card creation point/content. However impacted cards do seem to be pre-upgrade and have no description, but there are also cards which match those criteria and do work.

The `uncaught exception: Object` entry in the console occurs for click on any card. While the error

`Error: "Input data should be a String"
    parse index.js:518
    render index.js:543
    renderedDescription CardSidebar.vue:328
    VueJS 3
    Ur CardSidebar.vue:1
    VueJS 7
vue.esm.js:1897:12`

[...]

PR #1897, opened 2020-05-15
Fix issue if card description might be null

juliushaertl:
Fixes https://github.com/nextcloud/deck/issues/1848


juliushaertl:
I've pushed two more commits to make error reporting more sane in firefox and fix the duplicate navigation error as well.

Issue #1894, opened 2020-05-15
Stay in compact mode


reox:
**Is your feature request related to a problem? Please describe.**

When enabling the compact mode, it is only enabled until the page is reloaded.
I would rather like to enable the expanded view than enabling the compact mode all the time

**Describe the solution you'd like**

After the compact mode is enabled, it should stay on unless explicitly the expanded mode is enabled.

Deck v1.0.0

juliushaertl:
Yes, i think that was lost during the vue rewrite. Iirc we stored the state of compact mode in the browsers local storage in the past.

reox:
Ah okay, I wasn't sure if it worked before :sweat_smile: Because I was switching between too many workstations, I probably never noticed that it stayed for one browser.

juliushaertl:
Probably worth to persist that setting on the server as well in a second step, right :+1:

PR #1918, opened 2020-05-18
save compact mode on localstorage

jakobroehrl:

* Resolves: #1894 
* Target version: master

### DIFF
--- a/src/store/main.js
+++ b/src/store/main.js
@@ -55,7 +55,7 @@ export default new Vuex.Store({
 	state: {
 		showArchived: false,
 		navShown: true,
-		compactMode: false,
+		compactMode: localStorage.getItem('deck.compactMode'),
 		sidebarShown: false,
 		currentBoard: null,
 		currentCard: null,
@@ -176,6 +176,7 @@ export default new Vuex.Store({
 		},
 		toggleCompactMode(state) {
 			state.compactMode = !state.compactMode
+			localStorage.setItem('deck.compactMode', state.compactMode)
 		},
 		setBoards(state, boards) {
 			state.boards = boards


src/components/Controls.vue
```diff
@@ -204,6 +204,12 @@ export default {
 		}
 	},
 
+	created: function() {
+		if (localStorage.getItem('nc_deck_compact_active') === 'true') {
```
juliushaertl, line 208: Let's use `deck.compactMode` as key here since that was also used in 0.8

```
js/controller/BoardController.js
186:		localStorage.setItem('deck.compactMode', JSON.stringify($rootScope.compactMode));
```

src/components/Controls.vue
```diff
@@ -253,6 +259,7 @@ export default {
 		},
 		toggleCompactMode() {
 			this.$store.dispatch('toggleCompactMode')
+			localStorage.setItem('nc_deck_compact_active', this.compactMode)
```
juliushaertl, line 262: Please move the localStorage handling to the store where toggleCompactMode is actually taking place.

The getItem can then also be fetched as initial value for the store data.

jakobroehrl: How to add an created function in the main.js store to fetch the store data?

juliushaertl: You can't and don't need one. You can just set it in the initial state of the store.

jakobroehrl: How to do this? This is not working (main.js)
state: {
		showArchived: false,
		navShown: true,
		**compactMode: localStorage.getItem('deck.compactMode'),**

juliushaertl: You don't set it in your current pull request, once it is set that works like a charm here.

jakobroehrl: Changed it, it's not working for me. 
On every page reload it's compact

Issue #1807, opened 2020-05-06
Add to filter: unassigned cards


Dubidubiduu:
**Is your feature request related to a problem? Please describe.**
I guess most of the time I want to filter for cards assigned to me, and unassigned cards (that overdue or tagged with xyz)

**Describe the solution you'd like**
Add an option to filter for unassigned cards. Should be the first checkbox in assigned users part of the filter.

PR #1926, opened 2020-05-19
new filter: unassigned cards

jakobroehrl:

* Resolves: #1807
* Target version: master

### DIFF
--- a/src/store/card.js
+++ b/src/store/card.js
@@ -35,7 +35,6 @@ export default {
 				const { tags, users, due, unassigned } = rootState.filter
 				let allTagsMatch = true
 				let allUsersMatch = true
-				let allUnassigned = true
 
 				if (tags.length > 0) {
 					tags.forEach((tag) => {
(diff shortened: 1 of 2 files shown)


src/store/card.js
```diff
@@ -58,6 +58,10 @@ export default {
 					}
 				}
 
+				if (unassigned) {
+					return card.assignedUsers.length === 0
```
juliushaertl, line 62: This will not work when both unassigned and a due filter are set since it returns to early

src/store/card.js
```diff
@@ -58,6 +59,16 @@ export default {
 					}
 				}
 
+				if (unassigned) {
+					if (card.assignedUsers.length > 0) {
+						allUnassigned = false
+					}
+
+					if (!allUnassigned) {
```
juliushaertl, line 67: Why not just returning false directly then instead of adding the `allUnassigned` variable? This code looks way more complicated than it needs to be :wink:

jakobroehrl:
It's working, ready to merge

Issue #955, opened 2019-03-21
Impossible to unsubscribe from this board


lvarnava:
Nextcloud 14.0.8.2
ubuntu 18.04.1
postgresql 10.6
apache 2.4.29
php 7.2.15
Deck 0.5.2

Suppose someone shared a board with me.
I can not unsubscribe from this board.

juliushaertl:
From #1929 

>> How to solve this in the backend? new function unshareBoard in BoardService.php? Or is there a better way? Thanks!

> This depends a bit on how the board was shared. If it is a user share we could just remove the ACL rule. This would basically require a check so that the user with no manage permission is still allowed to remove their own acl assignemnt. 

> For groups/circles this gets more complicated as you just want to exclude the share for the users who decided to opt-out. I guess a nice approach would be to extend the acl table with an active field that is true by default but can be set to false on unshare. Then we could insert an additional user acl for the unshare with active=false and make sure to check that in addition to group acls.

merli-hsa:
I faced the same problem that someone shared a board with me and does not care removing it.

Another idea for solving that without touching ACL rules:
Creating folders and sorting boards into these folders.

Results:
- One could create a folder "Archive", "Trash", "Not Interesting Anymore" and move the annoying board(s) to this folder.
- Structuring boards might be also have further advantages besides this, e.g. being able to separate boards from one team/group/project from others

juliushaertl:
But that seems like a totally different feature that could only be used to workaround that we don't have any unshare self-service function :wink: Let's keep that separate in #2006

loelkes:
Please correct me if I'm wrong: as of now only the owner of the board can remove people from the list? Unlike other apps and services users can't unshare themselves? I feel like I'm missing something. Is someone working on this?

putt1ck:
Although if I'm the project manager related to that board I'm going to be quite annoyed if you decide to no longer get notifications and add you right back in. OTOH if you were added by accident or for a task that is complete, you can just ask to be removed.

NB on pay-for Trello, when you have teams I don't think you get to remove yourself from a team board.

loelkes:
> OTOH if you were added by accident or for a task that is complete, you can just ask to be removed.

There are situations where you can't talk to these people anymore. It then has to involve an Admin using the 'impersonate' feature which can be problematic. It's much easier for everyone if you can just remove yourself the same way you can remove yourself from file shares, chats or calendars.

[...]

PR #1929, opened 2020-05-19
unshare board

jakobroehrl:

* Resolves: #955
* Target version: master 

@juliushaertl 
How to solve this in the backend? new function unshareBoard in BoardService.php? Or is there a better way? Thanks!

### DIFF
--- a/src/components/navigation/AppNavigationBoard.vue
+++ b/src/components/navigation/AppNavigationBoard.vue
@@ -236,6 +236,18 @@ export default {
 					text: t('deck', 'Board details'),
 				})
 
+				if (this.board.shared) {
+					actions.push({
+						action: () => {
+							this.hideMenu()
+							this.loading = true
+							this.$store.dispatch('unshareBoard', this.board)
+						},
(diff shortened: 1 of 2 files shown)


juliushaertl:
> How to solve this in the backend? new function unshareBoard in BoardService.php? Or is there a better way? Thanks!

This depends a bit on how the board was shared. If it is a user share we could just remove the ACL rule. This would basically require a check so that the user with no manage permission is still allowed to remove their own acl assignemnt. 

For groups/circles this gets more complicated as you just want to exclude the share for the users who decided to opt-out. I guess a nice approach would be to extend the acl table with an active field that is true by default but can be set to false on unshare. Then we could insert an additional user acl for the unshare with active=false and make sure to check that in addition to group acls.

jakobroehrl:
@juliushaertl Should we close this PR, it's more work than I thought...

juliushaertl:
Sure, added my backend comment to https://github.com/nextcloud/deck/issues/955 in case someone picks it up again.

Issue #1721, opened 2020-04-19
Archived card listing


juliushaertl:
The current way of listing archived cards is not optimal as we just show the archived cards in the same way as regular cards, grouped by their stack. In a kanban-like organization process cards most likely end up on a "Done" list or something similar where they are then archived from later on. 

This results in the archived view being a long list of cards in the Done stack and is not really helpful. 

My proposal would be to move displaying of archived cards to the board sidebar similar to how it is done in github projects and trello. 

![image](https://user-images.githubusercontent.com/3404133/79687626-ec2f4b00-8248-11ea-9869-8772e28c790c.png)

This way archived cards would be a separate list just for the archiving purpose and not interfere with the column layout that is just ment for managing currently active cards.

cc @jancborchardt @nextcloud/designers @nextcloud/deck

stefan-niedermann:
I really really like this approach. We will definitely go for this in the [Android app](https://github.com/stefan-niedermann/nextcloud-deck) because separating this into a completely independent view is way easier to implement and maintain.

juliushaertl:
So we discussed this recently in our deck design review and came to the conclusion that it might make sense to just have a toggleable  additional list in the regular board view for all the archived cards.

tomasmark79:
It's super. I would be interested in one more thing. I ran a full text search and everything works absolutely great.

However, as I will use Deck for individual projects and want to return to them, I would like elasticsearch to search archived cards as well. And that's not happening now. Is there a way to turn it on, please?

Thank you

dodedodo:
> So we discussed this recently in our deck design review and came to the conclusion that it might make sense to just have a toggleable additional list in the regular board view for all the archived cards.

This may make it a bit more confusing as to what that list actually is. If it looks like a regular list, why not make a list yourself and call it 'archive'? Why can you toggle the archive list, but not other lists? 

Maybe you could consider giving the archive list a greyed-out look to denote that it's a special function.

PR #1930, opened 2020-05-19
new archive sidebar

jakobroehrl:
* Resolves: #1721
* Target version: master 

Cards are listed in the sidebar now:
grafik

### DIFF
--- a/src/components/board/ArchivedTabSidebar.vue
+++ b/src/components/board/ArchivedTabSidebar.vue
@@ -1,0 +1,66 @@
+<template>
+	<div>
+		<ul>
+			<li v-for="archivedCard in archivedCards" :key="archivedCard.id">
+				<CardItem :id="archivedCard.id" />
+			</li>
+		</ul>
+	</div>
+</template>
+
+<script>
+import { mapGetters } from 'vuex'
+import relativeDate from '../../mixins/relativeDate'
+import CardItem from '../cards/CardItem'
+
+export default {
+	name: 'ArchivedTabSidebar',
+	components: {
+		CardItem,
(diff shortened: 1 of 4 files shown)


jakobroehrl:
@juliushaertl 
Do you have any idea why this is not working?
https://github.com/nextcloud/deck/blob/92e1d2798fedf16837c5ce9cd3dccf973fe37839/src/store/card.js#L93-L95 I don't get any cards back, thanks!

juliushaertl:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find Have a look on what find does :wink:

jakobroehrl:
@juliushaertl 
The Sidebar only shows card if I click on show archived cards in the top bar, why?

juliushaertl:
>  The Sidebar only shows card if I click on show archived cards in the top bar, why?

Are they fetched already and in the store?

juliushaertl:
I'm not sure about this from the UI perspective as it would mean that looking at a card would cause a loss of context since the sidebar switches to the card view then.

jakobroehrl:
> I'm not sure about this from the UI perspective as it would mean that looking at a card would cause a loss of context since the sidebar switches to the card view then.

OK, what should I do? Implement it like this?
grafik

juliushaertl:
> OK, what should I do? Implement it like this? 

Not sure what you mean with the screenshot. This won't help as then the board sidebar still switches to the card sidebar and you no longer see the context of archived cards.

jakobroehrl:
We want to use a model for the card content, I will test it...

Issue #805, opened 2018-12-18
Open/Assigned Tasks in all board [$100]


mixvoip:
Hi,

we are using several boards in Deck right now. To avoid manual copying and to avoid missed tasks, we want to request some sort of filter where all tasks for a specific user across all board can be found. For Follow-Up etc it should not be only "my tasks".
As we are requesting to have this for all boards in one view, we think it is not a duplicate.

Best regards
Clemens

mixvoip:
Bounty of 75$ linked: https://www.bountysource.com/issues/67677436-open-assigned-tasks-in-all-board

juliushaertl:
@mixvoip Thanks for your feature request and for adding a bounty on it. Can you describe a bit more in detail what you mean by this:

> For Follow-Up etc it should not be only "my tasks".

mixvoip:
@juliushaertl I just wanted to make clear that we are not requesting a feature to see only all own tasks in all boards at one place. We need to be able to list all tasks across all boards for all existing users.
So to make it somehow visible:
Imagine you have 3 boards (X,Y,Z) with different tasks for 3 Users (A,B,C).
We want to have a view where we can see all tasks from A or from B or from C (via a filter or something like that). There should be no filitering like "A can see only tasks from A".

Is that now clearer?

BKapelari:
I would really like to see something like the "today" and "this week" view in Wunderlist: [(LINK)](https://www.wunderlist.com/blog/wunderlist-for-osx-yosemite/01@2x.png)

So showing all the tasks asigned to me for today/this week: [LINK](https://www.google.com/search?client=firefox-b-d&biw=1405&bih=815&tbm=isch&sa=1&ei=WKl6XPiXDumZjLsPmum72Ag&q=wunderlist+week&oq=wunderlist+week&gs_l=img.3...362436.363136..363281...0.0..0.98.470.5......1....1..gws-wiz-img.......0j0i30j0i24.lgnGkBEYmKo#imgrc=PoMoUGA2WwMs5M:)

If I click on the task, there could be a link to the board who is containing it.
This would be a simple solution for not missing tasks. What do you think about it?

Hopefully I can help in develop this. I'm new in Nextcloud extension programming

putt1ck:
> @juliushaertl I just wanted to make clear that we are not requesting a feature to see only all own tasks in all boards at one place. We need to be able to list all tasks across all boards for all existing users.
> So to make it somehow visible:
[...]

Issue #1878, opened 2020-05-13
change group by [discussion]


jakobroehrl:
- stack (default)
 - tags
 - assigned to
 - due

juliushaertl:
@jakobroehrl Please be a bit more verbose about what this is about. I don't get it from the title/description.

jakobroehrl:
> 
> @jakobroehrl Please be a bit more verbose about what this is about. I don't get it from the title/description.

Sorry, better now?
What do you think?

juliushaertl:
I assume a card that has multiple labels would also show up multiple times then? How would drag-and drop behave with multiple assigned labels/users then?

PR #1934, opened 2020-05-20
card dashboard

jakobroehrl:
* Resolves: #805 and touches #1878
* Target version: master 
* New impl of https://github.com/nextcloud/deck/pull/1319


juliushaertl:
> Wouldn't it be better to return a special kind of board from the backend?

Not sure what you mean with that.

lib/Db/CardMapper.php
```diff
@@ -141,6 +141,11 @@ public function findAllByStack($stackId, $limit = null, $offset = null) {
 		return $this->findEntities($sql, [$stackId], $limit, $offset);
 	}
 
+	public function findAllWithDue() {
+		$sql = 'SELECT * from `*PREFIX*deck_cards` WHERE duedate IS NOT NULL AND NOT archived AND deleted_at = 0';
```
juliushaertl, line 145: :warning: This will return also cards of other users.

src/store/main.js
```diff
@@ -101,6 +101,9 @@ export default new Vuex.Store({
 				return board.shared && !board.deletedAt
 			})
 		},
+		dashboards: state => {
```
juliushaertl, line 104: Please create an extra store module

src/services/CardApi.js
```diff
@@ -195,4 +195,19 @@ export class CardApi {
 			})
 	}
 
+	findAllWithDue(data) {
+		return axios.get(this.url(`/cards/findAllWithDue`))
+			.then(
+				(response) => {
+					return Promise.resolve(response.data)
+				},
```
juliushaertl, line 203:
```suggestion
				(response) => Promise.resolve(response.data),
```

Same for the other arrow functions :wink:

lib/Db/CardMapper.php
juliushaertl: So we either need a larger query with joining the board and acl tables together to get only the cards which are part of the users boards or we need to iterate over all users boards an fetch the cards per board.

juliushaertl:
As discussed, I don't see an easy way around fetching the users boards first as done [here](https://github.com/nextcloud/deck/blob/52febb396cc79fd162efdd87550947f35d3e3f84/lib/Service/BoardService.php#L113-L116) and then fetch the list of cards for the dashboards for each board.

src/components/navigation/AppNavigation.vue
```diff
@@ -23,6 +23,12 @@
 <template>
 	<div id="app-navigation" :class="{'icon-loading': loading}">
 		<ul id="deck-navigation">
+			<AppNavigationBoardCategory
+				id="deck-navigation-dashboard"
+				:text="t('deck', 'Card Dashboard')"
```
juliushaertl, line 28: How about 'Your cards'. I'd rather avoid the dashboard term in the user interface in regards to https://github.com/nextcloud/server/pull/21346

lib/Service/DashboardService.php
```diff
@@ -0,0 +1,150 @@
...
+	
+	/**
+	 * @return array
+	 */
+	public function findAllWithDue($userId) {
+		$userInfo = $this->getBoardPrerequisites();
+		$userBoards = $this->findAllBoardsFromUser($userInfo);
+		$allDueCards = [];
+		foreach ($userBoards as $userBoard) {
+			$allDueCards[] = $this->cardMapper->findAllWithDue($userBoard->getId());
```
juliushaertl, line 121:
```suggestion
			$allDueCards = array_merge($allDueCards, $this->cardMapper->findAllWithDue($userBoard->getId()));
```

lib/Db/CardMapper.php
```diff
@@ -141,6 +141,23 @@ public function findAllByStack($stackId, $limit = null, $offset = null) {
...
 
+	public function findAllWithDue($boardId) {
+		$sql = 'SELECT c.* FROM `*PREFIX*deck_cards` c
+	  	INNER JOIN `*PREFIX*deck_stacks` s ON s.id = c.stack_id
+		INNER JOIN `*PREFIX*deck_boards` b ON b.id = s.board_id
+		  WHERE `s`.`board_id` = ? AND duedate IS NOT NULL AND NOT c.archived AND c.deleted_at = 0 AND s.deleted_at = 0 AND NOT b.archived AND b.deleted_at = 0';
+		return $this->findEntities($sql, [$boardId]);
+	}
+
+	public function findMyAssignedCards($boardId, $username) {
```
juliushaertl, line 152: if it is find**My**AssignedCards it should not take a username but just pick up the current one, otherwise please rename to something like findAssignedCardsForUser

appinfo/routes.php
```diff
@@ -65,7 +65,12 @@
 		['name' => 'card#removeLabel', 'url' => '/cards/{cardId}/label/{labelId}', 'verb' => 'DELETE'],
 		['name' => 'card#assignUser', 'url' => '/cards/{cardId}/assign', 'verb' => 'POST'],
 		['name' => 'card#unassignUser', 'url' => '/cards/{cardId}/unassign', 'verb' => 'PUT'],
+		
+		// dashboard
+		['name' => 'card#findAllWithDue', 'url' => '/dashboard/due', 'verb' => 'GET'],
```
juliushaertl, line 70: Please move this to an ocs route and use an OCSController. That way we don't need separate api routes and external apps can use the same ones that the web UI uses.

src/components/dashboards/Dashboards.vue
```diff
@@ -98,7 +98,7 @@
 			</div>
 
 			<div class="dashboard-column">
-				<h2>this week</h2>
```
juliushaertl: Please always make strings translatable

appinfo/routes.php
```diff
@@ -110,6 +111,8 @@
 		['name' => 'card_api#reorder', 'url' => '/api/v1.0/boards/{boardId}/stacks/{stackId}/cards/{cardId}/reorder', 'verb' => 'PUT'],
 		['name' => 'card_api#delete', 'url' => '/api/v1.0/boards/{boardId}/stacks/{stackId}/cards/{cardId}', 'verb' => 'DELETE'],
 
+		['name' => 'card_api#findAllWithDue', 'url' => '/api/v1.0/dashboard/due', 'verb' => 'GET'],
```
juliushaertl, line 114: No need for this anymore with the OCS routes

lib/Controller/CardController.php
```diff
@@ -33,10 +34,11 @@ class CardController extends Controller {
 	private $cardService;
 	private $assignmentService;
 
-	public function __construct($appName, IRequest $request, CardService $cardService, AssignmentService $assignmentService, $userId) {
+	public function __construct($appName, IRequest $request, CardService $cardService, DashboardService $dashboardService, AssignmentService $assignmentService, $userId) {
```
juliushaertl, line 37: This file changes are no longer needed.

lib/Service/DashboardService.php
```diff
@@ -0,0 +1,187 @@
...
+		AssignedUsersMapper $assignedUsersMapper,
+		IUserManager $userManager,
+		IGroupManager $groupManager,
+		ActivityManager $activityManager,
+		EventDispatcherInterface $eventDispatcher,
+		ChangeHelper $changeHelper,
+		$userId
+	) {
+		$this->boardMapper = $boardMapper;
+		$this->stackMapper = $stackMapper;
```
juliushaertl, line 87: Please cleanup unused properties

src/components/cards/CardMenu.vue
```diff
@@ -80,9 +80,13 @@ export default {
 	name: 'CardMenu',
 	components: { Actions, ActionButton, Modal, Multiselect },
 	props: {
-		id: {
+		/* id: {
```
juliushaertl, line 83: :broom:

src/components/cards/CardMenu.vue
```diff
@@ -103,9 +107,9 @@ export default {
 			currentBoard: state => state.currentBoard,
 		}),
 
-		card() {
+		/* card() {
```
juliushaertl, line 110: :broom:

juliushaertl: There are also still usages of `this.id` in this file which is no longer available

src/components/dashboards/Dashboards.vue
```diff
@@ -0,0 +1,202 @@
...
+				nodue: [],
+				overdue: [],
+				today: [],
+				tomorrow: [],
+				thisWeek: [],
+				later: [],
+			}
+			dataset.forEach(card => {
+
+				if (card.duedate === null) {
```
juliushaertl, line 156: This should be done in getters in the dashboard store

src/services/CardApi.js
```diff
@@ -195,4 +195,24 @@ export class CardApi {
 			})
 	}
 
+	findAllWithDue(data) {
```
juliushaertl, line 198: I think this is no longer needed it is in DashboardApi

src/store/dashboard.js
```diff
@@ -0,0 +1,63 @@
...
+			state.withDue = withDue
+		},
+		setAssignedCards(state, assignedCards) {
+			state.assignedCards = assignedCards
+		},
+
+	},
+	actions: {
+		async loadDashboards({ commit }) {
+			const withDue = await apiClient.findAllWithDue()
```
juliushaertl, line 53: One action per dashboard please, otherwise we do to unneeded requests here.

juliushaertl (approved):
Let's get this in as a first part for the my assigned cards overview and iterate over it in follow ups.

Issue #1306, opened 2019-10-25
New Owner for a board


Dubidubiduu:
**Is your feature request related to a problem? Please describe.**
I accidentally created a board with a wrong user, now it is full of decks.

**Describe the solution you'd like**
I want to give the ownership to another user.

**Describe alternatives you've considered**
Copying deck to another user, would be needed that this is an exact copy with history, discussion and all.

**Additional context**
I want to have everything under one user which is used for administration purposes. I think this way I don't get any problems if people pass of my team.

sgdox:
Hello, we need this enhancement, because when a user left the organization, or change their functions, its impossible to change the owner.

Dubidubiduu:
@juliushaertl as you asked in #119, if there is a use case in also copy cards in a run with copying boards, yeah, so here is one use case.

Just wondering: Do you think, as for me it's just two boards, it's enough to change owner in "oc_deck_boards"-table in the database to another owner? Will this break anything?

Then I could finally delete the old users...

juliushaertl:
> Just wondering: Do you think, as for me it's just two boards, it's enough to change owner in "oc_deck_boards"-table in the database to another owner? Will this break anything?

That should work fine. 

Maybe we could start of with a simple occ command to transfer the ownership of a board to a different user (similar to how it works for files).

Dubidubiduu:
> That should work fine.

Almost.

My situation:

Board x - users:
A (deprecated user) - owner
B - shared with user

So ownership should be transformed from A to B. When I just change owner in database it becomes:

Board x - users:
B - owner
B - shared with user

And with this situation, somehow, I wasn't able to delete the share. So I gave the owner to another user C, deleted the share of B, and then gave the ownership to B.

Maybe it was just some caching and I wasn't patient enough, just wanted to let you know.

matchish:
I've implemented the feature for my needs as an `occ` command. If it has a value for the community I can create PR

juliushaertl:
@matchish Cool, a pull requests for that would bei highly appreciated. 🎉

benjamindahl:
in occweb deck doesn't offer deck:transfer-ownership for me. Am I doing something wrong?

`Sun Mar 14 2021 23:26:26Press Enter for more information on occ commands.
occ $ deck:transfer-ownership
                                                       Command "deck:transfer-ownership" is not defined.                                                         Did you mean one of these?                               deck:export                                          files:transfer-ownership                                                                            `

Cheers
Benny

[...]

PR #1955, opened 2020-05-25
Transfer ownership

matchish:
* Resolves: #1306
* Target version: master 

### Summary
Add `deck:transfer-ownership` command

### TODO

- [ ] Add tests

### Checklist

- [ x] Code is properly formatted
- [ x] Sign-off message is added to all commits
- [ x] Documentation (manuals or wiki) has been updated or is not required

### DIFF
--- a/lib/Command/TransferOwnership.php
+++ b/lib/Command/TransferOwnership.php
@@ -9,11 +9,9 @@ use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 
 final class TransferOwnership extends Command {
-
 	protected $boardService;
 
-	public function __construct(BoardService $boardService)
-	{
+	public function __construct(BoardService $boardService) {
 		parent::__construct();
 
 		$this->boardService = $boardService;
@@ -45,5 +43,4 @@ final class TransferOwnership extends Command {
 
 		$output->writeln("Transfer deck entities from $owner to $newOwner completed");
 	}
-
 }
--- a/lib/Command/TransferOwnership.php
+++ b/lib/Command/TransferOwnership.php
@@ -1,0 +1,63 @@
+<?php
+declare(strict_types=1);
+
+namespace OCA\Deck\Command;
+
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+final class TransferOwnership extends Command {
+
+	protected function configure() {
+		$this
+			->setName('deck:transfer-ownership')
+			->setDescription('Change owner of deck entities')
+			->addArgument(
+				'owner',
+				InputArgument::REQUIRED,
+				'Owner uid'
+			)
+            ->addArgument(
+                'newOwner',
+                InputArgument::REQUIRED,
+                'New owner uid'
+            );
+	}
+
+	protected function execute(InputInterface $input, OutputInterface $output) {
+		$owner = $input->getArgument('owner');
+		$newOwner = $input->getArgument('newOwner');
+        $db = \OC::$server->getDatabaseConnection();
+
+        $output->writeln("Transfer deck entities from $owner to $newOwner");
+        $params = [
+            'owner' => $owner,
+            'newOwner' => $newOwner
+        ];
+
+        $output->writeln('update oc_deck_assigned_users');
+        $stmt = $db->prepare('UPDATE `oc_deck_assigned_users` SET `participant` = :newOwner WHERE `participant` = :owner');
+        $stmt->execute($params);
+
+        $output->writeln('update oc_deck_attachment');
+        $stmt = $db->prepare('UPDATE `oc_deck_attachment` SET `created_by` = :newOwner WHERE `created_by` = :owner');
+        $stmt->execute($params);
+
+        $output->writeln('update oc_deck_boards');
+        $stmt = $db->prepare('UPDATE `oc_deck_boards` SET `owner` = :newOwner WHERE `owner` = :owner');
+        $stmt->execute($params);
+
+        $output->writeln('update oc_deck_board_acl');
+        $stmt = $db->prepare('UPDATE `oc_deck_board_acl` SET `participant` = :newOwner WHERE `participant` = :owner');
+        $stmt->execute($params);
+
+        $output->writeln('update oc_deck_cards');
+        $stmt = $db->prepare('UPDATE `oc_deck_cards` SET `owner` = :newOwner WHERE `owner` = :owner');
+        $stmt->execute($params);
+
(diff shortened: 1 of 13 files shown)


matchish:
Before add tests I would like to have small feedback about the command.
Also I wondering it's ok if I add integration tests for the command instead of unit?
I don't see a way to create good unit tests for the command.

stefan-niedermann:
Can you please also update the REST API documentation so 3rd party clients can implement this new feature? 🙂

lib/Command/TransferOwnership.php
```diff
@@ -0,0 +1,63 @@
...
+		$newOwner = $input->getArgument('newOwner');
+        $db = \OC::$server->getDatabaseConnection();
+
+        $output->writeln("Transfer deck entities from $owner to $newOwner");
+        $params = [
+            'owner' => $owner,
+            'newOwner' => $newOwner
+        ];
+
+        $output->writeln('update oc_deck_assigned_users');
```
juliushaertl, line 40: It would be nice if that could actually be moved to the BoardService class that it can be reused in the API as @stefan-niedermann mentioned. Ideally we

lib/Command/TransferOwnership.php
```diff
@@ -0,0 +1,63 @@
...
+        $output->writeln('update oc_deck_attachment');
+        $stmt = $db->prepare('UPDATE `oc_deck_attachment` SET `created_by` = :newOwner WHERE `created_by` = :owner');
+        $stmt->execute($params);
+
+        $output->writeln('update oc_deck_boards');
+        $stmt = $db->prepare('UPDATE `oc_deck_boards` SET `owner` = :newOwner WHERE `owner` = :owner');
+        $stmt->execute($params);
+
+        $output->writeln('update oc_deck_board_acl');
+        $stmt = $db->prepare('UPDATE `oc_deck_board_acl` SET `participant` = :newOwner WHERE `participant` = :owner');
```
juliushaertl, line 53: please check for the type of the board acl as well. Otherwise this will break if a group with the id of $newOwner is part of the acl.

lib/Command/TransferOwnership.php
```diff
@@ -0,0 +1,63 @@
...
+            'owner' => $owner,
+            'newOwner' => $newOwner
+        ];
+
+        $output->writeln('update oc_deck_assigned_users');
+        $stmt = $db->prepare('UPDATE `oc_deck_assigned_users` SET `participant` = :newOwner WHERE `participant` = :owner');
+        $stmt->execute($params);
+
+        $output->writeln('update oc_deck_attachment');
+        $stmt = $db->prepare('UPDATE `oc_deck_attachment` SET `created_by` = :newOwner WHERE `created_by` = :owner');
```
juliushaertl, line 45: I don't really see a point in changing the created_by column of attachments. The previous owner is still the one who uploaded the file.

lib/Command/TransferOwnership.php
```diff
@@ -0,0 +1,63 @@
...
+                'newOwner',
+                InputArgument::REQUIRED,
+                'New owner uid'
+            );
+	}
+
+	protected function execute(InputInterface $input, OutputInterface $output) {
+		$owner = $input->getArgument('owner');
+		$newOwner = $input->getArgument('newOwner');
+        $db = \OC::$server->getDatabaseConnection();
```
juliushaertl, line 32: Please use dependency injection in the constructor to get the required class instances.

lib/Command/TransferOwnership.php
```diff
@@ -0,0 +1,63 @@
...
+
+        $output->writeln("Transfer deck entities from $owner to $newOwner");
```
juliushaertl, line 34: We use tabs instead of spaces :wink:

juliushaertl:
Thanks a lot for looking into this. :heart: 

> Before add tests I would like to have small feedback about the command.
> Also I wondering it's ok if I add integration tests for the command instead of unit?

Sure, I agree that an integration test makes the most sense here.

matchish:
> Can you please also update the REST API documentation so 3rd party clients can implement this new feature?

It's an `occ` command. Not  a REST API what documentation I have to update?

stefan-niedermann:
Oh - uhm... Basically to gain feature parity we would need a REST endpoint additionally then? Should i open a new issue then?

How does the vue app trigger an owner change? Doesn't it has to call some kind of REST endpoint, too? Or am i completely wrong here?

matchish:
I believe this feature mostly for admins and they have access to occ commands. And if we want REST endpoint for the feature I think is better to open a new issue for it(for me smaller tasks always better)

lib/Command/TransferOwnership.php
matchish: done

lib/Command/TransferOwnership.php
matchish: done

lib/Command/TransferOwnership.php
matchish: Agree. removed

lib/Command/TransferOwnership.php
matchish: Outdated

lib/Command/TransferOwnership.php
matchish: Fixed

matchish:
Ready for review again) There is one issue. If user already participant of some card or in acl then command will stop with error unique constrain violation.  Do you think this could be a new task(issue), because this implementation already has an value?

juliushaertl:
Sorry for taking so long to review.

> There is one issue. If user already participant of some card or in acl then command will stop with error unique constrain violation. Do you think this could be a new task(issue), because this implementation already has an value?

I'd love to get this addressed before merge, since that is probably a common use case that the ownership is transferred to a user that is already part of the board. I'm not sure if there is an easy solution without manually iterating over the entries and update them one by one. 

MySQL would support an `UPDATE IGNORE` statement modifier but this doesn't seem to be available in PostgreSQL or sqlite so that is not an option for us here.

juliushaertl:
The code in general looks good to me :+1:

juliushaertl:
>  Lint / php-cs check (pull_request) Failing after 34s — php-cs check

juliushaertl:
The BoardServiceTest needs to be adjusted to the additional constructor parameter

matchish:
> The BoardServiceTest needs to be adjusted to the additional constructor parameter

What do you mean?

matchish:
@juliushaertl There is failed check https://drone.nextcloud.com/nextcloud/deck/4233/2/2
Could you help me with debug, don't understand how to reproduce it locally? All my tests pass locally

juliushaertl:
Sorry for the delay, unfortunately I totally missed your message. I've rebased and pushed it to a branch of this repo which makes collaborating on the branch a bit easier https://github.com/nextcloud/deck/pull/2496

Since you are a member of the nextcloud organization you should also have push access to that.

PR #1976, opened 2020-06-02
Delete boards that users own once they are deleted

juliushaertl:
Make sure that users boards are properly removed on user delete


stefan-niedermann:
Will this solution cover also boards that has been shared with the "deleted" user?

juliushaertl:
> Will this solution cover also boards that has been shared with the "deleted" user?

Yes, user acl entries already get removes some lines above https://github.com/nextcloud/deck/blob/master/lib/AppInfo/Application.php#L120

stefan-niedermann:
Sorry for nut picking... but maybe you should also add a clean step on app upgrade which check for cases where boards are present but user has been deleted, just to make sure?

It's still potential leaked data...

PR #2047, opened 2020-06-19
show card details in modal

jakobroehrl:

* Target version: master 

grafik
grafik
grafik


stefan-niedermann:
May i ask what the intention of this modal is?
Is it to enable a user to edit distraction-free?
Because if the intention is to give the user more space to edit and view the description - it's not that much difference yet :wink:

jakobroehrl:
> 
> May i ask what the intention of this modal is?
> Is it to enable a user to edit distraction-free?
> Because if the intention is to give the user more space to edit and view the description - it's not that much difference yet 😉

Sure! 
 - more space and distraction-free editing
 - use it for showing the content of archived cards

juliushaertl:
- [ ] Modal text should not be centered - I assume there is a to wide css rule somewhere
- [ ] Check how multiple modals behave e.g. when adding an attachment
- [ ] Move the action to the sidebar actions menu
- [ ] The sidebar should not be shown if the modal is used
- [ ] Show activity/attachments/comments in the modal. This is especially required for showing archived cards

In a second step we could allow users to use the modal as a default opener.

juliushaertl:
@jakobroehrl I took the liberty to push two commits that I came up with when trying out your branch. This basically makes sure that the modal can be properly reopened and in the same run adds a checkbox to toggle between default open for the sidebar and default open the modal. I've added another TODO to the above list for persisting that state.

jakobroehrl:
> Fix styling in the modal view, not sure if this works properly with just reusing the sidebar view or of we need to come up with our own tab implementation, but the first should of course be preferred

@juliushaertl To do this, we need to remove the following CSS style from the sidebar: width, max-width, border-left
Is this possible?

juliushaertl:
> @juliushaertl To do this, we need to remove the following CSS style from the sidebar: width, max-width, border-left Is this possible?

Yes, i think for that we can just overwrite the styles with a v-deep selector in in the modal.

jakobroehrl:
@juliushaertl 
This does work if I add it here: https://github.com/nextcloud/deck/blob/f93be516beb5cb9937c768bb46b3d27764c3f579/src/App.vue#L111-L115

But it affects all sidebars as well, how can we change this?

juliushaertl:
Pushed the rebased version and removed the for now unneeded details tab component. Let's keep the changes minimal for now.

@jakobroehrl Mabye you can have another look at the modal styling, the scrollbar looks out of place:

image

jakobroehrl:
@juliushaertl 
Thanks for your commit, let's merge this PR!

juliushaertl (approved):
Changes are good to go from my side now :+1: 

@jakobroehrl Mind to make sure that CI passes? :wink:

PR #2115, opened 2020-07-11
Dashboard panel

juliushaertl:
First attempt to implement a dashboard panel for deck to display assigned cards with a due date soon.

Based on the work by @jakobroehrl in https://github.com/nextcloud/deck/pull/1934 as it uses similar collections of data that we can reuse then.

image

- [x] The requirement of being assigned doesn’t work well for instances you use on your own. Can we check if there’s assigned cards at all, and if not, show the non-assigned ones you created for example?
- [x] No emptycontent view
- [ ] auto update of the deck dashboard?

cc @jancborchardt

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -136,7 +136,6 @@ return [
 		['name' => 'comments_api#delete', 'url' => '/api/v1.0/cards/{cardId}/comments/{commentId}', 'verb' => 'DELETE'],
 
 		// dashboard
-		['name' => 'overview_api#findAllWithDue', 'url' => '/api/v1.0/overview/due', 'verb' => 'GET'],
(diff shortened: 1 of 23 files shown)


jakobroehrl:
1. Which kinds of cards should be shown here? I get some due of my due cards and one without due
2. Is there an auto update of the dashboard? Or do I need to reload the dashboard page?

jancborchardt (approved):
Looks good from what you showed me @juliushaertl – as said only the "More items …" should better be "More assigned cards …" for accessibility. :)

juliushaertl:
> as said only the "More items …" should better be "More assigned cards …" for accessibility. :)

Pushed already :)

juliushaertl:
@jakobroehrl Could you have another look? I'd say we get this in as it is and iterate over the rest in follow-ups.

jakobroehrl (approved):
super nice!

PR #2118, opened 2020-07-11
Cleanup unneeded cardId parameter

juliushaertl:


stefan-niedermann:
This is a new minor version of the REST API, isn't it? Minor because it is fully backward compatible, but it should be documented in the REST API documentation in this case.

juliushaertl:
> No, we don't change anything, just the parameter is not really needed so we don't need to pass it around

Issue #2048, opened 2020-06-19
inconsistent design of mouseover text


fm-sys:
This button bar is using the browser mouseover design...
![2020-06-19 14_37_06-Deck - Nextcloud](https://user-images.githubusercontent.com/64581222/85136265-ab9b8000-b23f-11ea-8e16-57d6f5010e51.png)

And here it's the nextcloud stile...
![2020-06-19 14_37_51-Deck - Nextcloud](https://user-images.githubusercontent.com/64581222/85136268-ab9b8000-b23f-11ea-9cc7-8fd066e22353.png)

I can reproduse this on all my browsers (firefox and chromium edge)

juliushaertl:
Yes, this should all be the Nextcloud tooltip.

@jakobroehrl Maybe something for you :)

cloud2018:
Could the text also standardized?

- Add card
- Add new list
- Create an new board (item and tooltip)

jakobroehrl:
@cloud2018 
You mean to change eg from "New board title" to "Create new board"?

fm-sys:
I think it should be **list name / board title** or **new list name / new board title**

**Add a new card** is fine, because it is displayed even after the creation is finished and you may add second card...

PR #2123, opened 2020-07-15
fix control tooltip

jakobroehrl:

* Resolves: # #2048
* Target version: master


juliushaertl:
@jakobroehrl Sorry, aparently I've merged to early since it was still on "2. developing" 

Please just open another PR for follow up work.

PR #2158, opened 2020-07-31
Fix description in dark mode

juliushaertl:


css/icons.scss
```diff
@@ -4,6 +4,9 @@
 .icon-deck {
 	background-image: url('../img/deck-dark.svg');
 }
+body.dark .icon-deck {
```
nickvergessen, line 7: to me this sounds like a hack.
With the navigation icon there is a `filter: invert` applied.

I see a similar problem now with the dashboard and unified search. I think we should think about a generic solution to this problem.

juliushaertl: I've removed the whole hardcoded .icon-deck now since it was a fallback for times before we had the icon-back-white mixin anyways.

Issue #1650, opened 2020-04-01
Caching-Options via REST-API


stefan-niedermann:
@desperateCoder @juliushaertl 

The REST-API works great, but it is a huge problem to cache the data. This causes a full synchronization of a client (hmm, let's say [Nextcloud Deck for Android 😉](https://github.com/stefan-niedermann/nextcloud-deck)) to take up to 30 seconds (with only a few test-boards and -cards) and causing *very much* traffic.

This issue is about providing the necessairy capabilities for proper caching via the REST-API.

# Implement `ETag` and `If-None-Match`

I took some time and thought about it. The obvious way is to implement `ETag`- and `If-None-Match`-Headers. The basic idea is, that the REST-API has a checksum about the current state of a response and sends it to the client with the first response. The client will send the value of the retrieved `ETag`-Header as value of a `If-None-Match`-header at the next request and the server will simply respond with a `304` status code if it matches the current `ETag`.

## Desired behavior

1. Only need one request given nothing has changed on the server.
2. Walking in tree-style through the entity hierarchy - only need request for the subtree items that actually have changed

![grafik](https://user-images.githubusercontent.com/4741199/78116380-7c296400-7404-11ea-9ac3-909fbdaa6801.png)

## Problem with the standard implementation

Though there is one problem with the standard implementation: We have multiple entities in our hierarchy:

1. Account
2. Board
3. List
4. Card

The standard implementation generates a hash from all the properties of the entity (e.g. in the case of a board this might only be `title` and `color` but **not** the contained lists and cards). Therefore we can *save the traffic* generated by each request but we still have *one request for each entity* (One per account, one per board, one per list, one per card).

## Proposal 1: Implement `ETag` in a "non-standard" way

The `ETag` of each entity could not only be generated from the direct properties, like

`BoardETag: etag(title, color)`

but also from all descendant entities. Sample:

`BoardETag: etag(title, color, ListETag[])`

where the etag of a list item is generated like

`ListETag: etag(title, CardETag[])`

and so on.

### Pros

This is my personal preferred solution as

- One does not have to deal with multiple custom HTTP headers
- One could argue, that the descendant entities actually *are* properties of the current entity, but they are just not send to the clients in their responses 😉
- Quoting from Wikipedia: `The method by which ETags are generated has never been specified in the HTTP specification.` which means, that we are not sooo "non-standardish" but it's just not so common maybe.

### Contra

[...]

Issue #925, opened 2019-03-08
Exclusive write access


jgrete:
We use Deck in small team of five to manage our work. We would like to have a checkbox to gain exclusive write access to a card. We often had the situation that one team member overwrites some other members changes.
Regards
Juergen

juliushaertl:
Yes, we should definitely have some locking in place. This is also related to https://github.com/nextcloud/deck/issues/256 so the content gets updated once someone else pushes changes.

juliushaertl:
https://fideloper.com/etags-and-optimistic-concurrency-control might be the way to go with this together with https://github.com/nextcloud/deck/issues/1650

juliushaertl:
cc @stefan-niedermann @desperateCoder as this might be also very useful for the mobile app

jgrete:
Nice to hear that my wish isn't lost in time and spacer :-)

Issue #256, opened 2017-07-11
Contents synchronisation / Live update for changes


9662:
Just putting this up for discussion, this is neither a bug report nor necessarily a feature request.

### Steps to reproduce
1. Access the Deck application on computer Alpha as user Albert
2. Access the Deck application on computer Beta as user Albert
3. Access the Deck application on computer Gamma as user Bettie
4. Have Albert@Alpha make some changes, e.g.:
   * CRUD operation on deck
   * CRUD operation on board
   * CRUD operation on card
   * CRUD operation on tag

### Expected behaviour
Either:
* changes are pushed to other clients; or
* other clients are notified of changes to the board (they can then pull the actual changes, perhaps manually, or F5-refresh).

### Actual behaviour
Different users have inconsistent views of the Deck state.

## Current status

- [ ] Clarify UX
- [ ] Lay out plan for live update mechanism and possible integration with notify_push

juliushaertl:
@9662 Yes, this is something we should have. Since we have no way of pushing information to the client we need some kind of polling mechanism like the notifications app does to check if there were changes on a board and update it then.

9662:
On second thought, maybe this should be considered at a higher level in NC? After all, the same problem is evident in other applications, e.g., if viewing a given director and a file is uploaded to / deleted from it, the changes are not reflected until a manual refresh. Same with contacts, calendar...

A generic update framework (piggybacking on notifications?) might make sense, or not? I am not familiar enough with NC's architecture to answer that question.

pixelipo:
Why not use Activity app instead? We could log changes to boards there and notify current user in case a shared board has been updated

9662:
> and notify current user in case a shared board has been updated

To confirm, would the following user story accurately describe your suggestion?

“As a user having a board open on a browser page, when that board is updated from a different session, I should receive a conspicuous but non-intrusive notification that the board's state has changed.”

UkeFan:
Yes! That would be gorgeous!

bsherwoodofdaptiv:
As a note / thought - if the browser received a 'dirty page' message, it could auto reload the boards (or affected column / or card). Just thinking out load here.

putt1ck:
Even just a refresh button would be neat

bsherwoodofdaptiv:
Adding another thought - an application (single page javascript app) I wrote (as an amateur) recently has a ping routine that just updates the central server every 60 seconds to say 'hey, I am still running (ie: user is still logged in)'.
[...]

PR #2159, opened 2020-08-01
WIP: Naive approach on content sync

juliushaertl:
First attempt to tackle https://github.com/nextcloud/deck/issues/256

- [x] Currently uses the lastModified date to detect changes with polling
- [ ] Optionally sent current server time with responses (can be avoided when switching to etags)

UI wise we currently don't trigger an update of the user interface to avoid ui element that are changing places without user interaction, so instead a small banner is shown that the view is outdated which allows the user to trigger a refresh manually. cc @nextcloud/designers Does that make sense? I guess ideally we could just update all data that doesn't affect any item position changes. So a card reordering would still trigger the banner while simple label or text changes would be shown immediately in the UI.

image

## Possibly better approaches
- Use https://github.com/nextcloud/push once available
- Block updates and detect changes on ETag mismatch (#1650 #925)
  - https://fideloper.com/etags-and-optimistic-concurrency-control

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5254,29 +5254,24 @@
 			}
 		},
 		"@nextcloud/event-bus": {
-			"version": "1.1.4",
-			"resolved": "https://registry.npmjs.org/@nextcloud/event-bus/-/event-bus-1.1.4.tgz",
-			"integrity": "sha512-It27KzmUaSQ7w22nHFwOn8XgeVG0HYYOSNG9gs4UkP5VqcZ16m4ydt3GkMpWcyFec4OUjJc+yf7omRc3pNxsSw==",
+			"version": "1.2.0",
+			"resolved": "https://registry.npmjs.org/@nextcloud/event-bus/-/event-bus-1.2.0.tgz",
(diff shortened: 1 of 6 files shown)


dodedodo:
> UI wise we currently don't trigger an update of the user interface to avoid ui element that are changing places without user interaction

I feel like this should be optional. Situations arise where collaborative team meetings trigger a lot of overdue workflow updates. It's useful to actually see these changes being made in real-time (or within seconds).

juliushaertl:
> UI wise we currently don't trigger an update of the user interface to avoid ui element that are changing places without user interaction, so instead a small banner is shown that the view is outdated which allows the user to trigger a refresh manually. cc @nextcloud/designers Does that make sense? I guess ideally we could just update all data that doesn't affect any item position changes. So a card reordering would still trigger the banner while simple label or text changes would be shown immediately in the UI.

@jancborchardt Any input on this?

juliushaertl:
:ping_pong: @jancborchardt :)

honzapatCZ:
How si this going? This could be real usefull.

juliushaertl:
Closing for now, to be reopened when continuing, to get the list of open prs a bit more sorted.

PR #2166, opened 2020-08-01
Adjust board background to new design of tasks app

### LINKED ISSUE #1892: Change the background
I know that it's cosmetic thing but it would be really nice if the possibility to change the background and color could be available. Of course there should be a parameter to set the opacity of the background.

As you can see on the below screen, the items/notes are merging with the white backround. Item/note card is white, the background is white as well. 

image

On below example is presented that difference colors could make it more interesting and cleared for view. A light gray background and small shadow can do it better! 🥇 

image

Please, keep going, your project organizes my work well.

simonspa:
* Target version: master 
* Resolves #1892

### Summary

This adjusts the background of boards to be slightly darker, following the recent design change in the Tasks app (https://github.com/nextcloud/tasks/pull/1136). i think the individual cards stick out better from this background.

image

This would of course be quite a change that, I guess, should be discussed (also cc'ing @jancborchardt here...)

Also, with stronger visual distinction between board and app sidebar, it feels like the sidebar elements are too close to its edge.

### Checklist

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -98,6 +98,7 @@ export default {
 			position: relative;
 			overflow-x: hidden;
 			align-items: stretch;
+			background-color: var(--color-background-dark) !important;
 		}
 
 		#app-sidebar {
--- a/src/components/board/Stack.vue
+++ b/src/components/board/Stack.vue
@@ -240,7 +240,6 @@ export default {
 		margin-right: -10px;
 		margin-top: 0;
 		margin-bottom: 3px;
-		background-color: var(--color-main-background-translucent);
 		cursor: grab;
 
 		h3, form {
@@ -267,7 +266,6 @@ export default {
 		height: 52px;
 		z-index: 100;
 		display: flex;
-		background-color: var(--color-main-background);
 		margin-left: -10px;
 		margin-right: -10px;
 		padding-top: 3px;
@@ -279,6 +277,7 @@ export default {
 			margin-top: 0;
 			margin-bottom: 10px;
 			box-shadow: 0 0 3px var(--color-box-shadow);
+			background-color: var(--color-main-background);
 			border-radius: 3px;
 		}
 


juliushaertl:
@jancborchardt What do you think? I think it still could be a bit lighter than the default background-dark

karlitschek:
Hmm. I'm not sure. Looks a bit boring and old school. ;-) Maybe make the background a bit lighter. Maybe some more rounded corners of the cards. And in general maybe take some inspiration from the new dashboard? @juliushaertl @jancborchardt

jancborchardt:
Several points or questions here:
- Do we really want to go with a non-white background in general? As this will make the Deck app look separate from the rest of Nextcloud apps
- If we go with non-white, it shouldn’t be so dark. Needs to be an off-white and not grey, like e.g. `#f7f9fc` (tinted slight blue, should possibly be adjusted based on the primary theming color)
- Do we even want to go with e.g. simply the background image of the Dashboard?
- Agree with Frank we can go with more rounding of the cards, like `var(--border-radius-large)`

raimund-schluessler:
For Tasks we introduced a grey background with https://github.com/nextcloud/tasks/pull/1136. Having it equal in Deck would make sense, I think.

The background image from Dashboard could also be nice, then you could choose whether an image or plain color or whatever is shown (ideally complemented by the ability to select a different background color than the primary one for the Dashboard).

juliushaertl:
> Do we really want to go with a non-white background in general? As this will make the Deck app look separate from the rest of Nextcloud apps

I personally prefer the white one as well, but we have this request from time to time where people suggest to add some more contrast between cards and the background.

> If we go with non-white, it shouldn’t be so dark. Needs to be an off-white and not grey, like e.g. #f7f9fc (tinted slight blue, should possibly be adjusted based on the primary theming color)
> Agree with Frank we can go with more rounding of the cards, like var(--border-radius-large)

The screenshot in here was a bit outdated, the border radius is already adjusted. Here is an updated one with the slight blue-ish background as you proposed. Would be fine by me this way:

image

> Do we even want to go with e.g. simply the background image of the Dashboard?

I personally think that a background image adds a lot of distraction from the actual content, so I'd rather not have this by default at least.

juliushaertl:
@simonspa Mind to adjust that to the proposal above and rebase? :)

juliushaertl (changes requested):
@simonspa Mind to adjust that to the proposal above and rebase? :)

juliushaertl:
Moving the milestone, since we would need a server css variable first as discussed via chat.

PR #2208, opened 2020-08-20
Fix design, dark mode and keyboard navigation of the board list

juliushaertl:
- Add 2px border in dark mode
- Increase card border radius
- Fix design of the all boards listing and allow keyboard navigation
- Remove legacy fallback for icons to make sure they are properly set in dark mode

### DIFF
--- a/css/collections.css
+++ b/css/collections.css
@@ -1,7 +1,3 @@
-.icon-deck {
-	background-image: url('../img/deck-dark.svg');
-}
-
 .resource-type-deck img {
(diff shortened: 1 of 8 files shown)


jakobroehrl:
Looks good, some possible dark mode changes:
grafik
grafik

General:
This could be smaller? 
grafik

juliushaertl:
The deck icon seems to be a caching issue, works fine here. The others are fixed (with a new release of nextcloud-vue-collections@0.8.1)

Issue #2184, opened 2020-08-08
Make columns independently scrollable


despens:
When doing triage, you usually start with a first column of cards that you want to distribute across your columns. This means that the first column is very tall, filled with a lot of cards and the other columns are empty. To reach bottom cards in the first column it is required to scroll down. When you drag a card from the bottom of the first column into an empty column, the card gets placed on top of that column, which in many cases can would then be scrolled off-screen. The cards appears to vanish.

If each column could be scrolled independently, you could look at the bottom of one column and the top of another, making such triaging tasks much easier.

stefan-niedermann:
Using the <kbd>START</kbd> and <kbd>END</kbd> keys on your keyboard might be a sufficient workaround?

juliushaertl:
Yes, that seems to be also very common among other kanban-like solutions like trello and github projects. cc @nextcloud/designers

PR #2230, opened 2020-08-24
Implement scrolling per stack

juliushaertl:
Implements #2184 

Peek 2020-10-13 11-36

- Scroll each stack individually
- Fade out cards at the top when scrolling
- Scroll to card once inserted

@nextcloud/designers What do you think? Feels a bit more natural to use than always scrolling all lists

### DIFF
--- a/src/components/board/Stack.vue
+++ b/src/components/board/Stack.vue
@@ -282,7 +282,6 @@ export default {
 		margin-right: -10px;
 		margin-top: 0;
 		margin-bottom: 3px;
-		background-color: var(--color-main-background-translucent);
 		cursor: grab;
 
 		h3, form {
--- a/src/components/board/Stack.vue
+++ b/src/components/board/Stack.vue
@@ -23,14 +23,14 @@
 
 <template>
 	<div class="stack">
-		<div v-click-outside="stopCardCreation" class="stack--header" :class="{'stack--header--add': showAddCard }">
+		<div v-click-outside="stopCardCreation" class="stack__header" :class="{'stack__header--add': showAddCard }">
 			<transition name="fade" mode="out-in">
 				<h3 v-if="!canManage || isArchived">
 					{{ stack.title }}
 				</h3>
 				<h3 v-else-if="!editing"
 					v-tooltip="stack.title"
-					class="stack--title"
+					class="stack__title"
 					@click="startEditing(stack)">
 					{{ stack.title }}
 				</h3>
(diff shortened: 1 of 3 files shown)


stefan-niedermann:
Really *awesome* it was if the scrollbar thumbs were a bit thiner and hidden until one hovers a column. Should be possible with pure CSS, at least in Chrome.

juliushaertl:
> Really awesome it was if the scrollbar thumbs were a bit thiner and hidden until one hovers a column. Should be possible with pure CSS, at least in Chrome. (imho they have too much attention)

I also thought about that but was a bit concerned about the discoverability then.

juliushaertl:
Let's move this to the next release

jancborchardt (commented):
Looks great, only some details – your call if to be fixed here or separately:
- When you click the "Add" button, the entire stack moves a bit down cause of the new element. Would be nicer if it wouldn’t move
- The autoscrolling to the inserted card could use a quick animation so people know what happened
- Agree with @stefan-niedermann about the scrollbar, but we could also leave that for now for discoverability indeed.

juliushaertl:
> When you click the "Add" button, the entire stack moves a bit down cause of the new element. Would be nicer if it wouldn’t move

But then we would overlay the existing cards, which I would prefer not to do.

> The autoscrolling to the inserted card could use a quick animation so people know what happened

:+1: Let me see if i can make that happen.

> Agree with @stefan-niedermann about the scrollbar, but we could also leave that for now for discoverability indeed.

Then let's keep it for now. We can still experiment with hiding it in a follow up.

juliushaertl:
:ping_pong: for reviews again :)

src/components/board/Stack.vue
```diff
@@ -23,7 +23,7 @@
 
 <template>
 	<div class="stack">
-		<div v-click-outside="stopCardCreation" class="stack--header">
+		<div v-click-outside="stopCardCreation" class="stack--header" :class="{'stack--header--add': showAddCard }">
```
skjnldsv, line 26: That's a weird BEM syntax ^^

juliushaertl: Indeed :see_no_evil: adjusted with https://github.com/nextcloud/deck/pull/2230/commits/78299b3466dfa84b416d219e4e9e35e64724ff5e

skjnldsv (approved):

der-flo:
This change makes it hard to scroll on iOS because every finger tip immediately starts dragging a card.

der-flo:
> This change makes it hard to scroll on iOS because every finger tip immediately starts dragging a card.

@juliushaertl Should I open a new issue?

PR #2244, opened 2020-08-29
Toggle filter on clicking card labels

juliushaertl:
Apply or remove a filter for labels by clicking on the label in a card item

I'd also like to apply https://docs.vuestorefront.io/guide/vuex/vuex-conventions.html for vuex naming conventions in the future, so whenever we change something we should regarding stores we should move this over step by step.


jakobroehrl (approved):
very nive!

PR #2245, opened 2020-08-29
ETag support for API endpoints

juliushaertl:
Implements a simplified version of https://github.com/nextcloud/deck/issues/1650

- [x] ETag support for listing endpoints (board/stack)
- [x] ETag support for single object endpoints
- [x] Check individual modifying operations if they properly update the etag (last modified)
  - [x] Comments do not update the last modified date yet

## Follow up

- [ ] ETag support for changing operations (ref https://github.com/nextcloud/deck/issues/925)
- [ ] See if we need a separate field to store the etag so that it works somehow independent from the actual last modified date which is presented to the user
	- This would kind of make sense to not change the last modified date of a board, e.g. if a card was moved to the archive

cc @desperateCoder @stefan-niedermann

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -69,6 +69,31 @@ curl -u admin:admin -X GET \
     -H "If-Modified-Since: Mon, 05 Nov 2018 09:28:00 GMT"
 ```
 
+### ETag
+
+An ETag header is returned in order to determine if further child elements have been updated for the following endpoints:
+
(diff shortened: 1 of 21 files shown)


juliushaertl:
@stefan-niedermann sorry but I'll need to move this to the next minor since lots of other stuff blocked me from working on this

stefan-niedermann:
@desperateCoder let's rock 'n' roll 🤘 🎸 🎵

Issue #2185, opened 2020-08-09
Links in the deck notifications of the Activity app do not refer to the respective cards


baraksoa:
In NextCloud 19.0.1 server on the automatic Activity emails, there is a wrong NextCloud Deck/Card link, in the link there is an extra / sign. Example: mynextcloudserver/apps/deck/#/board/1//card/3

## Steps to reproduce
1. Configure the Activity e-mail settings(Deck/Card options too). 
2. Do something in the Deck app, make a new card... 
3. Check the automatic email Deck/Card link.

### Expected behaviour
Correct, existing Deck/Card link.

### Actual behaviour
Wrong Deck/Card link, with an extra / sign. Example: mynextcloudserver/apps/deck/#/board/1//card/3

![nextcloud_deck](https://user-images.githubusercontent.com/49455883/89768016-4d7e7100-dafb-11ea-8323-93a9344a7281.png)

putt1ck:
I can confirm this is issue in earlier versions of NC also; in addition the extra "/" is also found in the activity stream entry for Deck items.

stefan-niedermann:
Maybe related to https://github.com/nextcloud/deck/issues/1698

putt1ck:
@stefan-niedermann could be, but that's marked as fixed. Maybe a regression?

strassan:
Same issue for me. I don't use the emails but it's in the activity log.

stefan-niedermann:
I renamed the title as suggested in #2199 - please check whether this matches your issue and change back if you meant something else 🙂

putt1ck:
NB: links are similarly broken in the timeline for a card (presumably is just a filter of/feed to the NC activity log, but shows it is consistent. @juliushaertl if nobody else is looking at this we can take it on

PR #2251, opened 2020-09-01
Fix double slash in the deck activity links

baraksoa:
* Resolves: #2185


baraksoa:
I hurried this one and forgeted about the archived cards.

PR #2255, opened 2020-09-02
Use unified search events to apply on board filtering

juliushaertl:
Requires https://github.com/nextcloud/server/pull/22526 on the server

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3479,29 +3479,24 @@
 			}
 		},
 		"@nextcloud/event-bus": {
-			"version": "1.1.4",
-			"resolved": "https://registry.npmjs.org/@nextcloud/event-bus/-/event-bus-1.1.4.tgz",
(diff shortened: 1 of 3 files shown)


jakobroehrl (approved):
This is working, but I dont know how useful it is. 
If the user does any click while showing the search results the search results are gone

juliushaertl:
Yeah, we could just not reset the search query in deck when the unified search is closed but there would be no indicator or way to reset it then afterwards of course.

@skjnldsv Were there any discussions about that already UX wise? Issue in summary: 
- Unified search is opened, somthing is typed
- The event is passed and the app filters the current view
- The user clicks outside 
- The search gets reset

skjnldsv (approved):

skjnldsv:
> @skjnldsv Were there any discussions about that already UX wise? Issue in summary:
> 
> * Unified search is opened, somthing is typed
> * The event is passed and the app filters the current view
> * The user clicks outside
> * The in app search gets reset

sorry I missed this comment. Yeah, you fixed it with the reset/close event differences :tada:

PR #2270, opened 2020-09-07
Fix empty content view to align with other widgets

juliushaertl:
Implements what was described in https://github.com/nextcloud/server/issues/22339

image


eneiluj (approved):
Nicer indeed.

Issue #1794, opened 2020-05-03
Click outside card should close sidebar


simonspa:
The sidebar opens when clicking on a card, and switches when clicking on another card. To stay in this logic, clicking outside of a card should close the sidebar.

Currently the user has to navigate to the `x` button top-right to close it.

jakobroehrl:
@juliushaertl How can solve this issue? Could you give me a hint?

simonspa:
There is a Vue module called `v-click-outside` which is used somewhere in `spreed` for example - maybe that helps? (don't know myself...)

fm-sys:
If you edit a card title and click outside the card, the editing bar is closed. This may give us a hint how we could implement this feature for closing the sidebar...
![2020-06-22 12_49_15-Deck - Nextcloud](https://user-images.githubusercontent.com/64581222/85279561-eb07dd80-b486-11ea-81b7-be4c1682df14.png)

fm-sys:
Still very important feature! Can someone please, please implement this?

stefan-niedermann:
@fm-sys Please use **[GitHub reactions](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/)** 👍 to show that you are affected by the same issue. Please don't comment if you have no relevant information to add.

jakobroehrl:
The Esc key should close the sidebar, too

PR #2272, opened 2020-09-07
click outside to close sidebar

jakobroehrl:

* Resolves: #1794
* Target version: master 

With this change the sidebar no longer opens at all. 
@juliushaertl do you have an idea how to solve it?

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -23,7 +23,7 @@
 <template>
 	<NcContent app-name="deck" :class="{ 'nav-hidden': !navShown, 'sidebar-hidden': !sidebarRouterView }">
 		<AppNavigation />
-		<NcAppContent @click="clickCloseSidebar()">
+		<NcAppContent>
 			<router-view />
 		</NcAppContent>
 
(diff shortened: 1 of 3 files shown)


juliushaertl (changes requested):
Doesn't work that easy, as this basically blocks the sidebar from being opened in the first run when clicking on a card.

jakobroehrl:
@juliushaertl How to do this?
Would it be easier to implement closing with the esc key?

juliushaertl:
Well escape key should probably also work but the click would be still useful. To make this work you would basically need to make sure that the click was triggered on the board dom element and not on a card in the v-click-outside handler.

jakobroehrl:
It doesn't work with that either: Sidebar no longer opens
https://github.com/nextcloud/deck/blob/5e3661cd6a72af9d4ed2b1011e735d0d8f26171d/src/App.vue#L26

marcelklehr:
> To make this work you would basically need to make sure that the click was triggered on the board dom element and not on a card in the v-click-outside handler.

Fixed by using @juliushaertl's approach :)

juliushaertl (approved):

juliushaertl:
Seems to have caused a regression when trying to open the board sidebar as that click opens the sidebar but closes it directly again. Maybe we should only trigger it when clicking the board content area somehow.

PR #2275, opened 2020-09-08
Wait for services to be registered before performing further setup that requires services

juliushaertl:
Otherwise databaseType and database4ByteSupport are not being resolved properly

### DIFF
--- a/lib/AppInfo/Application20.php
+++ b/lib/AppInfo/Application20.php
@@ -59,6 +59,7 @@ use OCP\IServerContainer;
 use OCP\IUser;
 use OCP\IUserManager;
 use OCP\Util;
+use Psr\Container\ContainerInterface;
 
 class Application20 extends App implements IBootstrap {
 	public const APP_ID = 'deck';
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):

lib/AppInfo/Application20.php
```diff
@@ -84,6 +85,12 @@ public function boot(IBootContext $context): void {
 		$notificationManager = $context->getServerContainer()->get(\OCP\Notification\IManager::class);
 		$notificationManager->registerNotifierService(Notifier::class);
 		\OCP\Util::addStyle('deck', 'deck');
+
+		$this->registerUserGroupHooks();
+
+		$this->registerCommentsEntity();
```
ChristophWurst, line 91: tip: use ``\OCP\AppFramework\Bootstrap\IBootContext::injectFn`` so you don't have to locate services manually but have them injected as args :)

juliushaertl: Oh nice :+1:

lib/AppInfo/Application20.php
```diff
@@ -154,8 +155,8 @@ private function registerUserGroupHooks(): void {
 		});
 	}
 
-	public function registerCommentsEntity(): void {
-		$this->server->getEventDispatcher()->addListener(CommentsEntityEvent::EVENT_ENTITY, function (CommentsEntityEvent $event) {
+	public function registerCommentsEntity(SymfonyAdapter $symfonyAdapter): void {
```
ChristophWurst, line 158: `\OCP\Comments\CommentsEntityEvent` is a "new" event based on `\OCP\EventDispatcher\Event`. Aren't you able to use our dispatcher abstraction?

Issue #589, opened 2018-08-16
Card Title can be empty and hence not be edited anymore


fpetry:
**Describe the bug**
If I create a card and then click it and edit the title  to an empty string on the right, I cannot edit it afterwards because of the way it is displayed (field gets smaller and cannot be clicked anymore)

**To Reproduce**
1. Click the big "+" to create a new card
2. Write anything as title
3. Click the card and edit the title to an empty string
4. Click somewhere out of the edit field (Pressing enter pops up a message that the field should not be empty)
5. Click the card again
6. The edit field is much smaller and cannot be written to

**Expected behavior**
The title before the edit should be restored

**Screenshots**
![screenshot_20180816_153859](https://user-images.githubusercontent.com/4649876/44212434-a95fca80-a16b-11e8-8014-bc24907e3848.png)
![screenshot_20180816_154147](https://user-images.githubusercontent.com/4649876/44212436-ad8be800-a16b-11e8-9144-4ad64c0f1851.png)
![screenshot_20180816_154611](https://user-images.githubusercontent.com/4649876/44212441-afee4200-a16b-11e8-8b09-67225b1b0b66.png)

**Client details:**
 - OS: Kubuntu 18.04
 - Browser Chromium 
 - Version:  68.0.3440.75  
 - Device: desktop

fpetry:
It also works if you just write a " " when creating the card.

weeman1337:
I'll have a look at this.

matchish:
Regression bug in `1.0.1` version. Do I have to open a new issue or it's ok to continue here?

juliushaertl:
Ideally please open a new issue, but let's keep this for now

juliushaertl:
@jakobroehrl Make something small for you :wink:

PR #2318, opened 2020-09-22
card title: prevent space and no text

jakobroehrl:

* Resolves: #589 
* Target version: master

### DIFF
--- a/src/components/card/CardSidebar.vue
+++ b/src/components/card/CardSidebar.vue
@@ -466,7 +466,7 @@ export default {
 		},
 		updateTitle(newTitle) {
 			if (newTitle.trim === '') {
-				showError('Could not update card title')
+				showError(t('deck', 'The title cannot be empty.'))
 				return
 			}
 			this.$set(this.copiedCard, 'title', newTitle)
--- a/src/components/card/CardSidebar.vue
+++ b/src/components/card/CardSidebar.vue
@@ -227,6 +227,7 @@ import {
 	getMonthNamesShort,
 } from '@nextcloud/l10n'
 import moment from '@nextcloud/moment'
+import { showError } from '@nextcloud/dialogs'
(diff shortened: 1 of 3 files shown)


jakobroehrl:
@juliushaertl 
Do you have an idea why this is working here https://github.com/nextcloud/deck/blob/7c55d052e25ff49870426b0548e70eac9d9d7f6b/src/components/board/Stack.vue#L86

but here not: https://github.com/nextcloud/deck/blob/7c55d052e25ff49870426b0548e70eac9d9d7f6b/src/components/cards/CardItem.vue#L49

juliushaertl:
The form will only be checked if submitted not when a regular button is used with a click event 
https://github.com/nextcloud/deck/blob/7c55d052e25ff49870426b0548e70eac9d9d7f6b/src/components/cards/CardItem.vue#L51

src/components/card/CardSidebar.vue
```diff
@@ -464,6 +464,9 @@ export default {
 			this.descriptionSaving = false
 		},
 		updateTitle(newTitle) {
+			if (newTitle === '' || newTitle === ' ') {
```
juliushaertl, line 467:
```suggestion
			if (newTitle.trim() === '') {
```

src/components/card/CardSidebar.vue
```diff
@@ -464,6 +464,9 @@ export default {
...
+				return
```
juliushaertl, line 468: Would be nice to have some error handling, though I'm not sure the sidebar currently supports this, so maybe just show a toast with an error message.

stefan-niedermann:
So you are only limiting it in the UI - please keep in mind that performing this validation at the backend, it will be a breaking change :)

juliushaertl (approved):

PR #2323, opened 2020-09-24
Due date shortcuts

jakobroehrl:
* Target version: master 

Implemented like here: https://mengxiong10.github.io/vue2-datepicker/index.html#Shortcut 
But I see no shortcut in the Picker
grafik

@juliushaertl Do you have an idea what the problem could be? There are ne errors in the console.


juliushaertl:
We do not support this in our vue components yet: https://github.com/nextcloud/nextcloud-vue/blob/master/src/components/DatetimePicker/DatetimePicker.vue

I'd suggest to open an issue over there and see if we can get a standardized way of the shortcuts in to the component itself.

jakobroehrl:
@juliushaertl 
I could add to the card action menu 
 - Due: Today
 - Due: Tomorrow
 - Due: This week

This would be a very fast way to change the due date. What do you think about this?
grafik

juliushaertl:
Not a huge fan of this, there is already quite a lot in the menu there and I'd rather keep it simple. We should not overload the UI with tons of options.

jakobroehrl:
https://github.com/nextcloud/nextcloud-vue/issues/1419

jakobroehrl:
https://github.com/nextcloud/deck/pull/2684

Issue #2231, opened 2020-08-25
Card Overview: Move cards


jakobroehrl:
Deck will have a new Card Overview soon:
https://user-images.githubusercontent.com/24530680/90889077-b72e3300-e3b7-11ea-9b9d-8a93e3fb304d.png

On shared boards my assigned cards are shown and on non shared boards all cards with due dates.
The question is, what should happen to the due date if the card is moved to another column:
My suggestion:
 - Overdue: now
 - Today: now + 12h
 - Tomorrow: now + 1d
 - This week: now  + 1w
 - Later: now +1m

juliushaertl:
@nextcloud/designers @nextcloud/deck

stefan-niedermann:
My suggestion: On drop open a modal asking for a new due date. Your suggestions should be preselected.

Dubidubiduu:
For Today, Tomorrow and This week: Shouldn't it be the last minute of the actual naming?

Ex. Guess it's 18 o'clock, with your proposal moving a card to today will make it due at 06 o'clock tomorrow, which is confusing. Would be much better to make it due at 23:59 of today. Analog to tomorrow and this week (sunday 23:39 for most countries, but depending)

jakobroehrl:
Maybe it would be nice, to have in the every card sidebar the possibility to quickly change the due date to: Today, Tomorrow, this week. This should change the due date as when drag and drop a card in this Upcoming cards View

juliushaertl:
Something for a separate issue

jakobroehrl:
> 
> Something for a separate issue, but yes :) 
 -> https://github.com/nextcloud/deck/pull/2323

tomdereub:
@Dubidubiduu : now that there are calendars linked to the cards, if we set time at 23:59, it wont be visible on calendar without scrolling down. So I would suggest to keep the time of the day when moving a card to today, tomorrow etc..., just changing the day.
For example, if a card was set yesterday at 9:00 am, moving it to today will set the date to today at 9:00 am, moving it to next week will set the date to next tuesday at 9:00.
And with @stefan-niedermann 's idea it would be possible to easily change it if necessary.

Issue #1157, opened 2019-07-31
[Feature Request] All day due date


monsieurhannes:
Hi!
It would be great to be able to just select a date without having to define a time. I know I could just pick 23.55 and I have covered the whole day, but it would simplify the process.
A check box would do it for me, like in the calendar app:
![bild](https://user-images.githubusercontent.com/20651440/62213955-5dba2700-b3a4-11e9-8347-977de9045603.png)

Subsequently it would be great to show "today" in the card instead of "in 7 hours". But that's a minor thing...

juliushaertl:
Makes sense to me :+1:

jakobroehrl:
I think with the new VUE implementation this is solved, because 00:00 is the default time:

1. 
![grafik](https://user-images.githubusercontent.com/24530680/67361176-2c4a9500-f568-11e9-8639-6fed214e5ebf.png)

2. 
![grafik](https://user-images.githubusercontent.com/24530680/67361254-58feac80-f568-11e9-9caa-116a9064aa29.png)

3. 
![grafik](https://user-images.githubusercontent.com/24530680/67361278-6d42a980-f568-11e9-94b9-0e7accaef3b9.png)

4. 
![grafik](https://user-images.githubusercontent.com/24530680/67361298-792e6b80-f568-11e9-8793-a4ad8e89b9d1.png)

juliushaertl:
Not sure, a task being due on 00:00 is still different than just being due on a specific date.

monsieurhannes:
I think especially if there is a future implementation into the calendar app it would be good to differentiate. But I don't know how your thoughts are about that.
Still very good to have a default on 00:00!

sunjam:
I was looking for an `All Day` option, like is available on Tasks app.

roboticsalign:
> Not sure, a task being due on 00:00 is still different than just being due on a specific date.

I agree and would like to support this enhancement. I think, there are two relevant differences between 00:00 and all day:

1. The due date for an all day event is at the End of the day (~11:59pm), 00:00 is overdue before the work started: 
![Screenshot_20210107-100445_Chrome](https://user-images.githubusercontent.com/23562360/103873513-35257900-50d0-11eb-81ed-6e76f2acea2d.jpg)

2. A concrete time corresponds to a timezone and implies a worldwide unique date/time combination. An all day event could/should consider the timezone of the current user. If I create an all day event in UTC and move to a location in UTC-6, the due date should be shown as end of UTC-6.

[...]

PR #2332, opened 2020-10-01
add card in upcoming cards view

jakobroehrl:

grafik

### Ideas 
- [ ] close modal after card creation
- [ ] preselection of board and list
- [ ] second click on + doesn't open the modal
- [ ] update view after card creation
- [x] save last used board and stack (Android app does it)

### DIFF
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -201,6 +201,8 @@
 
 		<Modal v-if="modalShow" :title="t('deck', 'Add card on Today')" @close="modalShow=false">
 			<div class="modal__content">
+				{{ lastBoardId }}
+				{{ lastListId }}
 				<h3>{{ t('deck', 'Add card on Today') }}</h3>
 				<Multiselect v-model="selectedBoard"
 					:placeholder="t('deck', 'Select a board')"
@@ -280,6 +282,8 @@ export default {
 		...mapGetters([
 			'canEdit',
 			'canManage',
+			'lastBoardId',
+			'lastListId',
 		]),
 		...mapState({
 			compactMode: state => state.compactMode,
@@ -317,6 +321,15 @@ export default {
 	watch: {
 		board() {
 			this.clearFilter()
+		},
+		lastBoardId() {
+
+			if (this.lastBoardId === null || this.lastBoardId === 0) {
+				return
+			}
+			this.selectedBoard = this.boards.filter(board => {
+				return board.id === this.lastBoardId
+			})
 		},
 	},
 	methods: {
@@ -390,6 +403,8 @@ export default {
 					duedate: today.toISOString(),
 				})
 				this.newCardTitle = ''
+				this.$store.dispatch('storeLastListId', this.selectedStack.id)
+				this.$store.dispatch('storeLastBoardId', this.selectedBoard.id)
 			} catch (e) {
 				showError('Could not create card: ' + e.response.data.message)
 			}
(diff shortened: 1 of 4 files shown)


src/components/Controls.vue
```diff
@@ -32,6 +32,11 @@
 		</div>
 		<div v-if="overviewName" class="board-title">
 			<h2><a href="#">{{ overviewName }}</a></h2>
+			<Actions>
+				<ActionButton icon="icon-add" @click.stop="modalShow=true">
+					{{ t('deck', 'Add card on Today') }}
```
juliushaertl, line 37:
```suggestion
					{{ t('deck', 'Add card for today') }}
```

src/components/Controls.vue
```diff
@@ -193,18 +198,56 @@
 				</Actions>
 			</div>
 		</div>
+
+		<Modal v-if="modalShow" :title="t('deck', 'Add card on Today')" @close="modalShow=false">
+			<div class="modal__content">
```
juliushaertl, line 203: Please move this to a new component

src/components/Controls.vue
```diff
@@ -314,11 +383,59 @@ export default {
...
+				const response = await axios.get(url)
+				this.stacksFromBoard = response.data
+			} catch (err) {
+				return err
+			}
+		},
+		async addCard() {
+			try {
+				const today = new Date()
+				today.setHours(23, 59, 59, 999)
```
juliushaertl, line 398: I'm not sure about this, having a card due at 23:59 is probably not the best attempt.

How about adding a date picker and setting a default due date for in an hour there?

src/components/Controls.vue
```diff
@@ -314,11 +383,59 @@ export default {
...
+				await this.$store.dispatch('addCard', {
```
juliushaertl, line 399: addCard should use the data that is returned by the request which creates the card, otherwise we miss a lot of required fields

src/store/main.js
```diff
@@ -62,6 +62,8 @@ export default new Vuex.Store({
 		navShown: true,
 		compactMode: localStorage.getItem('deck.compactMode') === 'true',
 		cardDetailsInModal: localStorage.getItem('deck.cardDetailsInModal') === 'true',
+		lastBoardId: localStorage.getItem('deck.lastBoardId'),
```
juliushaertl, line 65: Do we really need a global state for this? This could be easily stored in the modal component itself.

juliushaertl (changes requested):
See inline

src/components/Controls.vue
jakobroehrl: Most of Todo Tools use Today as default. 
Regarding https://github.com/nextcloud/deck/issues/2231#issuecomment-679945453 Today is the current date and 23:59

jancborchardt:
Very nice! :) Some design feedback:
- If it’s adding a card for today, shouldn’t we put it next to the "Today" column heading again?
- Better wording for the modal heading would be "Add card for today" (not "on" and "today" in lowercase so it’s a sentence)
- What’s "b1" and "liste" in this case – I assume boards? In that case I would say it should go in a different element, e.g. a dropdown with the default board selected, and also the board color reflected in a circle on the left of the name (like in Calendar). (And if there’s only one board, that dropdown is not needed.)
- I know it says "Card name" elsewhere in the app, but maybe we can improve that wording? Maybe "Card title …" – but possibly a separate discussion. :)
- The "Cancel" button should be on the left side like in other dialogs, right?

src/components/Controls.vue
roboticsalign: In #1157 we are discussing the different purpose of a due date at a specific time/timezone and a due date at a day. 

Regarding Desk and the current issues 23:59 seems to be a quick win for me, that can be imoroved later.

jakobroehrl:
closing in favour of https://github.com/nextcloud/deck/pull/2854

Issue #713, opened 2018-11-17
Ability to choose to be notified per board


putt1ck:
If you are involved with lots of projects or sub-projects, being notified of changes for every board can become a distraction. If you could choose to be notified only on a per board basis, you could decide which boards were a priority for you and not receive updates from every board.

Bonus addition: board admin to have option to set for all users (useful for boards tracking reported issues on a new release or rollout)

Alternative approach: have 2 notification types for Deck, general and "mention", where mention would include being @'ed in a card or being added/removed from a card or board. This might be a simple workaround i.e. you tell people if they want you to look at something they need to mention you

jakobroehrl:
Does it make sense to favourite single cards and get notifications only for them?

putt1ck:
Only if people who are assigned to a card don't get notifications. Although you could treat favouriting a card similar to watching a issue on github; you're not assigned (not directly involved) but want to know how it is progressing

juliushaertl:
Let's keep this simple for now. Most of the use cases where you want to be notified about cards should be covered with the board wide setting I'd say. cc @jancborchardt

jancborchardt:
Agree with @juliushaertl. For the board-wide setting, we can do it exactly the same as Talk and have the actions menu with "Notifications":
- All changes
- @-mentions only
- Off

![image](https://user-images.githubusercontent.com/925062/76601535-b73e2300-653b-11ea-9fbd-e7d8af99b07a.png)
(Header alignment is off for some reason, ignore that.)

juliushaertl:
For deck we would have basically these options:
- Card notifications
  - All cards
  - Assigned cards
  - Off

jakobroehrl:
@juliushaertl 
Could you give me a rough sketch how to start here?

PR #2430, opened 2020-10-12
Due date notification setting per board

juliushaertl:
* Resolves: #713
* Target version: master 

### Summary

- Implement a notification setting per board that allows to adjust if due date notifications should be sent
  - for all cards
  - only for assigned cards **default**
  - never

The user interface was slightly adjusted since having all those settings in the main menu feels to clutter a lot already, so [in contrast to Talk where the options are shown directly](https://github.com/nextcloud/deck/issues/713#issuecomment-598598598) this is now hidden in a submenu:

image

image

### TODO

- [x] Proper icons instead of radio buttons
- [x] Update API documentation for new config endpoints and extended board result
- [x] Check with @jancborchardt if we should make "all cards" the default for boards that are not shared yet
- [x] Cover assignement notification handling with tests

### Checklist


src/components/navigation/AppNavigationBoard.vue
```diff
@@ -34,39 +34,65 @@
...
+				{{ t('deck', 'Board details') }}
+			</ActionButton>
+			<ActionButton v-if="canManage && !board.archived && !showDueSettings"
 				icon="icon-rename"
 				:close-after-click="true"
 				@click="actionEdit">
 				{{ t('deck', 'Edit board') }}
 			</ActionButton>
-			<ActionButton v-if="canManage && !board.archived"
+			<ActionButton v-if="canManage && !board.archived && !showDueSettings"
```
nickvergessen, line 49: This is :see_no_evil:

src/components/navigation/AppNavigationBoard.vue
```diff
@@ -34,39 +34,65 @@
...
 				icon="icon-clone"
 				:close-after-click="true"
 				@click="actionClone">
 				{{ t('deck', 'Clone board ') }}
```
nickvergessen, line 53: btw trailing spaces inside the translation?

src/components/navigation/AppNavigationBoard.vue
```diff
@@ -34,39 +34,65 @@
...
-			<ActionButton v-if="canManage"
+
+			<ActionButton :icon="!showDueSettings ? 'icon-notifications-dark' : 'icon-view-previous' " @click="showDueSettings=!showDueSettings">
+				{{ t('deck', 'Due date reminders') }}
+			</ActionButton>
+			<ActionRadio v-if="showDueSettings"
+				name="notification"
+				:checked="board.settings['notify-due'] === 'all'"
+				@change="updateSetting('notify-due', 'all')">
+				All cards
```
nickvergessen, line 75: not translated :scream:

src/components/navigation/AppNavigationBoard.vue
```diff
@@ -34,39 +34,65 @@
...
+			</ActionRadio>
+			<ActionRadio v-if="showDueSettings"
+				name="notification"
+				:checked="board.settings['notify-due'] === 'assigned'"
+				@change="updateSetting('notify-due', 'assigned')">
+				Assigned cards
```
nickvergessen, line 81: not translated :scream:

src/components/navigation/AppNavigationBoard.vue
```diff
@@ -34,39 +34,65 @@
...
+			</ActionRadio>
+			<ActionRadio v-if="showDueSettings"
+				name="notification"
+				:checked="board.settings['notify-due'] === 'off'"
+				@change="updateSetting('notify-due', 'off')">
+				Off
```
nickvergessen, line 87: not translated :scream: 

This will be weird to translate, make `No notifications` is better?

src/components/navigation/AppNavigationBoard.vue
juliushaertl: :see_no_evil: Will clean up once the sub-menu is somewhat approved by @jancborchardt ;)

nickvergessen (changes requested):
See comments

src/components/navigation/AppNavigationBoard.vue
juliushaertl: Thanks, pushed a fix here to avoid conflicts, even though not very related, since it is a very minor change.

src/components/navigation/AppNavigationBoard.vue
```diff
@@ -82,7 +108,7 @@
 </template>
 
 <script>
-import { AppNavigationIconBullet, AppNavigationCounter, AppNavigationItem, ColorPicker, Actions, ActionButton } from '@nextcloud/vue'
+import { AppNavigationIconBullet, AppNavigationCounter, AppNavigationItem, ColorPicker, Actions, ActionButton, ActionRadio } from '@nextcloud/vue'
```
skjnldsv, line 111: Import components directly to gain some bytes?

juliushaertl: When I checked last time we used so many of them that the combined bundle was actually smaller than including each of them, but i check that in irregular intervals.

ChristophWurst (approved):
Code looks clean

skjnldsv (approved):

juliushaertl:
Quick update on the icons to align with talk:

image

jancborchardt:
Very nice with the submenu! :) Is that a component already which could be used elsewhere?

Generally it’s super nice already, here’s some possibility for enhancement, your call if or when it gets done. :D 
- [ ] For proper feedback: A subline for the "Due date reminders" entry with the chosen setting (like you know from Android and iOS) because otherwise you have to go into the submenu to know the active setting. The icon could also reflect the chosen setting.
- [ ] Nice to have: A quick animation left/right when the submenu gets shown and when you go back.
- [ ] This is more generally about the ActionMenu component (cause this is also the case for Talk), but the hover/focus and active style needs to be adjusted to move from the blue bar & text opacity towards just changing the background. cc @ma12-co

jancborchardt:
> Check with @jancborchardt if we should make "all cards" the default for boards that are not shared yet

Yeah, good question. Cause then would it switch to "Assigned cards" when you share it?

juliushaertl:
> Yeah, good question. Cause then would it switch to "Assigned cards" when you share it? 

Yes, we could of course just hide the assigned setting if a board is not shared and just offer on/off then. Of course this will need some careful handling of adjusting the setting if all shares get removed from a board.

jancborchardt:
> Yes, we could of course just hide the assigned setting if a board is not shared and just offer on/off then. Of course this will need some careful handling of adjusting the setting if all shares get removed from a board.

Right, then for a non-shared board it could just be a checkbox for "Due date notifications" without the need for a submenu.
Then the switch to a submenu with a "different" default setting wouldn’t be so strange, cause it’s more clear there’s a difference.

juliushaertl:
@jancborchardt I'll leave the animation for a later point in time, other comments in scope of the deck app should be addressed

Peek 2020-11-06 13-05

src/components/navigation/AppNavigationBoard.vue
juliushaertl: Refactored the submenu handling with the last commit

stefan-niedermann:
@juliushaertl don't want to nag, but did you think about REST-API documentation (if this feature will also be available in the API)? 😇

juliushaertl:
@stefan-niedermann I already thought about you :heart: 

https://github.com/nextcloud/deck/pull/2430/commits/b18ebe83b6c80e3db54e5eb07838d00b6d49c87c

Issue #2423, opened 2020-10-11
Card dialog UI/UX improvements


marcoambrosini:
- [x] Do not open the dialog automatically upon card creation, only upon click;
- [ ] If the card is not focus, sometimes opening the dialog requires a double click. I think it should open at first click;
- [x] Shoudln't the cursor be a pointer on the cards? 
- [ ] The content of the modal should have some ~ 12px padding on all 4 sides;

UPDATE:
- [x] When compact mode is toggled, a click on any point of the card should open the sidebar / dialog, we shouldn't allow to rename the card by clicking the title.

juliushaertl:
>  If the card is not focus, sometimes opening the dialog requires a double click. I think tt should open at first click;

Never seen that so far, is there anything in the console log?

ma12-co:
This happened in Firefox on windows. I can't reproduce again in gnu Linux. I'll check again when I'm back on win

ma12-co:
Added 1 more remark @juliushaertl

juliushaertl:
Makes sense :+1:

drfuture:
Hi,
why "Do not open the dialog automatically.." ?
before modal dialogs (this is a great new feature) i was able to create a new card, hit enter, enter the details like "tags, user..., description".

Normaly i have a filter to "my" Cards active. If i create a new card i now click on +, add the title, enter, remove filter, search the new card, klick on card, add details, enter.

Would be nice to have a more efficient way back 👍

FabianUntermoser:
Another improvement would be to add the ability to resize the modal window.
In my opinion, the window is way to small when youre working with limited screen real estate.

![image](https://user-images.githubusercontent.com/13196058/107827319-d1edcd00-6d86-11eb-8da2-594dac63c127.png)

Maybe tweaking some media queries to make the modal window bigger in small windows would do the deal here, as resizing might be overkill.

jancborchardt:
Since some of the points in here are fixed already, I will close this issue in favor of https://github.com/nextcloud/deck/issues/2702 which has more specification and mockups for a modal redesign. :)

PR #2435, opened 2020-10-13
Do not open the dialog automatically upon card creation, only upon click

juliushaertl:
Fix modal design issues as proposed in #2423

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -131,13 +131,6 @@ export default {
 			}
 		}
 	}
-</style>
-
-<style lang="scss">
-
-	.multiselect {
-		width: 100%;
-	}
 
 	.modal__card {
 		min-width: 320px;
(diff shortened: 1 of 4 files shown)


ma12-co (commented):
@juliushaertl I wouldn't open the sidebar either upon card creation, what do you think?

ma12-co:
Actually, @juliushaertl, the sidebar should probably never overlap the content, just like in talk

juliushaertl:
> Actually, @juliushaertl, the sidebar should probably never overlap the content, just like in talk

It doesn't, the scroll container is reduced to the remaining width. That is a bit special for deck of course, but there is no way around horizontal scrolling here.

ma12-co:
So this is the problem I'm thinking of here: 

Peek 2020-10-13 10-35

juliushaertl:
Ok, there it might help already if we scroll the cards into view after inserting them with #2230

ma12-co (approved):

PR #2496, opened 2020-11-03
Transfer ownership

### LINKED ISSUE #1306: New Owner for a board

juliushaertl:
Rebased version of https://github.com/nextcloud/deck/pull/1955

closes #1306 

## ToDo:

- [x] Ensure that unit tests pass
- [x] Make sure that psalm static code analysis passes
- [x] Add integration tests for transferring a board with example data to a different user
- [x] Follow up PR: UI
  - [x] Create API endpoint to trigger the transfer logic 
  - [x] Add option to the UI to transfer a board to a different user

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
 	},
 	"require-dev": {
 		"roave/security-advisories": "dev-master",
-		"christophwurst/nextcloud": "^21@dev",
+		"christophwurst/nextcloud": "dev-master",
 		"phpunit/phpunit": "^9",
 		"nextcloud/coding-standard": "^1.0.0",
 		"symfony/event-dispatcher": "^4.0",
--- a/lib/Controller/BoardController.php
+++ b/lib/Controller/BoardController.php
@@ -161,10 +161,6 @@ class BoardController extends ApiController {
 
 	/**
 	 * @NoAdminRequired
-	 * @param $boardId
-	 * @param $owner
-	 * @param $newOwner
-	 * * @return null|void
 	 */
 	public function transferOwner(int $boardId, string $newOwner): DataResponse {
 		if ($this->permissionService->userIsBoardOwner($boardId, $this->userId)) {
--- a/lib/Controller/BoardController.php
+++ b/lib/Controller/BoardController.php
@@ -24,6 +24,7 @@
 namespace OCA\Deck\Controller;
 
 use OCA\Deck\Db\Acl;
+use OCA\Deck\Db\Board;
 use OCA\Deck\Service\BoardService;
 use OCA\Deck\Service\PermissionService;
 use OCP\AppFramework\ApiController;
@@ -152,7 +153,7 @@ class BoardController extends ApiController {
 	/**
 	 * @NoAdminRequired
 	 * @param $boardId
-	 * @return \OCP\Deck\DB\Board
+	 * @return Board
 	 */
 	public function clone($boardId) {
 		return $this->boardService->clone($boardId, $this->userId);
--- a/lib/Controller/BoardController.php
+++ b/lib/Controller/BoardController.php
@@ -27,6 +27,8 @@ use OCA\Deck\Db\Acl;
 use OCA\Deck\Service\BoardService;
 use OCA\Deck\Service\PermissionService;
 use OCP\AppFramework\ApiController;
+use OCP\AppFramework\Http;
+use OCP\AppFramework\Http\DataResponse;
 use OCP\IRequest;
 
 class BoardController extends ApiController {
@@ -163,7 +165,11 @@ class BoardController extends ApiController {
 	 * @param $newOwner
 	 * * @return null|void
 	 */
-	public function transferOwner($boardId, $owner, $newOwner) {
-		return $this->boardService->transferOwnership($owner, $newOwner);
+	public function transferOwner(int $boardId, string $newOwner): DataResponse {
+		if ($this->permissionService->userIsBoardOwner($boardId, $this->userId)) {
+			return new DataResponse($this->boardService->transferBoardOwnership($boardId, $newOwner), HTTP::STATUS_OK);
+		}
+
+		return new DataResponse([], HTTP::STATUS_UNAUTHORIZED);
 	}
 }
--- a/lib/Controller/BoardController.php
+++ b/lib/Controller/BoardController.php
@@ -155,4 +155,15 @@ class BoardController extends ApiController {
 	public function clone($boardId) {
 		return $this->boardService->clone($boardId, $this->userId);
 	}
+
+	/**
+	 * @NoAdminRequired
+	 * @param $boardId
+	 * @param $owner
+	 * @param $newOwner
+	 * * @return null|void
+	 */
+	public function transferOwner($boardId, $owner, $newOwner) {
+		return $this->boardService->transferOwnership($owner, $newOwner);
+	}
 }
--- a/lib/Db/AclMapper.php
+++ b/lib/Db/AclMapper.php
@@ -64,33 +64,33 @@ class AclMapper extends DeckMapper implements IPermissionMapper {
 	 * @return void
 	 */
 	public function transferOwnership($ownerId, $newOwnerId) {
-        $params = [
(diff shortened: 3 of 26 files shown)


stefan-niedermann:
Will there also be some kind of REST API?

juliushaertl:
As of now this is only an occ command for admins. Whenever we expose it as a feature to users we will add an API for the user interface that can be used for the apps as well then.

juliushaertl:
@matchish I took the liberty to rebase and adjust some small issues i found during testing. Would be great if you could have another look if all that makes sense to you.

The major change is that when transfering ACL rules it should be enough to just drop the one for the new owner. The old owner will no longer have access unless there is a new rule added manually afterwards.

estan:
It would be great if some dev could find time to review this. We have employees leaving and naturally want to transfer the ownership of some of their boards to other members of the team. We rather not mess about with running manual SQL statements against our database, even if it is probably safe if we get it right.

max-nextcloud:
I rebased this, will look into the failing tests and review the code changes next.

max-nextcloud (changes requested):

lib/Db/AclMapper.php
```diff
@@ -57,4 +57,25 @@ public function findByParticipant($type, $participant): array {
 		$sql = 'SELECT * from *PREFIX*deck_board_acl WHERE type = ? AND participant = ?';
 		return $this->findEntities($sql, [$type, $participant]);
 	}
+
+	/**
+	 * @param $ownerId
+	 * @param $newOwnerId
+	 * @return void
+	 */
+	public function transferOwnership($boardId, $ownerId, $newOwnerId) {
```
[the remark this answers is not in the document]
juliushaertl: Renamed to deleteParticipantFromBoard and moved to use the query builder.

lib/Db/AssignmentMapper.php
```diff
@@ -146,4 +146,28 @@ private function getOrigin(Assignment $assignment) {
...
+	 * @param $newOwnerId
+	 * @return void
+	 */
+	public function transferOwnership($ownerId, $newOwnerId) {
+		$params = [
+			'newOwner' => $newOwnerId,
+			'type' => AssignedUsers::TYPE_USER
+		];
+		$qb = $this->db->getQueryBuilder();
+		$sql = "DELETE FROM `*PREFIX*{$this->tableName}`  WHERE `participant` = :newOwner AND `type`= :type";
```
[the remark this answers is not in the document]
juliushaertl: I adapted the logic quite a bit here in a fresh function called remapAssignedUser to process all cards of a given board id. Overall this will end up in having multiple queries per transferred board but processing all user boards that way now also allows to give intermediate status update on the transfer during the running operation.

juliushaertl:
@max-nextcloud I'd appreciate another review. I addressed the remaining comments and also added a dedicated test case for the bug you encountered about the too wide scope of changing assigned users and card ownership: https://github.com/nextcloud/deck/pull/2496/files#diff-aafe9743edf16cdb498c7c4f15245d4d7dd70b1567d28a2b329a709cb950aa03R260-R278

juliushaertl:
In addition now the logic of transferring all user boards was changed a bit to just iterate over all boards and transfer them individually, which makes code more sharable and the performance impact of individual queries is ok form my perspective for a one time action of the transfer.

There is also a command line option now if the assigned users and creator of the cards should actually be changed or remain unchanged as well.

lib/Controller/BoardController.php
```diff
@@ -155,4 +155,15 @@ public function deleteAcl($aclId) {
...
+
+	/**
+	 * @NoAdminRequired
+	 * @param $boardId
+	 * @param $owner
+	 * @param $newOwner
+	 * * @return null|void
+	 */
+	public function transferOwner($boardId, $owner, $newOwner) {
+		return $this->boardService->transferOwnership($owner, $newOwner);
```
juliushaertl, line 167:
```suggestion
		return $this->boardService->transferBoardOwnership($boardId, $owner, $newOwner);
```

To make use of the newly added method to only transfer a single board.

lib/Controller/BoardController.php
```diff
@@ -155,4 +155,15 @@ public function deleteAcl($aclId) {
...
 	}
+
+	/**
+	 * @NoAdminRequired
+	 * @param $boardId
+	 * @param $owner
+	 * @param $newOwner
+	 * * @return null|void
+	 */
+	public function transferOwner($boardId, $owner, $newOwner) {
```
juliushaertl, line 166:
```suggestion
	public function transferOwner(int $boardId, string $newOwner): DataResponse {
```

Things to add here:
- a permission check if the current user is actually the owner
- then the owner does not need to be passed as a parameter anymore through the API

lib/Service/BoardService.php
```diff
@@ -746,6 +746,12 @@ private function clearBoardsCache() {
 		$this->boardsCache = null;
 	}
 
+	private function getBoardOwner($boardId) {
```
juliushaertl, line 749: Seems to be unused

src/components/board/SharingTabSidebar.vue
```diff
@@ -53,6 +53,9 @@
 					<ActionCheckbox v-if="canManage" :checked="acl.permissionManage" @change="clickManageAcl(acl)">
 						{{ t('deck', 'Can manage') }}
 					</ActionCheckbox>
+					<ActionCheckbox v-if="canManage" :checked="acl.owner" @change="clickTranserOwner(acl.participant.uid)">
```
juliushaertl, line 56:
```suggestion
					<ActionCheckbox v-if="isCurrentUser(board.owner.uid)" :checked="acl.owner" @change="clickTranserOwner(acl.participant.uid)">
```

juliushaertl (commented):
Few additional comments, but looks nice :)

max-nextcloud (changes requested):
only checked the code so far. Saw other comments by @juliushaertl so leaving it to that for now.

luka-nextcloud:
@max-nextcloud @juliushaertl Please review again.

kabunatu (approved):
Good to go from my side 👍

juliushaertl (commented):
👍 Good to go from my side

max-nextcloud (approved):
Looks good to me. Adding one comment about uncaught exceptions. I have not tried these corner cases myself but i suspect they result in error traces being printed rather than a proper error message - which can be confusing for users.

lib/Controller/BoardController.php
```diff
@@ -150,9 +153,24 @@ public function deleteAcl($aclId) {
...
 	public function clone($boardId) {
 		return $this->boardService->clone($boardId, $this->userId);
 	}
+
+	/**
+	 * @NoAdminRequired
+	 * @param $boardId
+	 * @param $owner
+	 * @param $newOwner
+	 * * @return null|void
```
vinicius73, line 167:
```suggestion
	 * @return DataResponse
```

composer.json
```diff
@@ -14,7 +14,7 @@
 	},
 	"require-dev": {
 		"roave/security-advisories": "dev-master",
-		"christophwurst/nextcloud": "^21@dev",
+		"christophwurst/nextcloud": "dev-master",
```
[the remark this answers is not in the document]
juliushaertl: Yes, we should test master against master of server, so not related to the PR but should be that way and it solved some psalm issues in this PR.

Issue #2580, opened 2020-11-18
Usage of tasklist in description deletes entire description


markuman:
**Describe the bug**
When check off tasks of a markdown task list, the entire description will be deleted.

**To Reproduce**
1. create a task list

```
- [ ] a
- [ ] b
- [ ] c
```

2. check off one task after each other. 
3. you won't reach task "c", because after check off "b", everything is deleted.

It's also deleted in the database.

**Expected behavior**

description won't be deleted and the tasks list is just checked off.

**Screenshots**

![deckbug2](https://user-images.githubusercontent.com/3920157/99552084-e8210280-29bc-11eb-98cb-64f5b0b36a6d.gif)

**Client details:**
 - OS: linux
 - Browser firefox
 - Version 82.0.2
 - Device: desktop
 - Deck 1.2.1
 - Nextcloud 20.0.1.1
 - PHP 7.4.11

LuCu11:
same here, is there an option to get the description back?

markuman:
Afaik no. It's also deleted in the database.

markuman:
The put request sets `description: ""`. So it's related to the frontend/vuejs.

markuman:
https://github.com/nextcloud/deck/blob/master/src/components/card/Description.vue#L178

This is the function where the accident happen...

I'm a lousy js/vuejs hacker :D

markuman:
...but I like to hotfix it

LuCu11:
luckily the description was from an old deck and I could use a DB backup

markuman:
In v1.1.2 it was `this.copiedCard.description.replace` ...now it replaces the content in place.

markuman:
When you're on mariadb >= 10.3, a workaround to prevent dataloss is, simple add system versioning to that table.


```sql
alter table oc_deck_cards add system versioning;
```

https://mariadb.com/kb/en/system-versioned-tables/

You can than query the history and recover lost data

```sql
select title, description, row_start, row_end from oc_deck_cards FOR SYSTEM_TIME BETWEEN (NOW() - INTERVAL 1 YEAR) AND NOW() where title like 'YOUR CARD TITLE%';
```

But that's not a permanent solution. Hopefully @juliushaertl find a fix for that fast.

markuman:
` this.description` is already empty when the function is triggered as far as I can see.

[...]

PR #2583, opened 2020-11-18
fix tasklist update in descrition

markuman:
* Resolves: #2580
* Target version: master 

### Summary

Otherwise the description will be deleted, because it access the wrong property.


markuman:
duplicated with #2584

PR #2584, opened 2020-11-18
Fix updating checkbox state and avoid issues due to duplicate sidebar element

### LINKED ISSUE #2580: Usage of tasklist in description deletes entire description

juliushaertl:
Fixes #2580 

Besides the fix for replacing the proper description this also required some adjustment to make sure that the actual card sidebar was not hidden in the DOM during editing in the modal.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -37,7 +37,7 @@
 			</div>
 		</Modal>
 
-		<router-view v-show="!cardDetailsInModal || !$route.params.cardId" name="sidebar" />
+		<router-view name="sidebar" :visible="!cardDetailsInModal || !$route.params.cardId" />
 	</Content>
 </template>
 
--- a/src/components/Sidebar.vue
+++ b/src/components/Sidebar.vue
@@ -21,12 +21,18 @@
   -->
 
 <template>
-	<router-view name="sidebar" />
+	<router-view v-if="visible" name="sidebar" />
 </template>
 
 <script>
 export default {
 	name: 'Sidebar',
+	props: {
+		visible: {
+			type: Boolean,
+			default: true,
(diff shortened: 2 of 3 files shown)


src/App.vue
```diff
@@ -37,7 +37,7 @@
 			</div>
 		</Modal>
 
-		<router-view v-show="!cardDetailsInModal || !$route.params.cardId" name="sidebar" />
+		<router-view name="sidebar" :visible="!cardDetailsInModal || !$route.params.cardId" />
```
skjnldsv, line 40: Can't use v-if directly on the router-view?

juliushaertl: There seems to be some issue with the nested router views that are used for sub-routes, where the router tried to render the component to a non-existing router-view element then. That is why I went that way for now. But I have that on my list to investigate further, so this is more a temporary workaround.

skjnldsv (approved):

der-flo:
A timely release would be awesome because of the accidental data loss happening with a wrong mouse click.

Issue #2575, opened 2020-11-18
Content of modal view shines through header when scrolling


stefan-niedermann:
### Steps to reproduce
1. Open card details in the modal view
2. Make sure you have some content which is scrollable, e. g. 30 attachments or lots of activity entries
3. Scroll down
4. See the content scrolling

### Expected behaviour
The header area should be opaque.

### Actual behaviour
The modal view has some space at the top, where the scrolled content (e. g. attachments or activity entries) "shine through".

![grafik](https://user-images.githubusercontent.com/4741199/99507828-2c43e100-2984-11eb-9c14-5b7751524c2b.png)

## Server configuration detail

**Operating system:** Linux 4.4.0-148-generic #174~14.04.1-Ubuntu SMP Thu May 9 08:17:37 UTC 2019 x86_64

**Webserver:** Apache (fpm-fcgi)

**Database:** mysql 10.3.25

**PHP version:** 

7.4.12
Modules loaded: Core, date, libxml, openssl, pcre, zlib, bz2, calendar, ctype, hash, filter, ftp, gettext, gmp, SPL, iconv, Reflection, session, standard, SimpleXML, sockets, mbstring, tokenizer, xml, cgi-fcgi, mysqlnd, bcmath, curl, dba, dom, enchant, fileinfo, gd, imagick, imap, intl, json, ldap, exif, mysqli, odbc, PDO, pdo_mysql, PDO_ODBC, pdo_pgsql, pdo_sqlite, pgsql, Phar, posix, pspell, redis, soap, sodium, sqlite3, sysvmsg, sysvsem, sysvshm, tidy, xmlreader, xmlrpc, xmlwriter, xsl, zip, ionCube Loader, Zend OPcache

**Nextcloud version:** 19.0.4 - 19.0.4.2

**Updated from an older Nextcloud/ownCloud or fresh install:** 

**Where did you install Nextcloud from:** unknown

<details><summary>Signing status</summary>

Array
(
    [core] => Array
        (
            [EXTRA_FILE] => Array
                (
                    [.well-known/pki-validation/fileauth.txt] => Array
                        (
                            [expected] => 
                            [current] => ***REMOVED SENSITIVE VALUE***
                        )

                    [.well-known/acme-challenge/.htaccess] => Array
                        (
                            [expected] => 
                            [current] => ***REMOVED SENSITIVE VALUE***
                        )

                )

        )

)

</details>

<details><summary>List of activated apps</summary>

```
Enabled:
[...]

PR #2587, opened 2020-11-19
Move modal top spacing to the header to avoid side-effect when scrolling

juliushaertl:
Fixes https://github.com/nextcloud/deck/issues/2575


stefan-niedermann (approved):
Works :+1:

Issue #564, opened 2018-08-01
Nextcloud Files integration


schiessle:
Would be nice if I could not only upload new files as attachment but also select files which are already stored in my Nextcloud to connect then with a specific card.

## Updated specification:

* The user is able to upload a file as an attachment to a card.
* Files uploaded from local storage will be stored in a default folder, pre-defined by the user.
* The uploaded file will be shared with all users who have access to the board.
* If the owner of the uploaded file deletes it, it will be removed from the board and from all users storage.
* The user is able to select an already existing file from Nextcloud Files and attach it to a card.

## Mockups

Being able to attach files to a card, either by choosing an existing file from the Nextcloud files app or by uploading a new one.

![image](https://user-images.githubusercontent.com/1589737/75150914-c0d72680-5705-11ea-8d9f-88a936abe9b2.png)
The sidebar of a card will have an "Attachments" tab which will enable the user to upload files from the local storage and attach it to a card or to attach a file which was already uploaded to Nextcloud

![image](https://user-images.githubusercontent.com/1589737/75150919-c6cd0780-5705-11ea-8d9c-291935c4c544.png)
A file picker will enable the user to attach a file which already exists in Nextcloud, to a card 

![image](https://user-images.githubusercontent.com/1589737/75150931-cdf41580-5705-11ea-9997-936a119ca91a.png)
The user will be able to jump to the file directly ("Show in files")

juliushaertl:
Yes, definitely. There are some things that need to be discussed. It doesn't seem to be easily possible with the current share API the server provides, therefore i opened https://github.com/nextcloud/server/issues/10292. 

- Files should somehow be accessible by all users who have access to the board
  - Should they just show up in the home directory or in a deck specific subfolder
- Files need to be editable by all users who have write access to the board
  - How would editing with collabora/other apps be achieved?
- In case there is public board sharing (#14), how is the file being made accessible to guest users

schiessle:
True, didn't thought about shared boards. Maybe you could use the existing share api to automatically share the file/folder with all users to who the board is shared.

shashankmc:
Started using deck a couple of days ago, and this feature is definitely useful. Hoping to see this feature! :)

soerpoers:
yeah – should be possible. We want to use it in our team instead of trello, but this feature is missing. yes we know, we can go to the file and copy the share link but it's not pretty elegant. Is it possible to show this window of copy the link to share it?
i would help but haven't any skills in programming with API's :(

[...]

PR #2638, opened 2020-12-08
Sharing files to cards

juliushaertl:
Implements #564 

Requires https://github.com/nextcloud/server/pull/24605

- [x] Collaboration search provider (might be replaced with a modal to pick the card) 
  - [x] Card picker for file sharing (requires https://github.com/nextcloud/server/pull/24864)
  - [x] Drop collaboration provider for frontend search plugin
- [x] Share provider
- [x] Keep compatibility to API endpoints
  - [x] List attachments
  - [x] Create attachment
  - [x] Update attachment
  - [x] Delete attachment

### Frontend
- [x] File picker
- [x] Upload files

## Follow up polishing

- [ ] Configuation for target directory
- [ ] Propertly handle existing attachments
  - [ ] Add migration background job
  - [ ] Drop with occ command to migrate in a later version
- [ ] Conflict handling doesn't work with sharing since a file with the same name might be already present in the deck folder
- [ ] Upload could use chunking through webdav, but requires more work (probably something for a follow up)


stefan-niedermann:
> Propertly handle existing attachments

Have you already decided whether or not all existing attachments will be completely migrated to files?

This is important to know for us because in the case we can safely assume that each attachment *has* a file Id which we can use for thumbnail generation etc.

If existing attachments will not be migrated, we would need to check and we were only able to use thumbnails for newly added images... not so nice 😉

juliushaertl:
No final decision taken on that yet, but i think it would make most sense to just migrate all the attachments to the files app. The only concern about this on the server side is that we cannot do this during upgrade as depending on the number of files that are around, this might cause a quite long downtime, so we'd need to to this afterwards in a background job. However that would of course imply that there is a migration period, where old attachments are still not migrated.

In any case I'll make the API response different in terms that the `type` property of attachments will no longer be `deck_file` but `file` for those who are in the files app, meaning that you could for now only fetch previews for those. Does this make sense?

stefan-niedermann:
> In any case I'll make the API response different in terms that the type property of attachments will no longer be deck_file but file for those who are in the files app, meaning that you could for now only fetch previews for those. Does this make sense?

Not really IMHO because we get the information implicitely with the `fileId` property, no? If it ain't there, it is a "old" deck file.

I am concerning that the proposed way will create a historical dept for infinite time because there is no deterministic "end date" for this intermediate state where a file can or can not be from Deck.

I understand the issue of a long running migration. Maybe your proposal can be an intermediate solution (with a recommendation to migrate the files via the cron job or so), and define a fix version (e.g. 2.5) where the files *should* be migrated for most users and all users who are left at this point will be migrated hard while updating the Deck app?

juliushaertl:
> Not really IMHO because we get the information implicitely with the fileId property, no? If it ain't there, it is a "old" deck file.

The difference there is that the API was initially designed to also support other types of "attachments", the usecase i had in mind back then was to for example allow attaching a github ticket to a card. That is why I'd still take the type as a criteria for separation.

> I am concerning that the proposed way will create a historical dept for infinite time because there is no deterministic "end date" for this intermediate state where a file can or can not be from Deck.

> I understand the issue of a long running migration. Maybe your proposal can be an intermediate solution (with a recommendation to migrate the files via the cron job or so), and define a fix version (e.g. 2.5) where the files should be migrated for most users and all users who are left at this point will be migrated hard while updating the Deck app?

True, I think we can make this a fixed date migration where we then just notify the admin in case the attachments are still not migrated at that time, just ignore those in the API then and offer an occ command for manual migration. Then you'd no longer need to care about those. Since we are heading to 1.3 now I'd say just deprecating the old attachments now and dropping in a 2.0 release once it approaches would be fine, we can of course always do this earlier or in 1.5 if no major bump comes up.

stefan-niedermann:
Thank you for reevaluating this, sounds like a good plan 🙂

jakobroehrl:
I did some testing:
- [ ] Attachment counter in card is missing
- [ ] Upload new files is showing "hi there"
- [ ] Linked attachment in description is not shown
- [ ] Show in files is not working: "http://localhost/index.php/f/162". On my setup it should be: "http://localhost/server21/index.php/f/162"

juliushaertl:
>  Moving the Attachments into the first Tab over the description should help a lot:

This is a different topic, please open a new ticket for this.

juliushaertl:
Let's get this in so we can get a first beta out.

Issue #2640, opened 2020-12-09
Markdown URLs in markdown lists cause the checkbox to become a hyperlink and go to the different web page address when clicked


poikilos:
version 1.2.2

A captioned markdown URL eats into the preceding checkbox.

**To Reproduce**
1. Make a list of markdown URLs with display text such as:
```
- [ ] U6 [How online presentations work](https://money.howstuffworks.com/business-communications/how-online-presentations-work1.htm) 
- [x] U6 [Remember these 6 facial expression tips when presenting](https://www.duarte.com/presentation-skills-resources/facial-expressions-matter-when-presenting-heres-why/)
- [ ] U6 [Presentation rubrics](https://www.uwstout.edu/academics/online-distance-education/online-professional-development/educational-resources-rubrics/creating-and-using-rubrics-assessment#presentationrubrics)
```

2. The checkbox becomes a URL (The square itself is highlighted with the URL color. Clicking it checks it, but also goes to the webpage)

**Expected behavior**
Exclude the checkbox from the URL if the markdown brackets exclude it.

**Screenshots**
![kanban - link doesn't terminate - live](https://user-images.githubusercontent.com/7557867/101619413-3dab6680-39e1-11eb-9560-ca23cea1291a.png)

**Client details:**
 - OS: Windows 10
 - Browser Firefox
 - Version 83.0
 - Device: desktop

<details>
<summary>Server details</summary>
**Operating system**:
Ubuntu 20.04

**Web server:**
Nginx

**Database:**
PostgreSQL

**PHP version:**
7.3

**Nextcloud version:** (see Nextcloud admin page)
19.0.5

**Where did you install Nextcloud from:**
tarball (then upgraded via GUI several times)

**Signing status:**

```
No errors have been found.
```

**List of activated apps:**
N/A
```
If you have access to your command line run e.g.:
sudo -u www-data php occ app:list
from within your Nextcloud installation folder
```

**Nextcloud configuration:**
N/A
```
If you have access to your command line run e.g.:
sudo -u www-data php occ config:list system
from within your Nextcloud installation folder

or

Insert your config.php content here
Make sure to remove all sensitive content such as passwords. (e.g. database password, passwordsalt, secret, smtp password, …)
```

**Are you using an external user-backend, if yes which one:** LDAP/ActiveDirectory/Webdav/...
No
</details>

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
N/A
```
Insert your Nextcloud log here
[...]

PR #2682, opened 2020-12-29
Use vue-richtext for markdown rendering

juliushaertl:
It already contains a markdown renderer based on remark and we can actually reuse that and avoid shipping markdownit which is actually quite huge.

Fixes https://github.com/nextcloud/deck/issues/2640

## ToDo

- [ ] Checkboxes are disabled
- [ ] Make sure to only match toggleable checkboxes

### DIFF
--- a/src/components/card/Description.vue
+++ b/src/components/card/Description.vue
@@ -45,10 +45,14 @@
 			</Actions>
 		</h5>
 
-		<div v-if="!descriptionEditing && hasDescription"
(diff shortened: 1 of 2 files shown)


oneWaveAdrian (commented):
Looks good to me.

juliushaertl:
Closing this since were most likely aiming directly for Nextcloud Text integration now with the work started on a npm library for exposing its editor capabilities.

PR #2683, opened 2020-12-30
Handle clicks on calendar entries

juliushaertl:
Requires https://github.com/nextcloud/calendar/pull/2478

### DIFF
--- a/src/init-calendar.js
+++ b/src/init-calendar.js
@@ -1,5 +1,4 @@
-<?php
-/**
+/*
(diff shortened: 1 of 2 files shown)


raboof:
This is awesome, thanks so much!

When I click a task on the calendar I still briefly see the "Ask an admin to install the Tasks app" popup in the top-right, but that is quickly replaced by the relevant board. Not sure if that's an issue in the calendar app or in Deck though.

Issue #2607, opened 2020-11-23
Add active board to the move card dialog


PVTGoesen:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Describe the solution you'd like**
Please add the active board to the move card dialog, to move the card in the active board. In my opinion in bigger boards it is easier to move a card this way.

This should be possible by editing _src/components/cards/CardMenu.vue_  row 128-132
Current:
```js
		boards() {
			return this.$store.getters.boards.filter(board => {
				return board.id !== this.currentBoard.id
			})
		},
```
Solution:
```js
		boards() {
			return this.$store.getters.boards.filter(board => {
				return board.id
			})
		},
```

juliushaertl:
Makes sense. The computed property can then even be simplified more to not filter at all.

PR #2685, opened 2020-12-30
show all boards in move card dialog

jakobroehrl:

* Resolves: https://github.com/nextcloud/deck/issues/2607
* Target version: master

### DIFF
--- a/src/components/cards/CardMenu.vue
+++ b/src/components/cards/CardMenu.vue
@@ -107,6 +107,7 @@ export default {
 	computed: {
 		...mapGetters([
 			'isArchived',
+			'boards',
 		]),
 		...mapState({
 			showArchived: state => state.showArchived,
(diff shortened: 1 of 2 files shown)


src/components/cards/CardMenu.vue
```diff
@@ -127,7 +127,7 @@ export default {
 		},
 		boards() {
 			return this.$store.getters.boards.filter(board => {
-				return board.id !== this.currentBoard.id
+				return board.id
```
juliushaertl, line 130: No need to filter here at all then

src/components/cards/CardMenu.vue
```diff
@@ -126,9 +126,7 @@ export default {
 			return true
 		},
 		boards() {
-			return this.$store.getters.boards.filter(board => {
-				return board.id !== this.currentBoard.id
-			})
+			return this.$store.getters.boards
```
juliushaertl, line 129: Sorry missed that in the first review, we can actually just add it to use mapGetters in line 108

juliushaertl (approved):

PR #2684, opened 2020-12-30
date shortcut

jakobroehrl:
Signed-off-by: Jakob Röhrl <jakob.roehrl@web.de>
* Target version: master 
grafik

nextcloud/nextcloud-vue#1419

PR #2710, opened 2021-01-07
Due date shortcuts in the datepicker

jakobroehrl:
* Target version: master 
https://github.com/nextcloud/deck/pull/2684

grafik

### DIFF
--- a/src/components/card/CardSidebarTabDetails.vue
+++ b/src/components/card/CardSidebarTabDetails.vue
@@ -179,17 +179,43 @@ export default {
 			},
 			shortcuts: [
 				{
-					text: 'Today',
+					text: t('deck', 'Today'),
 					onClick() {
 						const date = new Date()
+						date.setDate(date.getDate())
+						date.setHours(23)
+						date.setMinutes(59)
+						return date
+						// new Date(today.getFullYear(), today.getMonth(), today.getDate(), 23, 59, 59)
+					},
+				},
+				{
+					text: t('deck', 'Tomorrow'),
+					onClick() {
+						const date = new Date()
+						date.setDate(date.getDate() + 1)
+						date.setHours(23)
+						date.setMinutes(59)
 						return date
 					},
 				},
 				{
-					text: 'Tomorrow',
+					text: t('deck', 'Next week'),
 					onClick() {
 						const date = new Date()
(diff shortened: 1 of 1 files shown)


jakobroehrl:
@juliushaertl Could we release this with 1.4?

jakobroehrl:
This is working fine. 
Could we merge it?

skjnldsv (approved):
Code looks good

src/components/card/CardSidebarTabDetails.vue
```diff
@@ -176,6 +177,48 @@ export default {
...
+						return date
+					},
+				},
+				{
+					text: t('deck', 'Tomorrow'),
+					onClick() {
+						const date = new Date()
+						date.setDate(date.getDate() + 1)
+						date.setHours(23)
+						date.setMinutes(59)
```
juliushaertl, line 197: I think we should rather use the current time for the entries that are not due today.

Using 23:59 for today is also not ideal but I guess it will be hard to find a good other default time.

juliushaertl: cc @nextcloud/designers on this one. Any good idea if the user picks a day how we should determine the best time to set then for a due date reminder?

jakobroehrl:
@raimund-schluessler 
What are your thoughts on this?
I would like to bring this into tasks, too. If we have an consense here.

src/components/card/CardSidebarTabDetails.vue
jancborchardt: @juliushaertl @jakobroehrl as usual: What do others like e.g. Trello do? :) One could argue for any of various solutions:
- Use 23:59 (end of day)
- Use the current time (or e.g. closest half hour to current time)
- Use 12:00 midday or some other default
- Remember whatever was used last

But it’s always best to check what another widely used solution does, because that’s what people are used to. :)

jakobroehrl: Todo Tools like todoist or ticktick have All Day due dates. So I use 23:59 for today.
Trello uses the now due date. 
Meistertask uses 12:00 for the current date.

I would use 23:59. :)

juliushaertl (approved):

PR #2791, opened 2021-02-08
Open description links in new tab

### LINKED ISSUE #1975: Open description links in new tab
Especially when there are multible references or links in the description of a card, it's not good to lose the board when you click on a link. So why not open the website in a new tab?

fm-sys:
* Resolves #1975
* Target version: master 

### Summary
With this change, clicking on links in the description will open them in a new tab...

### Checklist

- [x] ~Tests (unit, integration, api and/or acceptance) are included~ -> not necessary

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
 		"lodash": "^4.17.21",
 		"markdown-it": "^12.2.0",
 		"markdown-it-task-lists": "^2.1.1",
+		"markdown-it-link-attributes": "^3.0.0",
 		"moment": "^2.29.1",
 		"nextcloud-vue-collections": "^0.9.0",
 		"p-queue": "^6.6.2",
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,6 +27,7 @@
 				"lodash": "^4.17.21",
 				"markdown-it": "^12.2.0",
 				"markdown-it-task-lists": "^2.1.1",
+				"markdown-it-link-attributes": "^3.0.0",
 				"moment": "^2.29.1",
 				"nextcloud-vue-collections": "^0.9.0",
 				"p-queue": "^6.6.2",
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -47,6 +47,7 @@
 		"lodash": "^4.17.20",
 		"markdown-it": "^12.0.4",
 		"markdown-it-task-lists": "^2.1.1",
+		"markdown-it-for-inline": "^0.1.1",
```
juliushaertl, line 50: From this plugins readme:

> This plugin allows to apply function to certain types of inline tokens. Speed will be not fastest of possible, but you can do quick prototyping of certain rule types.

I'd rather pick https://www.npmjs.com/package/markdown-it-link-attributes for this

juliushaertl (changes requested):
See inline

package.json
q-wertz: Not an expert at all, just looked into it because I would also like to have that feature. Would
https://github.com/markdown-it/markdown-it/blob/master/docs/architecture.md#renderer
also be a possibility without importing another library?

fm-sys:
@juliushaertl would you mind another review?

juliushaertl:
Thanks a lot, i took the liberty to fix the linting to use tabs instead of spaces and add the `noreferrer` option as we use that in Nextcloud for all external links.

fm-sys:
> 
> 
> Thanks a lot, i took the liberty to fix the linting to use tabs instead of spaces and add the `noreferrer` option as we use that in Nextcloud for all external links.

Oh, sorry. Had done the same fix and forced pushed as well. Will try to revert the force push...

juliushaertl:
Ah, I just wondered about that. I'll let you take care then :)

fm-sys:
@juliushaertl Force push reverted to commit  6335f50

juliushaertl (approved):

PR #2822, opened 2021-02-23
Also include /apps/spreed urls in the listener for loading the talk integration

juliushaertl:
Otherwise the deck scripts would only be loaded when a /call/... url is accessed


PVince81 (approved):
makes sense :+1:

Issue #2494, opened 2020-11-03
"Create new card/board" for the Dashboard widget


splitt3r:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Describe the bug**
There should be a "Create new card/board" button in the dashboard widget. I think it is not as easy as the calendar one because there is no route for the new form yet + you can delete everything in Deck but the calendar app has at least a personal calendar all the time. So there needs to be some logic to check if new card / board makes sense.

**Screenshots**
![image](https://user-images.githubusercontent.com/1869584/97908683-4dcb8880-1d47-11eb-8a50-b1a813892ed9.png)

_Originally posted by @splitt3r in https://github.com/nextcloud/deck/issues/2492#issuecomment-720672725_

juliushaertl:
Yes, I'd say a modal with basic inputs for board/stack selection and card title would be a good starting point here.

PR #2853, opened 2021-03-04
nice dashboard add card button

jakobroehrl:

* Resolves: #2494
* Target version: master 

image
image

### DIFF
--- a/src/views/Dashboard.vue
+++ b/src/views/Dashboard.vue
@@ -21,32 +21,40 @@
   -->
 
 <template>
-	<DashboardWidget :items="cards"
-		empty-content-icon="icon-deck"
-		:empty-content-message="t('deck', 'No upcoming cards')"
-		:show-more-text="t('deck', 'upcoming cards')"
-		:show-more-url="showMoreUrl"
-		:loading="loading"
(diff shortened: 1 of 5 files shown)


src/views/Dashboard.vue
```diff
@@ -21,32 +21,40 @@
...
+					<ul v-if="item.labels && item.labels.length"
+						class="labels">
+						<li v-for="label in item.labels" :key="label.id" :style="labelStyle(label)">
+							<span>{{ label.title }}</span>
+						</li>
+					</ul>
+				</a>
+			</template>
+		</DashboardWidget>
+		<div class="center-button">
```
juliushaertl, line 51: The issue with this is that the DashboardWidget is supposed to be always the same height, which is 7 items so with 7 showing this will exceed the size that is is supposed to have.

I remember we had an issue in server about allowing to add a create button to the top header, but I cannot find it at the moment. 

So for a possible solution in the meantime I think we should rather limit the the listed cards then and manually add a show more action and the button as you do it right now.

https://github.com/nextcloud/nextcloud-vue-dashboard/blob/master/src/DashboardWidget.vue#L70-L76

Issue #2856, opened 2021-03-05
cannot append files to cards [nextcloud 21]


t1nux:
**Describe the bug**
Attaching a file to a card does not work.

**To Reproduce**
1. select a card
2. click "Upload new files"
3. select file and click ok
4. no file is uploaded and the following message is shown

`fclose(): supplied resource is not a valid stream resource`

**Expected behavior**
- no error message
- file is actually uploaded

**Screenshots**
![image](https://user-images.githubusercontent.com/4058407/110099172-4a054c80-7da1-11eb-8d46-092f25dbdc14.png)

**log**
```
[...]

PR #2857, opened 2021-03-05
Don't close tempfile as it is already done

juliushaertl:
Should fix https://github.com/nextcloud/deck/issues/2856 though I cannot reproduce the original issue


juliushaertl:
Actually it should already be closed when calling putContent throuh the Node api https://github.com/nextcloud/server/blob/19eb48b8a3597943d5d677d8caf318392613d86b/lib/private/Files/View.php#L683

rullzer (approved):

ChristophWurst (approved):

Issue #1635, opened 2020-03-27
Images for Cards


ttytm:
Images tell more then thousand words. I would love to have the possibility to add an image to my cards that is displayed with the card on the board. It would make it more attractive and easier to get an overview for my users.

Here is a Screenshot how it could look.

![FireShot Capture 132 - trello1 jpg (1999×1087) - www blogdumoderateur com](https://user-images.githubusercontent.com/34311583/77793645-bb555f00-706a-11ea-90ec-af4365a9fdd5.jpg)

best regards,
Toby

juliushaertl:
cc @jakobroehrl Since we talked about this already.

The main concern that is currently blocking such a limitation is, that we don't have any preview generation with our current way of storing the files in appdata. So either we need to wait for #564 or add some additional logic to generate previews for appdata files, but I would definitely prefer the first.

tobealive:
oh i see. Looks like 564 is a big one for some issues. Unfortunately i'm missing the knowledge to offer some good help :/

juliushaertl:
Yes, we plan to tackle #564 after our vue migration (1.0 release) is finished, but no ETA for that yet.

tm9k1:
Hi, sorry for the necroposting, but is this being worked on?

ArtyRivera:
Checking back in on this item -- still a popular request from my team, which is how I found this post. Curious if there's an ETA on this and/or if there's any way I or my team could help to contribute to this?

gri38:
Hello. +1.
Thanks for your great project !

t-maes:
I also think this could be a great enhancement of the card view, in brief, a thumbnail of the first attached file. 
Isn't there a way to put a feature request on it, it seems so obvious?

bueroblanko:
any update here?

Gobytego:
how is the mobile app doing this feature?? maybe use the same implementation? i would also like to see this.

juliushaertl:
There also has been some work started in https://github.com/nextcloud/deck/pull/2858 already, but discontinued due to lack of time. Anyone interested can of course pick this up again.

jszeibert:
I would like to revive this discussion as I really like the CoverImge feature from the mobile app.
Above you can find a draft of a possible/ simpler implementation I did a while ago for my own nextcloud setup and just ported and improved slightly to work with stable26.

My goal with this draft PR would be to collect feedback and bepending on that, maybe improve/ complete my imprementation to a proper PR.

[...]

PR #2858, opened 2021-03-05
cover images

jakobroehrl:
* Resolves: #1635 
shows the 1st image attachment:

image

**Questions/Todo**
- [ ] change of the board setting "show cover images" is only after page upload applied
- [ ] new uploaded image attachments are only after page reload shown
- [ ] show the first two pics? (then I have to shrink the image height to 1/2)

### DIFF
--- a/lib/Controller/BoardController.php
+++ b/lib/Controller/BoardController.php
@@ -73,10 +73,11 @@ class BoardController extends ApiController {
 	 * @param $title
 	 * @param $color
 	 * @param $archived
+	 * @param $coverImages
 	 * @return \OCP\AppFramework\Db\Entity
 	 */
-	public function update($id, $title, $color, $archived) {
-		return $this->boardService->update($id, $title, $color, $archived);
+	public function update($id, $title, $color, $archived, $coverImages) {
+		return $this->boardService->update($id, $title, $color, $archived, $coverImages);
 	}
 
 	/**
--- a/lib/Db/BoardMapper.php
+++ b/lib/Db/BoardMapper.php
@@ -79,18 +79,10 @@ class BoardMapper extends QBMapper implements IPermissionMapper {
 	 * @throws \OCP\AppFramework\Db\MultipleObjectsReturnedException
 	 * @throws DoesNotExistException
 	 */
-	public function find($id, $withLabels = false, $withAcl = false): Board {
-		if (!isset($this->boardCache[$id])) {
-			$qb = $this->db->getQueryBuilder();
-			$qb->select('*')
-				->from('deck_boards')
-				->where($qb->expr()->eq('id', $qb->createNamedParameter($id, IQueryBuilder::PARAM_INT)))
-				->orderBy('id');
-			$this->boardCache[$id] = $this->findEntity($qb);
-		}
-
-		// FIXME is this necessary? it was NOT done with the old mapper
-		// $this->mapOwner($board);
+	public function find($id, $withLabels = false, $withAcl = false) {
+		$sql = 'SELECT id, title, owner, color, archived, deleted_at, last_modified, cover_images FROM `*PREFIX*deck_boards` ' .
+			'WHERE `id` = ?';
+		$board = $this->findEntity($sql, [$id]);
 
 		// Add labels
 		if ($withLabels && $this->boardCache[$id]->getLabels() === null) {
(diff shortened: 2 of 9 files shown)


lib/Migration/Version10400Date20210305.php
```diff
@@ -0,0 +1,27 @@
...
+
+class Version10400Date20210305 extends SimpleMigrationStep {
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		// Add cover image database field
+		$table = $schema->getTable('deck_boards');
+		if (!$table->hasColumn('coverImages')) {
+			$table->addColumn('coverImages', 'boolean', [
```
nickvergessen, line 20: Please never use uppercase in database table names or column names

nickvergessen:
```suggestion
		if (!$table->hasColumn('cover_images')) {
			$table->addColumn('cover_images', 'boolean', [
```

lib/Migration/Version10400Date20210305.php
```diff
@@ -0,0 +1,27 @@
...
+				'notnull' => false,
+				'default' => true,
+			]);
+		}
+		return $schema;
```
nickvergessen, line 25:
```suggestion
			return $schema;
		}
		return null;
```

appinfo/info.xml
```diff
@@ -17,7 +17,7 @@
 - 🚀 Get your project organized
 
 </description>
-    <version>1.4.0-alpha1</version>
```
juliushaertl, line 20: If you need to bump the version for migrations, please only bump the `alpha1` to `alpha2`, the regular version bump will be done during the release process.

lib/Controller/BoardController.php
```diff
@@ -73,10 +73,11 @@ public function create($title, $color) {
 	 * @param $title
 	 * @param $color
 	 * @param $archived
+	 * @param $coverImages
 	 * @return \OCP\AppFramework\Db\Entity
 	 */
-	public function update($id, $title, $color, $archived) {
-		return $this->boardService->update($id, $title, $color, $archived);
+	public function update($id, $title, $color, $archived, $coverImages) {
```
juliushaertl, line 79: We would need to support this in the API Controller as well for the mobile app.

lib/Db/BoardMapper.php
```diff
@@ -66,7 +66,7 @@ public function __construct(
 	 * @throws DoesNotExistException
 	 */
 	public function find($id, $withLabels = false, $withAcl = false) {
-		$sql = 'SELECT id, title, owner, color, archived, deleted_at, last_modified FROM `*PREFIX*deck_boards` ' .
+		$sql = 'SELECT id, title, owner, color, archived, deleted_at, last_modified, cover_images FROM `*PREFIX*deck_boards` ' .
```
juliushaertl, line 69: I'm not entirely sure if we should rather make this a user setting. In any case we should reuse the ConfigService for this instead of adding more fields to the board table to keep it flexible.

stefan-niedermann:
We (Android app) are quite confused at the moment. It was really nice if one could describe the target state.
1. How many images will be shown? 1-3? max. 1? ...?
2. Is there a max-height for the image?
3. What about this setting? I guess it's to turn it off / on at all, not to make the shown image configurable per card by the user?
4. Will or won't this setting be synchronizable? Since @juliushaertl [said it should not go into the boards column](https://github.com/nextcloud/deck/pull/2858#pullrequestreview-637819992), we should talk about [synchronizable per-board settings](https://github.com/stefan-niedermann/nextcloud-deck/issues/745) in the long term

jakobroehrl:
@stefan-niedermann 
I like the new implementation in the Android App. :)
The first 3 pics are shown in the corresponding card and there is a global setting for showing this or not

@juliushaertl: How  we will do it in the WebApp? Start like doing it in the Android App?

juliushaertl:
Open questions i would like to clarify before going any further:
- Will only pictures be used for this or also previews of for example office files?
- In which order will the images be picked? Should there be a manual way to select or will it always be the 3 newest?

> How many images will be shown? 1-3? max. 1? ...?

Trello has a dedicated option for making an image a cover for example which I personally would prefer rather than trying to magically find a picture as that could always lead to seeing an image there which you might not want or might not help to get a better idea on what the card is about.

image


> Is there a max-height for the image?

There should be one, yes.

> What about this setting? I guess it's to turn it off / on at all, not to make the shown image configurable per card by the user?
> Will or won't this setting be synchronizable? Since @juliushaertl said it should not go into the boards column, we should talk about synchronizable per-board settings in the long term

Yes, those settings could still be exposed through the board api endpoints even if they are stored separately. My main concern is that just adding an individual column for each possible setting to the board table doesn't scale very well

stefan-niedermann:
> Will only pictures be used for this or also previews of for example office files?

The Android app started with only pictures because otherwise this will cause confusion why attachments which have been uploaded with Deck `< 1.3.0` are not shown. We will probably change this behavior when the attachment migration is provided.

> In which order will the images be picked? Should there be a manual way to select or will it always be the 3 newest?

In the first draft, the Android app just picks the three first ones in the database. The three most recent sounds better, but i would avoid to make this selectable by the user because this needs a manual user interaction, it will clutter the UI and it will cause maintenance efforts for the options which are probably put into something other more important stuff. But ringing in @nextcloud/designers for this :wink: 

> Yes, those settings could still be exposed through the board api endpoints even if they are stored separately.

Is this really board related? IMHO a global setting for this makes more sense (also a question for @nextcloud/designers)

stefan-niedermann:
Oh, another hint: Think about how cover images work in combination with compact mode.
I put the images on the left side to not increase the height of each item:

Screenshot_20210423-080458_Nextcloud_Deck

jakobroehrl:
@weeman1337 
Could you maybe help me on getting this done?

 - change of the board setting "show cover images" is only after page upload applied
 - new uploaded image attachments are only after page reload shown
 - show the first two pics? (then I have to shrink the image height to 1/2)

weeman1337:
Hi @jakobroehrl . If it is okay for you I would submit some commits.

jakobroehrl:
> Hi @jakobroehrl . If it is okay for you I would submit some commits.

This would be awesome, thanks!

juliushaertl:
When I talked to @jakobroehrl last time he mentioned that he would appreciate if someone could take over the pull request.

So I think the best would be to checkout the changes and push them to your own fork to open a new pull request then.

Let me know if you have any questions, looking forward to your contribution.

juliushaertl:
Let me close this for now. If anyone wants to pick up the work feel free to open a new pull request with it.

Issue #2771, opened 2021-02-01
assign card to user when adding a due date on a shared board


jakobroehrl:
trello does this so
otherwise the card is not shown in the Upcoming card view

stefan-niedermann:
And to which user? I mean if the board is shared, there are multiple possible users...?

jakobroehrl:
To the current user.
If I add a due the card will be assigned to me.

stefan-niedermann:
Hmm, while i understand the motivation to display the cards in the Upcoming cards view, i do vote against this automatic behavior (or at least make it configurable).

A team having a meeting has usually one person who writes down the stuff and with your proposal all cards would be assigned to this person and need changes (which will produce notifications and clutter the activity stream).

Issue #2807, opened 2021-02-15
Improved upcoming cards view


nimisha-vijay:
The current landing page of the Deck app shows only your upcoming cards based on due dates, which means that for people who don't use due dates the view in empty.

To make that better, we could have 2 views: one arranged by due dates like it currently exists, and one arranged by boards. This would allow for more flexibility in the landing page, and can also have some simple operations like adding boards and cards.

Here's a mockup that you can click around in:
<https://www.figma.com/proto/76zsYNIXApcZA9ZqpXdJWb/Nextcloud_Deck_mockup?node-id=37%3A748&scaling=scale-down-width> 

- [ ] You can switch between the 2 views with a toggle, the default view for a user would be their last selected view, or board view if it's their first time

**Board view**

![image](https://user-images.githubusercontent.com/52440189/107946214-b832cc80-6fb6-11eb-8545-cca19535f8fc.png)

![image](https://user-images.githubusercontent.com/52440189/107946194-aea96480-6fb6-11eb-8539-1093c291540a.png)

![image](https://user-images.githubusercontent.com/52440189/107946152-a18c7580-6fb6-11eb-80a1-93272e61923d.png)

- [ ] The board view has a board in each column, arranged by date created, and an add board option at the end. 
- [ ] You can add cards to a board, similar to the current way of adding cards to a list. The name would be entered as a text input, and the list it is in can be selected from a drop down menu of the existing lists in that board.
- [ ] Each board has its cards under it, sorted by due date if due dates are available, then by if that card is assigned to the user, then by last modified
- [ ] The card design would also be slightly different to indicate the name of the list it's in, and the last modified
- [ ] You can also add board from the board view by clicking on the add board button with the same flow as adding one from the sidebar. 
- [ ] New boards can have 3 lists by default: To do, Doing and Done
- [ ] As soon as you add a board from the overview page, you are automatically prompted to add a new card

**Due date view**

![image](https://user-images.githubusercontent.com/52440189/107946264-cbde3300-6fb6-11eb-9fd6-ec5cacc447d7.png)
![image](https://user-images.githubusercontent.com/52440189/107946304-da2c4f00-6fb6-11eb-8b94-87ac763c792f.png)

- [ ] The due date view is similar to what currently exists, each column represents a time frame
- [ ] The cards under each column are grouped by the board they are in. In each group the cards are sorted the same way as the board view: first by due date, then by assigned cards, then by last modified
- [ ] If there are no due dates there is an empty content component

PR #2894, opened 2021-03-17
overview: shared board option: show only assigned cards  or all with due dates

jakobroehrl:

* Resolves:  new idea to solve this: #2771
* Target version: master 

### TODO
- [ ] Get the open working
- [ ] change the sql statement for getting the overview data

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -17,7 +17,7 @@
 - 🚀 Get your project organized
 
 </description>
-    <version>1.4.0-alpha1</version>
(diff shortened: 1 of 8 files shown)


juliushaertl:
I think we should rather focus on implementing the enhancements discussed on https://github.com/nextcloud/deck/issues/2807 than adding another setting. As described in the design proposal we should make sure to rather have a proper sorting (assigned with duedate, assigned, just duedates, none) when showing the cards than adding another setting which needs to be tested separately and maintained.

PR #2933, opened 2021-04-01
Move full text search to proper events

juliushaertl:
Some cleanup in order to use proper events for triggering the full text search indexing and remove the fallback Application class since Deck master/1.3 already only supports Nextcloud versions that offer IBootstrap


lib/Listeners/FullTextSearchEventListener.php
```diff
@@ -0,0 +1,99 @@
...
+					static function (Card $card) {
+						return (string)$card->getId();
+					},
+					$this->service->getCardsFromBoard($acl->getBoardId())
+				);
+				$this->manager->updateIndexesStatus(
+					DeckProvider::DECK_PROVIDER_ID, $cards, IIndex::INDEX_META
+				);
+			}
+		} catch (\Exception $e) {
```
ChristophWurst, line 96: log just in case?

juliushaertl: Yep, makes sense, pushed :)

juliushaertl:
Psalm failure added to the baseline for now will be fixed as part of https://github.com/nextcloud/deck/pull/2934

PR #2935, opened 2021-04-01
Rich object string parameters for notifications

nickvergessen:
Before | After
---|---
Bildschirmfoto von 2021-04-01 16-54-00 | Bildschirmfoto von 2021-04-01 16-57-22

Fixes https://github.com/nextcloud/deck/issues/2949

### DIFF
--- a/lib/Notification/Notifier.php
+++ b/lib/Notification/Notifier.php
@@ -25,6 +25,7 @@ namespace OCA\Deck\Notification;
 
 use OCA\Deck\Db\BoardMapper;
 use OCA\Deck\Db\CardMapper;
+use OCA\Deck\Db\StackMapper;
 use OCP\IURLGenerator;
 use OCP\IUserManager;
 use OCP\L10N\IFactory;
@@ -41,6 +42,8 @@ class Notifier implements INotifier {
 	protected $userManager;
 	/** @var CardMapper */
 	protected $cardMapper;
+	/** @var StackMapper */
+	protected $stackMapper;
 	/** @var BoardMapper */
 	protected $boardMapper;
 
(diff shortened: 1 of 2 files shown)


lib/Notification/Notifier.php
```diff
@@ -177,8 +238,14 @@ public function prepare(INotification $notification, string $languageCode): INot
 					(string) $l->t('The board "%s" has been shared with you by %s.', [$params[0], $dn])
 				);
 				$notification->setRichSubject(
-					(string) $l->t('{user} has shared the board %s with you.', [$params[0]]),
+					$l->t('{user} has shared {deck-board} with you.'),
```
nickvergessen, line 241: not sure if we should keep "… the board …"

lib/Notification/Notifier.php
```diff
@@ -100,6 +105,11 @@ public function prepare(INotification $notification, string $languageCode): INot
 				if (!$boardId) {
 					throw new AlreadyProcessedException();
 				}
+
+				$card = $this->cardMapper->find($cardId);
+				$stackId = $card->getStackId();
+				$stack = $this->stackMapper->find($stackId);
```
nickvergessen, line 111: Is there a shorter way for this?

lib/Notification/Notifier.php
juliushaertl: No strong opinions here, but I'd say the icon is already enough to indicate that it is deck related. Maybe @jancborchardt can comment about the general wording.

lib/Notification/Notifier.php
juliushaertl: Not as of right now as we always use an entity and don't have any way to map joined results to it currently. I'd say we keep it like this and if we encounter more places where this would be useful we can add a helper to do this in one query.

juliushaertl:
Rebased and adjusted the tests

juliushaertl (approved):

nickvergessen:
Sorry, thanks for the tests fixing I forgot about the PR

juliushaertl:
Sure thanks for the PR :+1:

PR #2950, opened 2021-04-07
Remove notification on unshare and add type hints

### LINKED ISSUE #2932: Delete notification on unshare
- Share a board to a group by accident
- Unshare it
- See that the notifications to the group users are still around

juliushaertl:
Remove notifications for board shares that are not dismissed yet when unsharing a board again, so there are no stale notifications left and add some typings to the NotificationHelper class.

Fixes #2932

### DIFF
--- a/lib/Notification/NotificationHelper.php
+++ b/lib/Notification/NotificationHelper.php
@@ -202,6 +202,9 @@ class NotificationHelper {
 				return;
 			}
 			foreach ($group->getUsers() as $user) {
+				if ($user->getUID() === $this->currentUser) {
+					continue;
+				}
 				$notification = $this->generateBoardShared($board, $user->getUID());
 				if ($markAsRead) {
 					$this->notificationManager->markProcessed($notification);
--- a/lib/Notification/NotificationHelper.php
+++ b/lib/Notification/NotificationHelper.php
@@ -1,4 +1,7 @@
 <?php
+
+declare(strict_types=1);
+
 /**
  * @copyright Copyright (c) 2017 Julius Härtl <jus@bitgrid.net>
  *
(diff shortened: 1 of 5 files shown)


lib/Notification/NotificationHelper.php
```diff
@@ -157,27 +169,41 @@ public function sendCardAssigned($card, $userId) {
...
 		}
 		if ($acl->getType() === Acl::PERMISSION_TYPE_GROUP) {
 			$group = $this->groupManager->get($acl->getParticipant());
+			if ($group === null) {
+				return;
+			}
 			foreach ($group->getUsers() as $user) {
 				$notification = $this->generateBoardShared($board, $user->getUID());
-				$this->notificationManager->notify($notification);
+				$notification->setDateTime(new DateTime());
```
nickvergessen, line 196: Should only be set in the `notify()` branch?

juliushaertl: Aye, indeed. Thanks :+1:

lib/Notification/NotificationHelper.php
```diff
@@ -157,27 +169,41 @@ public function sendCardAssigned($card, $userId) {
...
 		if ($acl->getType() === Acl::PERMISSION_TYPE_GROUP) {
 			$group = $this->groupManager->get($acl->getParticipant());
+			if ($group === null) {
+				return;
+			}
 			foreach ($group->getUsers() as $user) {
 				$notification = $this->generateBoardShared($board, $user->getUID());
-				$this->notificationManager->notify($notification);
+				if ($markAsRead) {
+					$this->notificationManager->markProcessed($notification);
```
nickvergessen, line 197: mention, due date and other notifications are dropped when showing them I guess?

nickvergessen (approved):

lib/Notification/NotificationHelper.php
juliushaertl: Went through them as well now and made sure that they are properly removed on actions that require it.

Issue #2739, opened 2021-01-19
Description Field: CSS / JS polish


aktivde:
**Describe the bug**
When editing the description field it changes its width.

**To Reproduce**
Just edit the description.

**Expected behavior**
Size staying the same.

**Screenshots**
![Bildschirmfoto 2021-01-19 um 09 23 24](https://user-images.githubusercontent.com/65334274/105008051-08ab1e80-5a39-11eb-8384-de38d62743ae.png)
![Bildschirmfoto 2021-01-19 um 09 23 11](https://user-images.githubusercontent.com/65334274/105008043-0779f180-5a39-11eb-8019-01347c67522d.png)

**Client details:**
 - OS: Mac OS
 - Safari / Chrome
 - Nextcloud 21 Beta, Nextcloud 1.3 Beta
 - Device: Macbook

stefan-niedermann:
I can't reproduce this issue on `Debian` with `Firefox 78.7.0esr (64-Bit)`.
Can you please be more specific about the version of Safari / Chrome and provide better steps to reproduce? I am missing some key information like

- Does this happen in the sidebar?
- Does this happen in the modal view?
- Did it work once?
- Is it dependent on the amount of characters or the length of the text you are entering?
- Do you already know what it is causing it (since you mentioned css polishing in the issue title)?
- Are you able to reproduce it on a Non-Mac-OS machine (if you have access to one) or with Firefox (since i couldn't reproduce it)?
- ...

aktivde:
> I can't reproduce this issue on `Debian` with `Firefox 78.7.0esr (64-Bit)`.
> Can you please be more specific about the version of Safari / Chrome and provide better steps to reproduce? I am missing some key information like
> 
> * Does this happen in the sidebar?
– Yes it only happens in the sidebar
> * Does this happen in the modal view?
– don't know what that is
> * Did it work once?
– Yes. This is a new bug.
> * Is it dependent on the amount of characters or the length of the text you are entering?
– No, it already happens if you haven't entered anything. But it only happens during the time between writing and saving. After being saved it is displayed correctly.
> * Do you already know what it is causing it (since you mentioned css polishing in the issue title)?
– Yes: It is the "(nicht gespeichert)" that is too long for the div container and leads to shifting the paperclip icon into the next row. Then causing the description field to also shift along that icon
> * Are you able to reproduce it on a Non-Mac-OS machine (if you have access to one) or with Firefox (since i couldn't reproduce it)?
[...]

Issue #2617, opened 2020-11-25
Description edit field is smaller than window


reox:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Describe the bug**

Since the update to 1.2.2, the description text fields in both sidebar and modal view are not full width.
It is only the edit field, as the rendered description is correct.

**To Reproduce**
1. Edit a card
2. Edit description

**Expected behavior**

Field should be 100% width.

(The new border around the textfield is good though, please do not change that back ;))

Does the class `CodeMirror` just require a `width:100%`?

**Screenshots**
![image](https://user-images.githubusercontent.com/431045/100189723-218dcc80-2eed-11eb-99f8-359bd0d066cc.png)

![image](https://user-images.githubusercontent.com/431045/100189757-35393300-2eed-11eb-9b7f-92689842d7dc.png)

**Client details:**
 - OS: W10
 - Browser: Chrome, Firefox
 - Version:  87.0.4280.66, 83.0
 - Device: Desktop

NC 19.0.5 / Deck 1.2.2

reox:
It looks like this got fixed with NC 20.0.4? At least since I upgraded to 20.x, the sizes are equal again.
Please, can someone confirm that? Then this issue can be closed.

juliushaertl:
I could reproduce a similar issue when zooming the webpack today see https://github.com/nextcloud/deck/pull/2983 for a possible fix

PR #2983, opened 2021-04-14
Fix codemirror description width

juliushaertl:
Fixes https://github.com/nextcloud/deck/issues/2739
Fixes https://github.com/nextcloud/deck/issues/2617

@aktivde I could not reproduce the exact same case but a similar one when having the browser zommed in to anything >100%


ChristophWurst (approved):
:100:

PR #3015, opened 2021-04-23
Do not use custom resolve order

juliushaertl:
Fixes a regression caused by #2962 and we should first attempt to use the dependencies package version as it is the default


juliushaertl:
Fixed by https://github.com/nextcloud/deck/pull/3022

PR #3016, opened 2021-04-23
Allow searching for filters without a query to match all that have a given filter set

juliushaertl:
- Allow searching for filters without a query to match all that have a given filter set (e.g. `tag:` will list all cards with a tag assigned)
- Fix handling of quotes due to an off by one error
- Catch canceled tokens and make sure the UI does end up in a stale state
- Show loading indicator when requests are pending where the placeholder might be hidden in the non-visible area

### DIFF
--- a/lib/Search/Query/AQueryParameter.php
+++ b/lib/Search/Query/AQueryParameter.php
@@ -37,7 +37,7 @@ class AQueryParameter {
 	
 	public function getValue() {
 		if (is_string($this->value) && mb_strlen($this->value) > 1) {
-			$param = ($this->value[0] === '"' && $this->value[mb_strlen($this->value) - 1] === '"') ? mb_substr($this->value, 1, -1): $this->value;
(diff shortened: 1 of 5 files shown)


lib/Search/Query/AQueryParameter.php
```diff
@@ -37,7 +37,7 @@ class AQueryParameter {
 	
 	public function getValue() {
 		if (is_string($this->value) && mb_strlen($this->value) > 1) {
-			$param = ($this->value[0] === '"' && $this->value[mb_strlen($this->value) - 1] === '"') ? mb_substr($this->value, 1, -1): $this->value;
+			$param = ($this->value[0] === '"' && $this->value[strlen($this->value) - 1] === '"') ? mb_substr($this->value, 1, -1): $this->value;
```
ChristophWurst, line 40: why do you use the non-mb function just once? should it also be changed on the line before and the same line after?

juliushaertl: Yep that was intentional to get the byte at the offset position, but thinking about it it might be a bit safer to rewrite with using a multibyte substring for those. Pushed a fix and test cases.

ChristophWurst (approved):

PR #3019, opened 2021-04-24
Replace node-sass with sass

skjnldsv:
Fix sass deps


package.json
```diff
@@ -95,9 +95,9 @@
 		"jest": "^26.6.3",
 		"jest-serializer-vue": "^2.0.2",
 		"minimist": "^1.2.5",
-		"node-sass": "^4.14.1",
 		"raw-loader": "^4.0.2",
-		"sass-loader": "^10.1.1",
+		"sass": "^1.32.11",
+		"sass-loader": "^8.0.2",
```
juliushaertl, line 100: Any specific reason for the downgrade to sass-loader 8.x ?

skjnldsv: Nope, that is a mistake!

PR #3025, opened 2021-04-28
Show comment counter and highlight if unread comments are available 

juliushaertl:
- Show counter on comments badge
- Indicate unread comments with a primary color icon

image

Based on #3014

### DIFF
--- a/css/icons.scss
+++ b/css/icons.scss
@@ -22,7 +22,9 @@
 .icon-activity {
 	@include icon-color('activity-dark', 'activity', $color-black);
 }
-
+.icon-comment--unread {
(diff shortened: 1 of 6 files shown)


skjnldsv (approved):

juliushaertl:
Ah tests need adjustments, will fix

ChristophWurst (approved):

stefan-niedermann:
Is this synchronizable? Is there any property which one can read per comment or so? Any documentation about the API?

juliushaertl:
Yes, those would be exposed in the card details reponse as a new field `commentsCount` in addition to `commentsUnread`

Issue #3076, opened 2021-05-17
cards from shared boards do not appear in the due overview


jakobroehrl:
Problem: I work with several teams. We use due cards but no person assignment, so the cards do not appear in the due overview.

Solution 1: https://github.com/nextcloud/deck/issues/2771
Solution 2: https://github.com/nextcloud/deck/pull/2894
Solution 3: If due card is not assigned on shared board, it appears in overview

jakobroehrl:
We have agreed on solution 3
@desperateCoder @stefan-niedermann for Android App.

PR #3078, opened 2021-05-17
show on shared boards unassigned cards to all users

jakobroehrl:

* Resolves: #3076
* Target version: master

### DIFF
--- a/lib/Db/CardMapper.php
+++ b/lib/Db/CardMapper.php
@@ -230,7 +230,7 @@ class CardMapper extends QBMapper implements IPermissionMapper {
 		return $this->findEntities($qb);
 	}
 
-	public function findAssignedCards($boardId, $username) {
+	public function findToMeOrNotAssignedCards($boardId, $username) {
 		$qb = $this->db->getQueryBuilder();
 		$qb->select('c.*')
 			->from('deck_cards', 'c')
--- a/lib/Db/CardMapper.php
+++ b/lib/Db/CardMapper.php
@@ -236,10 +236,12 @@ class CardMapper extends QBMapper implements IPermissionMapper {
 			->from('deck_cards', 'c')
 			->innerJoin('c', 'deck_stacks', 's', 's.id = c.stack_id')
 			->innerJoin('s', 'deck_boards', 'b', 'b.id = s.board_id')
-			->innerJoin('c', 'deck_assigned_users', 'u', 'c.id = u.card_id')
+			->leftJoin('c', 'deck_assigned_users', 'u', 'c.id = u.card_id')
 			->where($qb->expr()->eq('s.board_id', $qb->createNamedParameter($boardId, IQueryBuilder::PARAM_INT)))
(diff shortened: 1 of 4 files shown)


lib/Db/CardMapper.php
```diff
@@ -236,10 +236,12 @@ public function findAssignedCards($boardId, $username) {
...
 			->innerJoin('s', 'deck_boards', 'b', 'b.id = s.board_id')
-			->innerJoin('c', 'deck_assigned_users', 'u', 'c.id = u.card_id')
+			->leftJoin('c', 'deck_assigned_users', 'u', 'c.id = u.card_id')
 			->where($qb->expr()->eq('s.board_id', $qb->createNamedParameter($boardId, IQueryBuilder::PARAM_INT)))
-			->andWhere($qb->expr()->eq('u.participant', $qb->createNamedParameter($username, IQueryBuilder::PARAM_STR)))
-			->andWhere($qb->expr()->eq('u.type', $qb->createNamedParameter(Acl::PERMISSION_TYPE_USER, IQueryBuilder::PARAM_INT)))
+			->andWhere($qb->expr()->orX(
+				$qb->expr()->eq('u.participant', $qb->createNamedParameter($username, IQueryBuilder::PARAM_STR)),
+				$qb->expr()->isNull('u.participant'))
+			)
```
juliushaertl, line 244: As mentioned in the chat, the method should be renamed to also indicate that unassigned cards are fetched as well.

lib/Service/CardService.php
```diff
@@ -615,8 +615,8 @@ public function findAllWithDue($userId) {
 	 * @throws \OCA\Deck\NoPermissionException
 	 * @throws BadRequestException
 	 */
-	public function findAssignedCards($userId) {
-		$cards = $this->cardMapper->findAssignedCards($userId);
+	public function findToMeOrNotAssignedCards($userId) {
+		$cards = $this->cardMapper->findToMeOrNotAssignedCards($userId);
```
juliushaertl, line 619: We actually don't use that method anywhere so we can just drop it for now, it didn't match the previous method signature as well.

juliushaertl: That will also fix the psalm error then

juliushaertl (approved):

stefan-niedermann:
@juliushaertl can you tell me the exact Deck versions where this will be published? If i remember correctly, you are maintaining multiple versions in parallel, right?

juliushaertl:
Yes, currently this is only scheduled for the next release (see milestone). Basically all PRs merged to master will go to the next major. For maintenance releases PRs would need to be backported to the stable1.x branches.

juliushaertl:
@stefan-niedermann I changed plans here a bit since I'll push out a Nextcloud 22 compatible release ans as there are no major changes included we can keep the app compatbile to both 21 and 22. So the next version will be 1.5 including that change while any below that will not have it.

stefan-niedermann:
Thanks for the update!

PR #3090, opened 2021-05-27
[stable1.2] Always fetch attachments

juliushaertl:
Backporting a bugfix that was part of https://github.com/nextcloud/deck/pull/2638 which would also help on the 1.2 release branch to always load the attachment list when switching the sidebar to a different card.

### DIFF
--- a/src/components/card/AttachmentList.vue
+++ b/src/components/card/AttachmentList.vue
@@ -151,8 +151,13 @@ export default {
 			}
 		},
 	},
-	created() {
-		this.$store.dispatch('fetchAttachments', this.cardId)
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):

liejuntao001:
This change break the plugin on Nextcloud 18.0.14. I feel it depends on something missing from Nextcloud 18.
I have to revert to 1.2.7 to bring the system back.

juliushaertl:
Please open a new issue with logs/browser console output as the change doesn't seem to be related to anything that would depend on the Nextcloud version.

PR #3105, opened 2021-06-04
Circles 22

juliushaertl:
Restores compatibility with current circles master branch for Nextcloud 22, but needs some more work before merge as in the end we do not need the typeString anymore at all.

@daita @skjnldsv I assume just showing the displayname for anything related to the circle is fine, right? We used to show the type of the circle (personal/public/...) before next to it but that seems to have been dropped now with 22.

### DIFF
--- a/lib/Db/Circle.php
+++ b/lib/Db/Circle.php
@@ -36,8 +36,8 @@ class Circle extends RelationalObject {
 	public function getObjectSerialization() {
 		return [
 			'uid' => $this->object->getUniqueId(),
-			'displayname' => $this->object->getName(),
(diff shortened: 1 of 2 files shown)


daita:
DisplayName is good, but we should schedule a call about Circles 22 :)

juliushaertl:
Ah wait my additional changes got lost somewhere ...

juliushaertl:
:ping_pong: for review :)

skjnldsv (approved):

ArtificialOwl:
I am sure there is better/faster way to arrange this, but we can keep it this way at the moment !

ArtificialOwl (approved):

PR #3152, opened 2021-06-25
Only offer stack creation in emptycontent with proper permissions

juliushaertl:
- Share an empty board with read only
- Open the board as the recipient

This makes sure the input in the empty content view is properly hidden if the user cannot create lists.


GretaD (approved):
works

PR #3161, opened 2021-07-02
Reduce duplicate queries when fetching user boards an permissions

juliushaertl:
First step to reduce the amounts of queries that happened during collecting incoming shares when setting up the filesystem for users.

Already gives quite some improvement in test setup with 1 shared board having 3 cards with a handful of attachments on each:
https://blackfire.io/profiles/compare/67ec272e-9530-491b-afb2-0fb2c9b06f59/graph

image

### DIFF
--- a/lib/Db/BoardMapper.php
+++ b/lib/Db/BoardMapper.php
@@ -23,6 +23,7 @@
 
 namespace OCA\Deck\Db;
 
+use OC\Cache\CappedMemoryCache;
 use OCP\AppFramework\Db\DoesNotExistException;
 use OCP\IDBConnection;
 use OCP\IUserManager;
@@ -38,6 +39,8 @@ class BoardMapper extends DeckMapper implements IPermissionMapper {
 	private $logger;
 
 	private $circlesEnabled;
+
+	private $userBoardCache;
 
 	public function __construct(
 		IDBConnection $db,
@@ -55,6 +58,9 @@ class BoardMapper extends DeckMapper implements IPermissionMapper {
 		$this->userManager = $userManager;
 		$this->groupManager = $groupManager;
 		$this->logger = $logger;
+
+		$this->userBoardCache = new CappedMemoryCache();
+
 
 		$this->circlesEnabled = \OC::$server->getAppManager()->isEnabledForUser('circles');
 	}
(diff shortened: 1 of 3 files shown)


lib/Db/BoardMapper.php
```diff
@@ -89,13 +93,16 @@ public function find($id, $withLabels = false, $withAcl = false) {
...
 
 	public function findAllForUser(string $userId, int $since = -1, $includeArchived = true): array {
-		$groups = $this->groupManager->getUserGroupIds(
-			$this->userManager->get($userId)
-		);
-		$userBoards = $this->findAllByUser($userId, null, null, $since, $includeArchived);
-		$groupBoards = $this->findAllByGroups($userId, $groups,null, null, $since, $includeArchived);
-		$circleBoards = $this->findAllByCircles($userId, null, null,  $since, $includeArchived);
-		return array_unique(array_merge($userBoards, $groupBoards, $circleBoards));
+		if (!$this->userBoardCache) {
```
nickvergessen, line 96: cache needs to be user based

juliushaertl: Adjusted.

nickvergessen (approved):

lib/Db/BoardMapper.php
```diff
@@ -89,13 +91,20 @@ public function find($id, $withLabels = false, $withAcl = false) {
...
 
 	public function findAllForUser(string $userId, int $since = -1, $includeArchived = true): array {
-		$groups = $this->groupManager->getUserGroupIds(
-			$this->userManager->get($userId)
-		);
-		$userBoards = $this->findAllByUser($userId, null, null, $since, $includeArchived);
-		$groupBoards = $this->findAllByGroups($userId, $groups,null, null, $since, $includeArchived);
-		$circleBoards = $this->findAllByCircles($userId, null, null,  $since, $includeArchived);
-		return array_unique(array_merge($userBoards, $groupBoards, $circleBoards));
+		if (!array_key_exists($userId, $this->userBoardCache)) {
```
blizzz, line 94: `isset()` is said to be faster, but there is a small difference –https://www.php.net/manual/en/function.array-key-exists.php example #2. For consideration.

lib/Db/BoardMapper.php
```diff
@@ -39,6 +39,8 @@ class BoardMapper extends DeckMapper implements IPermissionMapper {
 
 	private $circlesEnabled;
 
+	private $userBoardCache = [];
```
blizzz, line 42: could be also CappedMemoryCache to avoid overflow.

lib/Service/PermissionService.php
```diff
@@ -168,13 +169,20 @@ public function userIsBoardOwner($boardId, $userId = null) {
...
-			$board = $this->boardMapper->find($boardId);
+			$board = $this->getBoard($boardId);
 			return $board && $userId === $board->getOwner();
 		} catch (DoesNotExistException | MultipleObjectsReturnedException $e) {
 		}
 		return false;
 	}
 
+	private function getBoard($boardId): Board {
+		if (!array_key_exists($boardId, $this->boardCache)) {
```
blizzz, line 180: also isset for consideration

lib/Db/BoardMapper.php
```diff
@@ -89,13 +91,20 @@ public function find($id, $withLabels = false, $withAcl = false) {
...
+			$userBoards = $this->findAllByUser($userId, null, null, $since, $includeArchived);
+			$groupBoards = $this->findAllByGroups($userId, $groups, null, null, $since, $includeArchived);
+			$circleBoards = $this->findAllByCircles($userId, null, null, $since, $includeArchived);
+			$allBoards = array_unique(array_merge($userBoards, $groupBoards, $circleBoards));
+			if ($since === -1 && $includeArchived === true) {
+				$this->userBoardCache[$userId] = $allBoards;
+			}
+			return $allBoards;
+		}
+		return $this->userBoardCache;
```
nickvergessen, line 107:
```suggestion
		return $this->userBoardCache[$userId];
```

lib/Db/BoardMapper.php
nickvergessen, line 94:
```suggestion
		if (!array_key_exists($userId, $this->userBoardCache) || $since !== -1 || $includeArchived !== true) {
```

juliushaertl: Fixed slightly adjusted

juliushaertl:
Looking into the tests

nickvergessen (approved):

PR #3177, opened 2021-07-08
Use async import for vue component on collections entrypoint

juliushaertl:
Same as in https://github.com/nextcloud/spreed/pull/5974

Use async loading of the component that get triggered in the files app or the collections as otherwise the imports of dependent components from @nextcloud/vue takes up quite some time when loading the files app initially without triggering the display.

### DIFF
--- a/src/views/FileSharingPicker.js
+++ b/src/views/FileSharingPicker.js
@@ -21,7 +21,6 @@
  */
 
 import Vue from 'vue'
-import CardSelector from '../CardSelector'
 import { createShare } from '../services/SharingApi'
 
 export default {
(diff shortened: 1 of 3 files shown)


src/views/FileSharingPicker.js
```diff
@@ -34,12 +33,12 @@ export default {
 			container.id = 'deck-board-select'
 			const body = document.getElementById('body-user')
 			body.append(container)
-			const SelectorView = Vue.extend(CardSelector)
-			const ComponentVM = new SelectorView({
-				propsData: {
+			const CardSelector = () => import('./../CardSelector')
```
Pytal, line 36:
```suggestion
			const CardSelector = () => import('../CardSelector')
```

Pytal (approved):

PVince81 (approved):
:+1: makes sense

PVince81:
can't merge due to "out-of-date" rule

Issue #3162, opened 2021-07-02
Reduce amount of SQL queries


juliushaertl:
- [x] Board details + ACL fetched multiple times for the same board https://github.com/nextcloud/deck/pull/3161/files
- [x] User boards may be fetched multiple times which could cause more expensive queries through https://github.com/nextcloud/deck/pull/3161/files
- [ ] Mapper implementations might not be performant or run the query to find the related boardId multiple times 

## Still to investigate

  - [ ] Check for all mappers implementing https://github.com/nextcloud/deck/blob/f031717a543cfd041459de93af6a03e2eb829b9c/lib/Db/IPermissionMapper.php#L44
- [ ] Notification preparation might not be optimal https://github.com/nextcloud/deck/blob/master/lib/Notification/Notifier.php#L109-L111
- [ ] `2399 SELECT * FROM oc_deck_stacks WHERE id = ?`

PR #3190, opened 2021-07-13
Reduce duplicate queries when fetching boards for card

diserens:
* Resolves: #3162
* Target version: master 

### Summary

This should prevent duplicate sql queries when fetching board for card.

### Checklist

### DIFF
--- a/lib/Db/CardMapper.php
+++ b/lib/Db/CardMapper.php
@@ -25,6 +25,7 @@ namespace OCA\Deck\Db;
 
 use DateTime;
 use Exception;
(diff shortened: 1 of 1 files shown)


juliushaertl (changes requested):
One small issue is that we would need to reset that cache when we update the stack id in https://github.com/nextcloud/deck/blob/b657967cacc726401fceadf79d3a25c60f8c6ada/lib/Db/CardMapper.php#L81 as otherwise when moving a card to a different board the cache might link to the old board.

Otherwise :+1:

marcelklehr:
This has since landed in master.

Issue #3194, opened 2021-07-15
Activating Deck leads to impossible Login in Nextcloud - internal Server Error


martin-77:
### Steps to reproduce
Did update from Nextcloud 21 -> 22
Deck was preinstalled in 21, worked without problems, had certain Decks running

### Expected behaviour
No Internal Server Error

### Actual behaviour
only with deactivated deck prossible to log in

## Server configuration detail

**Operating system:** Linux 5.11.22-1-pve #1 SMP PVE 5.11.22-2 (Fri, 02 Jul 2021 16:22:45 +0200) x86_64

**Webserver:** nginx/1.20.1 (fpm-fcgi)

**Database:** mysql 10.3.29

**PHP version:** 

7.4.21
Modules loaded: Core, date, libxml, openssl, pcre, zlib, filter, hash, Reflection, SPL, session, standard, sodium, cgi-fcgi, mysqlnd, PDO, xml, apcu, bcmath, bz2, calendar, ctype, curl, dom, mbstring, FFI, fileinfo, ftp, gd, gettext, gmp, iconv, igbinary, imagick, intl, json, ldap, exif, mysqli, pdo_mysql, apc, posix, readline, redis, shmop, SimpleXML, smbclient, sockets, sysvmsg, sysvsem, sysvshm, tokenizer, xmlreader, xmlwriter, xsl, zip, Phar, libsmbclient, Zend OPcache

**Nextcloud version:** 22.0.0 - 22.0.0.11

**Updated from an older Nextcloud/ownCloud or fresh install:** 21

**Where did you install Nextcloud from:** nextcloud.com

<details><summary>Signing status</summary>

Array
(
)

</details>

<details><summary>List of activated apps</summary>

```
Enabled:
 - accessibility: 1.7.0
 - activity: 2.15.0
 - admin_audit: 1.11.0
 - analytics: 3.5.1
 - announcementcenter: 6.0.0-alpha.1
 - apporder: 0.13.0
 - approval: 1.0.3
 - calendar: 2.3.1
 - circles: 22.0.0
 - cloud_federation_api: 1.4.0
 - cms_pico: 1.0.15
 - comments: 1.11.0
 - contacts: 4.0.0
 - contactsinteraction: 1.2.0
 - dashboard: 7.1.0
 - dav: 1.18.0
 - deck: 1.5.0
 - dicomviewer: 1.2.2
 - event_update_notification: 1.3.0
 - external: 3.9.0
[...]

PR #3225, opened 2021-08-03
Check for null value to avoid TypeError in the group manager

juliushaertl:
Fixes https://github.com/nextcloud/deck/issues/3194


Henk57:
I can confirm this fix works.

Cooler3D (approved):
Works for me

Issue #3248, opened 2021-08-11
Server error when accessing Nextcloud Deck on browser "cold" start


reind33r:
Hello!

In a new browser session, when accessing the Nextcloud instance for the first time with the Deck app, there is a server error. If I reload the page, without doing anything else, it shows up normally.

### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Steps to reproduce
1. Open a web browser ("cold" start, but your Nextcloud instance should "remember" you from a previous session)
2. Access /apps/deck

### Expected behaviour
The Deck app should show up normally.

### Actual behaviour
A server error happens. (I reproduced it with Firefox and the new Microsoft Edge, it's not browser-related)

It doesn't happen with other apps, such as dashboard, files or notes.

### Server configuration

**Operating system:** Debian

**Web server:** nginx

**Database:** mariadb  Ver 15.1 Distrib 10.6.3-MariaDB, for debian-linux-gnu (x86_64) using readline 5.2

**PHP version:** PHP 7.4.22

**Nextcloud version:** Hostux Cloud 22.1.0 (but it has been happening through different updates)

**Updated from an older Nextcloud/ownCloud or fresh install:** Updated

**Where did you install Nextcloud from:** Docker

**Signing status:** No errors have been found.

**List of activated apps:**
<details>
<summary>List of activated apps</summary>

```
Enabled:
  - accessibility: 1.7.0
  - activity: 2.15.0
  - bruteforcesettings: 2.2.0
  - calendar: 2.3.1
  - circles: 22.1.0
  - cloud_federation_api: 1.4.0
  - comments: 1.11.0
  - contacts: 4.0.1
  - contactsinteraction: 1.2.0
  - dashboard: 7.1.0
  - dav: 1.18.0
  - deck: 1.5.0
  - drawio: 1.0.1
  - extract: 1.3.2
  - federatedfilesharing: 1.11.0
  - federation: 1.11.0
  - files: 1.16.0
  - files_pdfviewer: 2.3.0
  - files_rightclick: 1.1.0
  - files_sharing: 1.13.2
  - files_trashbin: 1.11.0
  - files_versions: 1.14.0
  - files_videoplayer: 1.11.0
[...]

PR #3260, opened 2021-08-17
Check for userId null value in ConfigService

arrys:
* Resolves: #3248
* Target version: master 

### Summary
This is not really a fix, merely a quick patch to prevent the error page from showing.
It does not fix the underlying issue in any way.

### Checklist


juliushaertl:
Thanks a lot for your pull request. I've started working on a more wide fix for the underlying issue in https://github.com/nextcloud/deck/pull/3263 but will also extend that to apply a null check as you proposed, which definitiely also makes sense as an addition. I'll close this in favor of #3263

Issue #2877, opened 2021-03-10
Using a card to stop sharing an attachment deletes the entire file.


Czujackt:
**Expected behavior** and **Describe the bug**

I understand attaching files actually like this.
You create a card and write a description and comments about it, then you link files. One user links a file and another user finds out that this file was not linked to the appropriate card. So the user unlinks the file. But, he does not want to delete the file!

Another problem is with the deck folder in Nextcloud. When a user deletes the contents of their deck folder, all users automatically have their data deleted and all cards have all attachments deleted as well. So a user can delete all attachments by mistake.

**Client details:**
 - OS: macOS
 - Browser Safari
 - Version 21
 - Device: desktop

Czujackt:
Hi,
why is this a feature request?

If I share a file in a card and I don't want to share it anymore, then the file must not be deleted. The word share alone says that. If I share a file in the Nextcloud with another and then no longer share it, the file is not deleted. Why does Deck behave differently?

This problem is so significant that we can no longer use deck!

stefan-niedermann:
Calm down. I added the label `feature: attachments` because it *affects* the existing features in the `attachments` section.

This is grouping and housekeeping and no reason to get angry. Afaik @juliushaertl is currently on vacation, maybe can @jakobroehrl have a look or you start a PR yourself.

I also think the behavior *was* intentional, so i'd like also to ring the @nextcloud/designers in again here - maybe the idea was to avoid dead attachments which aren't used anymore. Maybe we can have a combined solution: Unshare if the file has been shared otherwise, Delete if the file is not shared / referenced anywhere else.

Czujackt:
Sorry for the harsh choice of words. We have been using Deck for a very long time and have gone through many development steps. But this step is a step backwards and means that we can no longer use Deck the way we do. 

I just think the idea behind it is wrong.
With attachments there are two ways to add files.

The first is 'upload new file'. Here, intuitively, it says upload a file that doesn't exist in Nextcloud to this card and store it in the deck card. If now another user, who can also edit this card, deletes this attachment again, then this attachment must also be deleted. Same right for all. It can't be that the attachment is gone for this user, but still exists for all other users. 

The second is 'Share from file'. Here it says quite clearly. A file from Nextcloud is shared with this card, so it is linked. So it is not the card that becomes the owner of the file. Now if you don't want to link the file anymore, then the file must not be deleted. Since the card did not become the owner of the file, it must not delete it.
[...]

PR #3298, opened 2021-09-06
Delete file shares through attachments API

juliushaertl:
Previously the file was deleted in the file structure of the user is not
expected as the file might not only be related to the card.

Fixes https://github.com/nextcloud/deck/issues/2877

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -959,6 +959,7 @@ For now only `deck_file` is supported as an attachment type.
 
 ### DELETE /boards/{boardId}/stacks/{stackId}/cards/{cardId}/attachments/{attachmentId} - Delete an attachment
 
+    
 #### Request parameters
 
 | Parameter    | Type    | Description                                   |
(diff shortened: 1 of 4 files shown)


eneiluj (commented):
I don't know exactly all that what is changed by this PR.

Did some tests though:
* Before: Removing an attachment (local or directly uploaded) deletes the file
* With this PR, removing an attachment does not delete the file
* Another scenario with this PR:
    * User A share a file F with user B (user share via Files)
    * User A shares a board with user B
    * User A adds file F as attachment to a card, file is shared with the card
    * User B leaves the user share related with file F
    * User B sees file F as attachment of the card but as a "pending share"

So it seems fine to me. Is there any other use case to check?

juliushaertl:
No the first case was what this PR was supposed to cover. Before still had the approach from the old attachments where the file would be deleted which doesn't fit the model with sharing a file from the files app anymore.

Issue #3292, opened 2021-09-04
This directory is not available, please check the log files or contact the administrator.


DasCanard:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Describe the bug**
After Updating Deck to version 1.5.1 i will get this error when i am sharing files
"This directory is not available, please check the log files or contact the administrator."

**To Reproduce**
1. Share a folder
2. Open the URL and see the "This directory is not available, please check the log files or contact the administrator." error

**Screenshots**
![image](https://user-images.githubusercontent.com/17070204/132094254-03afd9ac-2787-4fa5-9ca8-fbdc6c5728f8.png)

**Client details:**
 - OS: Windows 10
 - Browser: Edge
 - Version: 22.1.1
 - Device: Desktop

## Server configuration detail

**Operating system:** Linux 4.15.0-154-generic #161-Ubuntu SMP Fri Jul 30 13:04:17 UTC 2021 x86_64

**Webserver:** nginx/1.21.1 (fpm-fcgi)

**Database:** mysql 10.5.12

**PHP version:** 8.0.10

Modules loaded: Core, date, libxml, openssl, pcre, zlib, filter, hash, json, Reflection, SPL, session, standard, sodium, cgi-fcgi, mysqlnd, PDO, xml, apcu, bcmath, bz2, calendar, ctype, curl, dom, mbstring, FFI, fileinfo, ftp, gd, gettext, gmp, iconv, igbinary, imagick, intl, ldap, exif, mysqli, pdo_mysql, Phar, posix, readline, redis, shmop, SimpleXML, smbclient, sockets, sysvmsg, sysvsem, sysvshm, tokenizer, xmlreader, xmlwriter, xsl, zip, libsmbclient, Zend OPcache

**Nextcloud version:** 22.1.1 - 22.1.1.2

**Updated from an older Nextcloud/ownCloud or fresh install:** 

**Where did you install Nextcloud from:** unknown

<details><summary>Signing status</summary>

Array
(
)

</details>

<details><summary>List of activated apps</summary>

```
Enabled:
 - accessibility: 1.7.0
 - activity: 2.15.0
 - admin_audit: 1.11.0
 - calendar: 2.3.3
[...]

PR #3299, opened 2021-09-06
Return false instead of throwing when getting calendar setting

juliushaertl:
Fixes https://github.com/nextcloud/deck/issues/3292

To reproduce open a public share link and see that it works again with this PR.


eneiluj (approved):
:+1:

eneiluj:
Rebased on master.

Issue #2669, opened 2020-12-19
Improve stack drag and drop behavior


Mr-Tin-25:
## Summary from the issue discussion

A few points that would make the drag behavior feel more expected:

- [ ] Only allow drag and drop on the stack title
- [ ] Do not trigger the stack rename on release of the mouse button if dragging has started

## Original report

Hi,

could you please make columns/lists sticky even for people that are allowed to a add and remove them.
It happens to often to me that i move columns randomly.
Maybe you create an edit mode for columns.

Thanks
Martin

juliushaertl:
Not sure about this, other solutions seem to also always allow reordering. 

cc @jancborchardt @Clementine46 What do you think?

Mr-Tin-25:
and what about moving the columns only by grabbing them on top next to name?

quentinDupont:
> and what about moving the columns only by grabbing them on top next to name?

I agree : what is confusing is that we can move columns by grabbing them anywhere and that the pointer is not changing as when we grab the column on top :

![Peek 21-01-2021 16-01](https://user-images.githubusercontent.com/9005817/105368516-ec0e1280-5c01-11eb-9f0e-cf6f2662f5ff.gif)

quentinDupont:
Maybe a hover css on top column would be good ?

Mr-Tin-25:
Or an icon as handler to grab. Like an arrow left right e.g.

Feature request:
If columns only move when grabbing on top, it would be great to move the whole board view left and right when grabbing background.

juliushaertl:
A few points that would make the drag behavior feel more expected:

- [ ] Only allow drag and drop on the stack title
- [ ] Do not trigger the stack rename on release of the mouse button if dragging has started

juliushaertl:
Related pull request for the cards: https://github.com/nextcloud/deck/pull/3916

PR #3322, opened 2021-09-16
Add drag icon for cards stack

Artem4590:
* Resolves: #2669
* Target version: master 

### Summary

Add drag icon for cards stack

**After fix :**

dnd-l

dnd-d

### Checklist


src/components/board/Stack.vue
```diff
@@ -24,6 +24,14 @@
 <template>
 	<div class="stack">
 		<div v-click-outside="stopCardCreation" class="stack__header" :class="{'stack__header--add': showAddCard }">
+			<Actions v-if="canEdit">
+				<ActionButton
+					id="drag_indicator"
```
juliushaertl, line 29: An id should only be used once, here it would be used multiple times (once for each list)

juliushaertl:
Design-wise I'm not fully convinced about this. At least it seems to be way to present for an action that is probably used less often. We could just show it on hover which would make it still more discoverable than with the current drag and drop. @nextcloud/designers Any further ideas?

Artem4590:
@juliushaertl, @nimisha-vijay, if the behavior is expected I suggest we will close the PR and the corresponding issue (https://github.com/nextcloud/deck/issues/2669).

juliushaertl:
Yes, agreed on that. Let's close this but we should probably find a way to improve the drag behavior otherwise. I'll comment on the issue for that.

Issue #2898, opened 2021-03-17
Confusing button position/orientation


artemanufrij:
Dialog has confusing button position/orientation.

See Screenshot:

![image](https://user-images.githubusercontent.com/16517095/111477011-fbf03180-872e-11eb-9e3f-0c7f9b0ff8b3.png)

Best regards
Artem

artemanufrij:
What do you think about this layout?

![image](https://user-images.githubusercontent.com/16517095/111774716-26b5c380-88b0-11eb-86d9-927e75b258cc.png)

stefan-niedermann:
cc @nextcloud/designers @nimisha-vijay @jancborchardt

ma12-co:
Agree, this doesn't look great.

> What do you think about this layout?

That would make the clickable areas of those buttons intersect.

skjnldsv:
It should be next to it, not weirdly skewed. 
The initial design was both of them next to each other. We don't use stacked-up buttons in Nextcloud, but we put them next to each others :)

Why is it not aligned is the correct question, are you properly defining the background or the compact prop @juliushaertl @jakobroehrl

raimund-schluessler:
@skjnldsv This issue doesn't happen in the sidebar. It is the model that appears if you open a card in the enlarged mode:

![Screenshot_2021-03-19 Deck - Nextcloud](https://user-images.githubusercontent.com/2496460/111785382-d3963d80-88bc-11eb-87f1-0a67dbae61c7.png)

artemanufrij:
@raimund-schluessler , yes it's a screenshot from dialog.

raimund-schluessler:
The cause is this line: https://github.com/nextcloud/deck/blob/master/src/components/card/CardSidebar.vue#L194
The padding-top shifts down the header, but not the close button, leading to the misalignment. If you want to have the header more separated from the upper edge, moving to `margin-top: $modal-padding;` would solve the issue:

![Screenshot_2021-03-19 Deck - Nextcloud(1)](https://user-images.githubusercontent.com/2496460/111786587-448a2500-88be-11eb-8bdd-08f044f5fae7.png)

Clementine46:
> cc @Clementine46 I rememeber that you addressed this in your mockups but I can't find them right now.

I'm not specifically adressing it but my mockups are on the issue Improve modal view design #2702 if needed.

https://github.com/nextcloud/deck/issues/2702#issuecomment-791739112

PR #3324, opened 2021-09-16
Fix menu button position in card modal

valerydmitrieva:
* Resolves: #2898
* Target version: master 

### Summary

Change top indent of menu button in modal card only

#### Before
Screenshot from 2021-09-16 16-21-52

#### After
Screenshot from 2021-09-16 16-21-22

### Checklist


src/components/card/CardSidebar.vue
```diff
@@ -205,6 +205,10 @@ export default {
 				padding-top: $modal-padding;
 				z-index: 100;
 				background-color: var(--color-main-background);
+
+				.app-sidebar-header__menu {
+					top: -7px;
```
juliushaertl, line 210: I think it would be more reasonable to just remove the padding-top here so that the header is aligned at the very top again.

juliushaertl: Basically getting rid of https://github.com/nextcloud/deck/blob/master/src/components/card/CardSidebar.vue#L205 should do the trick here.

juliushaertl (changes requested):
Thanks for your contribution :) I've commented inline with some suggestion on a little bit cleaner approach.

src/components/card/CardSidebar.vue
valerydmitrieva: Ok, I removed `padding-top`.
Speaking about [this](https://github.com/nextcloud/deck/blob/master/src/components/card/CardSidebar.vue#L205), I tried to remove it too, but it doesn't change anything.

juliushaertl: The actual modal width is defined here https://github.com/nextcloud/deck/blob/master/src/App.vue#L152-L156, the rule you linked is just to workaround as we misuse the sidebar component here a bit :wink:

juliushaertl (approved):

Issue #2006, opened 2020-06-09
Group Boards by Project/Client/Team


Quin452:
Hi

Firstly, sorry if this has been addressed already, but I couldn't find any information about it (perhaps I am wording my searches wrong).

Basically, I was wondering if there was a way to group the boards in some way (like you can with Trello - these are called "Teams").
Essentially I have multiple projects on the go, constantly.  Some are my own websites, some are the games I'm working on, and some are with clients, or multiple projects with the same client.

It would be handy for me to quickly switch to my "game development boards" rather than have to scroll to find them.

juliushaertl:
Do you have some example how that is implemented at trello in terms of setting the project and organizing boards in those?

Quin452:
Hi @juliushaertl.  I would recommend having a look [here](https://help.trello.com/article/1177-personal-vs-team-boards), it probably gives the quickest and best explanation :-)

The Teams feature allows a lot more customisation, which I believe you've got implemented by integrating Circles (Board collaboration, etc.)

For a simple system for organisation (which is what I believe is missing/what I'd like), I'd recommend essentially adding a field to the Board, so it can be selected and filtered.
It could be free text, or an additional setting page where you can add groups to populate a dropdown/select box.

I'd honestly would offer to develop it myself, but I don't have the knowledge nor the time right now, and hoping it would be a quick addition.

Dubidubiduu:
Hey @Quin452,

a simple and not so powerful workaround is to use the abc and generate prefixes, since it's ordered alphabetically. So g_ = games, w_ = websites or you order by client name...

Quin452:
Hi @Dubidubiduu thanks for the suggestion.
The only problem I see with that is how can I group and view all the game boards, or website boards?

With Teams, and what I'm suggesting/would like, is the ability to group them together so I can view them all at once, rather than have to scroll through a huge list of boards, until I find "g_platformer" (because I have a lot of boards).

krestyanikov:
Hello everyone.

I also really need this feature because the number of boards is very large. I'd suggest we see how it's done in MaesterTask

https://www.meistertask.com/videos/project_groups.mp4

Thanks!

aproposnix:
Don't project groups already sort of exist? I mean you currently have two groups out of box "All Boards" and "archived boards". In this case, you would provide the user with a way of creating new ones.

stefan-niedermann:
@aproposnix this is currently a flag:

`archived: true / false`

[...]

PR #3325, opened 2021-09-17
add board categories

weeman1337:
* Resolves: #2006
* Target version: master 

### Summary

Adds board categories

- Uncategorised boards are will be displayed under "Uncategorised"
- If there are no categories the section will be named "All boards" as before
- Archived boards will be displayed under "Archived"
- Board categegories can be set in a new section in the board details

image

### TODO

- [ ] The settings icon has a wrong colour. Is there any icon I can use?
- [ ] "Approval" of the category selection field in the sidebar (I had no idea where else to put it)

### Checklist

### DIFF
--- a/lib/Controller/BoardApiController.php
+++ b/lib/Controller/BoardApiController.php
@@ -118,11 +118,13 @@ class BoardApiController extends ApiController {
 	 * @params $title
 	 * @params $color
 	 * @params $archived
+	 * @params $category
 	 *
-	 * Update a board with the specified boardId, title and color, and archived state.
+	 * Update a board with the specified boardId.
 	 */
-	public function update($title, $color, $archived = false) {
-		$board = $this->boardService->update($this->request->getParam('boardId'), $title, $color, $archived);
+	public function update($title, $color, $archived = false, $category = null) {
+		$board = $this->boardService->
+			update($this->request->getParam('boardId'), $title, $color, $archived, $category);
 		return new DataResponse($board, HTTP::STATUS_OK);
 	}
 
--- a/lib/Controller/BoardController.php
+++ b/lib/Controller/BoardController.php
@@ -31,7 +31,12 @@ use OCP\IRequest;
 
 class BoardController extends ApiController {
 	private $userId;
+
+	/**
+	 * @var BoardService
+	 */
 	private $boardService;
+
 	private $permissionService;
 
 	public function __construct($appName, IRequest $request, BoardService $boardService, PermissionService $permissionService, $userId) {
@@ -73,10 +78,11 @@ class BoardController extends ApiController {
 	 * @param $title
 	 * @param $color
 	 * @param $archived
+	 * @param $category
 	 * @return \OCP\AppFramework\Db\Entity
 	 */
-	public function update($id, $title, $color, $archived) {
-		return $this->boardService->update($id, $title, $color, $archived);
+	public function update($id, $title, $color, $archived, $category = null) {
+		return $this->boardService->update($id, $title, $color, $archived, $category);
 	}
 
 	/**
@@ -150,7 +156,7 @@ class BoardController extends ApiController {
 	/**
 	 * @NoAdminRequired
 	 * @param $boardId
-	 * @return \OCP\Deck\DB\Board
+	 * @return \OCA\Deck\Db\Board
 	 */
 	public function clone($boardId) {
 		return $this->boardService->clone($boardId, $this->userId);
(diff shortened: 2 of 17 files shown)


stefan-niedermann:
@weeman1337 as an API consumer i wonder how those new properties are exposed at and can be set with the Rest API?

Maybe you can add some documentation regarding this topic? 🙂 

cc @desperateCoder

desperateCoder:
Please allow me to add my two cents:

As I can see in the migration script, you plan exactly one category per board. Are you sure you want to have a hard time extending it once as customers need more than one?

Also, this leads to redundancy and has an impact on performance, even if its indexed. Please consider a separate table with an n:m relation table.

weeman1337:
> As I can see in the migration script, you plan exactly one category per board. Are you sure you want to have a hard time extending it once as customers need more than one?

If that happens at one day in the future - isn't it just another migration then? It should be one category per board keeping things simple.

> Also, this leads to redundancy and has an impact on performance, even if its indexed. Please consider a separate table with an n:m relation table.

I am fine with the db normalisation :+1: Will do that later.  
The n:m relation looks kind of over-engineered to me when not using it.

desperateCoder:
Thanks for considering!

I read "category" and was like expecting, that a board can have more than one, otherwise the wording would have been something like "folder" or something. I guess I'm wrong here.

Regarding migrations: extracting the categories into a dedicated table will make things much easier, so I'm totally fine with your suggested solution. I just thought it isn't that much of an additional effort to make the relationship n:m and I know users very well. They'll complain and I'll then have to send them right back to you (I'm the co-maintainer of the android deck app). 

Since migration in Android (and in general) is pure pain, I'm happy that you are at least willing to extract the categories, so: thanks! 😅

weeman1337:
> I read "category" and was like expecting, that a board can have more than one, otherwise the wording would have been something like "folder" or something. I guess I'm wrong here.

Regarding this point I may to refer to this https://github.com/nextcloud/deck/issues/2006#issuecomment-897036541 There has been more than a month time to give feedback on it. Getting feedback before starting the work helps keeping the motivation high and doesn't waste time :)

> Since migration in Android (and in general) is pure pain, I'm happy that you are at least willing to extract the categories, so: thanks!

I will update the DB structure because normalisation is a valid point. On the other hand I don't expect so many boards that this field may become a real problem or performance issue. Extracting it increases complexity. For example the relation has to be kept. Maybe this will be a tie performance wise :shrug: 

Anyway I am not planning to adapt the API to keep categories separated. This would make it much more complex.

desperateCoder:
> Getting feedback before starting the work helps keeping the motivation high and doesn't waste time :)

Totally agree! I was mentioned here and just threw my two cents in. @stefan-niedermann is the one keeping up with you guys, I'm kind of the coding monkey here. Didn't know there was a previous discussion on that, sorry.

> On the other hand I don't expect so many boards that this field may become a real problem or performance issue

Nextcloud GmbH is kind of proud of their scalability, so I thought that was a thing here. You're right in 99% of the cases, and I understand that.  Don't take me too serious, it's mostly that I just have done it differently, but since you're the one contributing, who am I to complain? Thanks for your contribution!

nimishavijay:
Agreed, it completely depends on the use case.

If it is called a category then I would expect a board would belong to only one category. Eg. Under category `Personal` there would be boards `Daily tasks`, `Personal project`. Under category `Work` there would be boards `Bug fixes`, `Marketing`.
Categories would be a way of grouping boards.

If a board belongs to multiple categories, then I would associate that with tags and groups more than folders or categories. For example a board `Personal website` could be tagged with `2021`, `college applications`, `personal`. 

It really depends on the use case, but I am leaning towards 1 category per board for simplicity and familiarity as most of us are used to a folder hierarchy.

weeman1337:
Thanks for all the feedback. So it should be 1 board has 0/1 categories? Do we want to make it user dependent like @juliushaertl or is it bound to the board?

nimishavijay:
It can be user dependent because that way it can be more flexible

ilippert:
> Should it be 1 board left_right_arrow 0..1 category or 1 board left_right_arrow 0..* categories

I am coming late to this - just want to share a usecase: I see myself one of many people working in small organisations - in which we create projects (represented here in boards). These boards/projects often come with synergistic targets: they belong then to different categories. My specific use case is in academia: a literature review board can be part of the category "teaching" as well as "research on X" as well as "research on Y".

weeman1337:
Closing that because it looks like it is still unclear what and how "categories" or "board grouping" should work.

The discussion can be found here #2006

PR #3343, opened 2021-09-30
Fix search pagination cursor

eneiluj:
If we get both boards and cards when searching and the number of cards is higher than the limit, the wrong index is used to get the pagination cursor.
This avoids the crash but the search is still not completely ok because we get the same boards in each page. Board search does not use the cursor.

### DIFF
--- a/lib/Search/DeckProvider.php
+++ b/lib/Search/DeckProvider.php
@@ -81,11 +81,11 @@ class DeckProvider implements IProvider {
 				$results
 			);
 		}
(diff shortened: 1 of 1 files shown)


nickvergessen (approved):

juliushaertl (approved):

juliushaertl:
Ah let me revert that in favor of https://github.com/nextcloud/deck/pull/3344 actually and do the backports there.

PR #3362, opened 2021-10-04
Improve search performance

eneiluj:
* Move some work from php to the database when searching for boards:
    * search query
    * cursor (max last modified date)
* Avoid `since` in DB queries when not needed
* Harmonize `since` and `before` filter parameters across services


juliushaertl (approved):
Nice one :rocket: 

At some point we should really switch over to the query builder there, but lets move on with this :+1:

PR #3365, opened 2021-10-06
Switch to QBMapper in BoardMapper

eneiluj:
One downside of this is that one query which was a UNION of 2 queries (in `findAllByUser()`) is now split in 2 queries because I couldn't convert it to QueryBuilder calls. Maybe there is a ninja way to still do it in one single query with joins only. Or it's possible to generate a join somehow...

This breaks some other PRs like #3325. I can wait until they are merged and then adapt this.

### DIFF
--- a/lib/Db/BoardMapper.php
+++ b/lib/Db/BoardMapper.php
@@ -25,12 +25,14 @@ namespace OCA\Deck\Db;
 
 use OC\Cache\CappedMemoryCache;
 use OCP\AppFramework\Db\DoesNotExistException;
+use OCP\AppFramework\Db\QBMapper;
+use OCP\DB\QueryBuilder\IQueryBuilder;
 use OCP\IDBConnection;
 use OCP\IUserManager;
 use OCP\IGroupManager;
 use Psr\Log\LoggerInterface;
 
-class BoardMapper extends DeckMapper implements IPermissionMapper {
+class BoardMapper extends QBMapper implements IPermissionMapper {
 	private $labelMapper;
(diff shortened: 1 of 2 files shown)


eneiluj:
We've got a problem :disappointed: : The trick to add a column with a fixed value to all result rows does not work with the query builder

Such query:
``` sql
SELECT 0 AS `alias` FROM `table` `b` WHERE `col` = "value";
```
can't be generated.
When using `selectAlias()` we get:
``` sql
SELECT `0` AS `alias` FROM `table` `b` WHERE `col` = "value";
```
Which (surprisingly) works in SQLite but not in MySQL/PostgreSQL.

`selectAlias()` always quotes the column name (https://github.com/nextcloud/server/blob/d0a1d99c89f83b4b71a6d8e89d7674f77e4654ed/lib/private/DB/QueryBuilder/QueryBuilder.php#L544)

Is it worth making a PR in server to add a boolean parameter in `selectAlias()` to avoid escaping the column?

juliushaertl (approved):
Tested, works and code looks good 👍

juliushaertl:
@eneiluj Mind to push the update of the psalm baseline, the ones are fine to ignore as you mentioned above.

juliushaertl:
Took the liberty to rebase and push the psalm fix as I based https://github.com/nextcloud/deck/pull/3444 on it 😉

eneiluj:
> Took the liberty to rebase and push the psalm fix as I based #3444 on it wink

Of course, thanks!

PR #3384, opened 2021-10-20
Keep exceptions http response generic

juliushaertl:
There is no need to return the exception details through the API as it is logged properly.

### DIFF
--- a/lib/Middleware/ExceptionMiddleware.php
+++ b/lib/Middleware/ExceptionMiddleware.php
@@ -32,6 +32,7 @@ use OCP\AppFramework\Http\JSONResponse;
 use OCP\AppFramework\OCS\OCSException;
 use OCP\AppFramework\OCSController;
 use OCP\ILogger;
+use OCP\IRequest;
 use OCP\Util;
 use OCP\IConfig;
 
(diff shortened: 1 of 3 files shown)


juliushaertl:
> I agree that proper hints may be a alternative solution, but i doubt the order of the implementation. This should be done first and then remove any trace that might lead to a solution.

As Lukas mentioned the trace may contain path and folder structures but depending on the server configuration and PHP version may also contain arguments of method calls.

> If one would prioritize clear hints higher than removing all information i wouldn't even have an issue, but in fact by far most of the users will complain at the Android client if an issue occurs.

Your point if of course understandable. Nextcloud offers a hint exception in order to provide user facing texts but that is so far not used in deck unfortunately. I'll open a follow up ticket on how we can improve exception handling in general as we indeed should rather catch them in the controller and return predictable API responses then. However that is unfortunately a larger restructuring task, so this is why this PR was the first take on the topic.

> I have a feeling that we are some kind of first level support / first line of defense and i feel a bit lost if we only are able to comment "sorry, but no idea - ask your million users hoster to enable debug mode for you shrug".

For now I've extended the message that will be returned in non-debug mode to always contain a hint to further request the server administrator for help in addition with the request ID which can then be used to get the actual log entry. This is the same behavior that we have in Nextcloud on pages that throw an exception.

stefan-niedermann:
Thank you for your considerations, we'll see how it will work out in the field.

PR #3428, opened 2021-11-16
Allow to download an attachment without navigating to the files app

juliushaertl:
This restores similar behavior that we had before moving attachments to the files app so that a file that has no viewer support can be downloaded right away without going to files first.

### DIFF
--- a/src/components/card/AttachmentList.vue
+++ b/src/components/card/AttachmentList.vue
@@ -22,7 +22,7 @@
 
 <template>
 	<AttachmentDragAndDrop :card-id="cardId" class="drop-upload--sidebar">
-		<div class="button-group" v-if="!isReadOnly">
+		<div v-if="!isReadOnly" class="button-group">
 			<button class="icon-upload" @click="uploadNewFile()">
 				{{ t('deck', 'Upload new files') }}
 			</button>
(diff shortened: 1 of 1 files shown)


src/components/card/AttachmentList.vue
```diff
@@ -79,6 +79,9 @@
 					<ActionLink v-if="attachment.extendedData.fileid" icon="icon-folder" :href="internalLink(attachment)">
 						{{ t('deck', 'Show in Files') }}
 					</ActionLink>
+					<ActionLink v-if="attachment.extendedData.fileid" icon="icon-download" :href="downloadLink(attachment)">
```
mejo-, line 82: How about adding `:download="attachment.name"` in order to really make it a download link?

mejo-: Actually, just adding `download` is enough. I tested it locally and it works as expected.

mejo-: Shall I push the change or do you take care of it @juliushaertl?

juliushaertl: Added and pushed :)

mejo- (approved):
Perfect, looks good to me :)

Issue #2797, opened 2021-02-10
duplicating a board copies its lists, but not the cards


laurentperrinet:
**Describe the bug**
The drop-down menu when right-clicking on the boards' list offers to duplicate a board. However, this copies lists, not cards

**To Reproduce**
1. Create an empty board "Test" - fill with 2 lists and 3 cards
2. Right-click on the board and use the "duplicate" menu item
3. Select the "Test (copy)" board

<img width="605" alt="image" src="https://user-images.githubusercontent.com/381808/107477984-03914900-6b79-11eb-8aa2-25589dd68724.png">

**Screenshots**

<img width="605" alt="image" src="https://user-images.githubusercontent.com/381808/107478027-14da5580-6b79-11eb-96e0-1980f283b403.png">

**Client details:**
 - OS: macOS
 - Browser: Firefox
 - Nextcloud: 20.0.7
 - Deck: 1.2.4

dodedodo:
Never thought about it that way. For me this is expected behaviour. Although I agree 'Use board as template' might be a better name for this action.

A-S-Evers:
"User board as template" would be really helpful...good suggestion...would use it right away...

jimsrobot:
Yes to copying the cards, not sure what the point of cloning is without using the cards.

stefan-niedermann:
@jimsrobot Please use **[GitHub reactions](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/)** 👍 to show that you are affected by the same issue. Please don't comment if you have no relevant information to add.

pascalBokBok:
Should be two options:

- Clone board
- Clone board with cards

stefan-niedermann:
For reference, this is the UI of the Android app, which provides a checkbox when attempting to clone a board:

[![grafik](https://user-images.githubusercontent.com/4741199/110789512-40cc2200-8270-11eb-802b-ce338c3a2322.png)](https://user-images.githubusercontent.com/4741199/110789457-2f831580-8270-11eb-8c47-970d15469922.png)

pascalBokBok:
Deck would be very useful for project-templates if it could copy the cards with the board.

martin-77:
Well, before this gets closed: I need this feature too.
We have recurring lists, always the same tasks, which have to be done.
It is structured like this: 

- before project starts
- 1. Week while project is running
- 2. Week while project is running
- evaluate after 5 Weeks
- ...

[...]

PR #3430, opened 2021-11-18
Clone cards together with the board

bahuma20:
* Resolves: #2797
* Target version: master 

### Summary
| Description | Screenshot |
|---|---|
| Add the possibility to clone cards together with the board. | image |
| Also allow customizations when cloning. | image |
| In advanced options, cards can be moved to the first list and archived cards can be restored. | image | 

### Follow Up
- [ ] Copy attachments
- [ ] Copy comments

### TODO

- [x] Write tests (I need help here, i don't know how to do this in Nextcloud)
- [x] Maybe adjust documentation. Right now only REST api doc was updated.

### Checklist

### DIFF
--- a/lib/Service/BoardService.php
+++ b/lib/Service/BoardService.php
@@ -16,6 +16,7 @@ use OCA\Deck\Db\AclMapper;
 use OCA\Deck\Db\AssignmentMapper;
 use OCA\Deck\Db\Board;
 use OCA\Deck\Db\BoardMapper;
+use OCA\Deck\Db\Card;
 use OCA\Deck\Db\CardMapper;
 use OCA\Deck\Db\ChangeHelper;
 use OCA\Deck\Db\IPermissionMapper;
@@ -29,6 +30,7 @@ use OCA\Deck\Event\AclCreatedEvent;
 use OCA\Deck\Event\AclDeletedEvent;
 use OCA\Deck\Event\AclUpdatedEvent;
 use OCA\Deck\Event\BoardUpdatedEvent;
+use OCA\Deck\Event\CardCreatedEvent;
 use OCA\Deck\NoPermissionException;
 use OCA\Deck\Notification\NotificationHelper;
 use OCA\Deck\Validators\BoardServiceValidator;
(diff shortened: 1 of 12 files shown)


stefan-niedermann:
1. We should use one term (`clone` *or* `copy`)
2. Copy assignments, labels and due dates should be checked by default in my opinion
3. @juliushaertl do we really want the other two checkboxes? I consider them being edge cases, but at least they should be visually seperated from the others because they have another context and are some kind of "destructive" (they "change" or "modify" the data whole copying)

danir-de:
> 3. @juliushaertl do we really want the other two checkboxes? I consider them being edge cases, but at least they should be visually seperated from the others because they have another context and are some kind of "destructive" (they "change" or "modify" the data whole copying)

Restoring archived cards would in no way be an edge case! 
People use deck cloning to create a complex deck preset they can clone for repeating projects/tasks.

juliushaertl:
Hello @bahuma20. I'm very sorry for the long delay on the feedback of this one, we'll have a look into it soonish to get this merged.

lib/Service/BoardService.php
```diff
@@ -80,11 +85,13 @@ class BoardService {
 	public function __construct(
 		BoardMapper $boardMapper,
 		StackMapper $stackMapper,
+		CardMapper $cardMapper,
```
luka-nextcloud, line 88: We have 2 injections of `CardMapper`, see below at line 97.

lib/Service/BoardService.php
```diff
@@ -641,13 +650,14 @@ public function deleteAcl($id) {
 	 * @throws \OCP\AppFramework\Db\MultipleObjectsReturnedException
 	 * @throws BadRequestException
 	 */
-	public function clone($id, $userId) {
+	public function clone($id, $userId, $withCards = false, $withAssignments = false, $withLabels = false, $withDueDate = false, $moveCardsToLeftStack = false, $restoreArchivedCards = false) {
```
luka-nextcloud, line 653: We need to add php docs function parameters to the clone function in the `BoardService` class.

lib/Service/BoardService.php
```diff
@@ -662,6 +672,16 @@ public function clone($id, $userId) {
 		]);
```
luka-nextcloud, line 672: We can separate the board cloning logics in a different class - The BoardMapper or the Service.

lib/Service/BoardService.php
```diff
@@ -679,9 +699,92 @@ public function clone($id, $userId) {
...
 		}
 
+		if ($withCards) {
+			$this->cloneCards($board, $newBoard, $withAssignments, $withLabels, $withDueDate, $moveCardsToLeftStack, $restoreArchivedCards);
+		}
+
 		return $newBoard;
 	}
 
+	private function cloneCards(Board $board, Board $newBoard, $withAssignments = false, $withLabels = false, $withDueDate = false, $moveCardsToLeftStack = false, $restoreArchivedCards = false) {
```
luka-nextcloud, line 709: We can put this Card cloning logic into the `CardMapper` class by creating a static method.

src/components/navigation/AppNavigationBoard.vue
```diff
@@ -20,124 +20,129 @@
...
-					icon="icon-user"
-					:disabled="updateDueSetting"
-					:class="{ 'forced-active': board.settings['notify-due'] === 'assigned' }"
-					@click="updateSetting('notify-due', 'assigned')">
-					{{ t('deck', 'Assigned cards') }}
+					<ActionButton
+						name="notification"
+						icon="icon-sound"
+						:disabled="updateDueSetting"
+						:class="{ 'forced-active': board.settings['notify-due'] === 'all' }"
```
luka-nextcloud, line 88: I think we'd better put this expression as a computed value.

tests/unit/Service/BoardServiceTest.php
```diff
@@ -103,11 +107,13 @@ public function setUp(): void {
 		$this->service = new BoardService(
 			$this->boardMapper,
 			$this->stackMapper,
+			$this->cardMapper,
```
luka-nextcloud, line 110: Two CardMapper usages in the BoardService constructor.

jancborchardt:
> 1. We should use one term (`clone` *or* `copy`)

Agree with @stefan-niedermann – we should use `Copy` everywhere since that is the more commonly used word.

danir-de:
> > 1. We should use one term (`clone` _or_ `copy`)
> Agree with @stefan-niedermann – we should use `Copy` everywhere since that is the more commonly used word.

Yes but `Copy` often requires pasting something and/or choosing where to copy it to.
`Clone` on the other hand is mostly used to create a copy (a clone) to the exact same location.

I'd say that `Clone` or `Duplicate` would be way more fitting than `Copy`.

small1:
This would be nice to have merged. This would solve a problem for a few projects where they have asked for this type of feature.

juliushaertl:
I rebased this as there were quite some conflicting changes. Will have a look what is missing to get this in.

lib/Service/BoardService.php
```diff
@@ -683,6 +648,87 @@ private function enrichBoards(array $boards, bool $fullDetails = true): array {
 		return $boards;
 	}
 
+	private function cloneCards(Board $board, Board $newBoard, bool $withAssignments = false, bool $withLabels = false, bool $withDueDate = false, bool $moveCardsToLeftStack = false, bool $restoreArchivedCards = false): void {
+		// TODO: Undelete cards
+		// TODO: Copy attachments (or not?)
+		// TODO: Copy comments (or not?)
+		// TODO: Move to specific column
```
juliushaertl, line 655: Still valid?

lib/Service/BoardService.php
```diff
@@ -683,6 +648,87 @@ private function enrichBoards(array $boards, bool $fullDetails = true): array {
...
+			$archivedCards = $this->cardMapper->findAllArchived($stack->getId());
+
+			/** @var Card[] $cards */
+			$cards = array_merge($cards, $archivedCards);
+
+			foreach ($cards as $card) {
+				$targetStackId = $moveCardsToLeftStack ? $newStacks[0]->getId() : $newStacks[$i]->getId();
+
+				// Create a cloned card.
+				$newCard = new Card();
```
juliushaertl, line 679: TODO: check if we have a better mechansim for cloning to just set the new related ids?

src/components/navigation/AppNavigationBoard.vue
```diff
@@ -247,15 +253,24 @@ export default {
...
+				this.loading = true
+				try {
+					const newBoard = await this.$store.dispatch('cloneBoard', {
+						boardData: this.board,
+						settings: data,
+					})
+					this.loading = false
+					await this.$router.push({ name: 'board', params: { id: newBoard.id } })
+				} catch (e) {
+					OC.Notification.showTemporary(t('deck', 'An error occurred'))
```
juliushaertl, line 271: TODO: Be more specific about the error

lib/Service/BoardService.php
```diff
@@ -683,6 +648,87 @@ private function enrichBoards(array $boards, bool $fullDetails = true): array {
...
+
+		$stacks = $this->stackMapper->findAll($board->getId());
+		usort($stacks, function (Stack $a, Stack $b) {
+			return $a->getOrder() - $b->getOrder();
+		});
+
+		$newStacks = $this->stackMapper->findAll($newBoard->getId());
+		usort($stacks, function (Stack $a, Stack $b) {
+			return $a->getOrder() - $b->getOrder();
+		});
```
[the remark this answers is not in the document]
bahuma20: No, has all stacks of the old board and the other one the stacks of the new board.

lib/Service/BoardService.php
grnd-alt: I've checked that and it seems to be not possible. As, when inserting, only fields marked as updated get Insterted into the DB, and according to [this](https://github.com/nextcloud/server/blob/500a9cb7244d3072f83eebc878e27833572e707d/lib/public/AppFramework/Db/QBMapper.php#L122) there is no other way to set all fields as updated other than using the corresponding setter.

lib/Service/BoardService.php
juliushaertl: Let's remove those from the code and track as enhancement follow up issues.

lib/Service/BoardService.php
```diff
@@ -669,6 +648,88 @@ private function enrichBoards(array $boards, bool $fullDetails = true): array {
...
+					foreach ($assignments as $assignment) {
+						$this->assignmentService->assignUser($newCard->getId(), $assignment->getParticipant(), $assignment->getType());
+					}
+				}
+
+
+				// Copied from CardService because CardService cannot be injected due to cyclic dependencies.
+				$this->activityManager->triggerEvent(ActivityManager::DECK_OBJECT_CARD, $card, ActivityManager::SUBJECT_CARD_CREATE);
+				$this->changeHelper->cardChanged($card->getId(), false);
+				$this->eventDispatcher->dispatchTyped(new CardCreatedEvent($card));
```
luka-nextcloud, line 726: `$card` should be replaced by `$newCard`

luka-nextcloud (commented):
I think cypress tests should be added.

cyrillappert:
Hello. We would also be very interested in this feature.

TheBestDestroyer:
I would also be very interested in this feature. Why it has not been merged yet?

juliusknorr:
Pushed some tests and fixed remaining review comments where reasonable.

I'll get this merged for now even if we lack attachment and comment cloning. I think this is valuable enough without so we can get this unblocked.

juliusknorr (approved):

Issue #3183, opened 2021-07-12
`ETag` is not updated if settings of a board have changed


stefan-niedermann:
1. Check `ETag` of a board
2. Change notification setting
3. Expect `ETag` to be different from step 1. - it is not.

Issue #2892, opened 2021-03-15
ETags not updated if card has been added to a new project


stefan-niedermann:
Given a card gets added to a new project in the Web UI (e. g. linking it to another card or a board), the `ETag`s are not updated. The changes are therefore not visible in the Deck Android app and other API clients.

stefan-niedermann:
Might be caused by / related to #2485

juliushaertl:
Yes, we cannot trigger anything there as the projects API is fully handled by the server and there are no events emitted yet so deck itself doesn't know about changes apart from the frontend which just queries the OCS API.

stefan-niedermann:
Closing given `Projects` feature has been deprecated in favor of `Related Ressources`

Issue #2866, opened 2021-03-07
Update stack / list: ETag not updated


desperateCoder:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Describe the bug**
Within the Android Deck app, we use the REST-Endpoints to fetch data from the server using ETags. When a User changes the name of a list, the ETags don't seem to get updated, so our Android client only gets notified, when the user also changes a card within the list. 

See this: https://github.com/stefan-niedermann/nextcloud-deck/issues/861

**To Reproduce**
1. WebUI: change list name
2. Android: pull to refresh
3. Name of the list won't get updated
4. WebUI: change a card in same list
5. Android pull to refresh
6. changed card and list both get updated

**Expected behavior**
On list-update, the etags of the list and the board should be updated

**Client details:**
see here: https://github.com/stefan-niedermann/nextcloud-deck/issues/861

CC: @stefan-niedermann

j-lakeman:
Same issue seems also to apply to the order of the lists.

abelbabel:
Same issue when moving a card: the ETag of the board changes and of the stack that receives the card, but ETag of the source-stack remains the same.

Jannik44:
same issue here, sync of list names is not working from server to app with etags enabled

PR #3444, opened 2021-11-23
Cache most frequent queries

juliushaertl:
Contributes to reducing the amount of sql queries that are issued when setting up the filesystem. 

This PR is based on a rebased version of https://github.com/nextcloud/deck/pull/3365 (needs another cleanup after the merge)

- First commit addesses the most frequent query that would resolve the relation from the card id to the board id. This one is now cached in a distributed memory cache. The only condition that the cache should become invalid is when a card is moved to a different board where it is invalidated manually, therefore no TTL is used. (44042 queries in todays query log)
- Second commit aggregates some more savings when fetching board details of single boards (and their labels or ACL). The ACL were the most frequent ones (19111 queries in todays log) but this will would also help in other places where board details may be obtained multiple times in the same request.


lib/Db/CardMapper.php
```diff
@@ -547,11 +557,17 @@ public function isOwner($userId, $cardId): bool {
...
 	public function findBoardId($id): ?int {
-		$sql = 'SELECT id FROM `*PREFIX*deck_boards` WHERE `id` IN (SELECT board_id FROM `*PREFIX*deck_stacks` WHERE id IN (SELECT stack_id FROM `*PREFIX*deck_cards` WHERE id = ?))';
-		$stmt = $this->db->prepare($sql);
-		$stmt->bindParam(1, $id, \PDO::PARAM_INT);
-		$stmt->execute();
-		return $stmt->fetchColumn() ?? null;
+		//$this->cache->remove('findBoardId:' . $id);
+		$result = $this->cache->get('findBoardId:' . $id);
+		if ($result === null) {
+			$sql = 'SELECT id FROM `*PREFIX*deck_boards` WHERE `id` IN (SELECT board_id FROM `*PREFIX*deck_stacks` WHERE id IN (SELECT stack_id FROM `*PREFIX*deck_cards` WHERE id = ?))';
```
nickvergessen, line 563: btw you can put this in query builders as some point:
https://github.com/nextcloud/server/blob/215aef3cbdc1963be1bb6bca5218ee0a4b7f1665/lib/private/Repair/RemoveLinkShares.php#L115

nickvergessen (approved):

lib/Db/CardMapper.php
```diff
@@ -107,6 +113,10 @@ public function update(Entity $entity, $updateModified = true): Entity {
 			} catch (Exception $e) {
 			}
 		}
+		// Invalidate cache when the card may be moved to a different board
+		if ($updatedFields['stackId']) {
+			$this->cache->remove('findBoardId:' . $entity->getId());
```
juliushaertl, line 118: - [ ] Probably also makes sense to invalidate on delete

lib/Db/BoardMapper.php
```diff
@@ -42,7 +42,10 @@ class BoardMapper extends QBMapper implements IPermissionMapper {
 
 	private $circlesEnabled;
 
+	/** @var CappedMemoryCache */
```
juliushaertl, line 45: - [ ] reuse boardCache and only store board ids in the userBoardCache

lib/Db/CardMapper.php
juliushaertl: @nickvergessen Will do here right away then. Thinking about the query, weren't joins usually also a bit faster than sub-queries?

nickvergessen: Yeah, looking at this query if there is always a 1:n (card to stack and stack to board) relation, you could optimise for reading and add a board id column to the cards table as well and then save 2 queries?

juliushaertl: Moved to a single join query for now as introducing a dedicated board_id column would be a larger topic with possibly heavy migrations involved depending on the table size, should already be a good improvement now with the additional caching in place.

lib/Db/BoardMapper.php
juliushaertl: Thinking more about that it should be fine as it is, the current way has some memory overhead but we can avoid juggling around the existing objects to map to board ids and the other way round when obtaining the boards from the cache.

stefan-niedermann:
Since i saw adjusted unit tests (changing properties from `[]` to `null`) i am wondering whether those changes are breaking API changes? cc @desperateCoder

juliushaertl:
Good catch, yes, though the fetching of acls/labels should happen and they should then be set with an empty array for the API requests that expose a Board entity this is worth double checking (or making sure that we stick to the current empty array in any case when performing the json serialization).

mejo- (approved):
Overall looks good to me. I went through the code and didn't spot anything obvious :)

eneiluj (approved):
L:eyes:ks good.

stefan-niedermann:
Maybe i did miss it while reviewing the code, but it looks like the `ETag`s are not updated in the cache? This would lead to more trouble with outdated data (like #3183, #2892 or #2866)

stefan-niedermann (changes requested):
Maybe i did miss it while reviewing the code, but it looks like the `ETag`s are not updated in the cache? This would lead to more trouble with outdated data (like #3183, #2892 or #2866)

juliushaertl:
> Maybe i did miss it while reviewing the code, but it looks like the `ETag`s are not updated in the cache? This would lead to more trouble with outdated data (like #3183, #2892 or #2866)

The board caching layer is a CappedMemoryCache which means it is only persisted during the same request. Meaning any request will at least do one fetch from the database and would then get the latest etag for the board.

stefan-niedermann (approved):

juliushaertl:
Should also get an adapted backport to 21/22.

juliushaertl:
Manual backport in https://github.com/nextcloud/deck/pull/3740

Issue #3451, opened 2021-11-25
1.6.0-beta3: upcoming cards view is empty


jakobroehrl:
On every call I get this message:
Exception: count(): Argument #1 ($value) must be of type Countable|array, null given in file '/var/www/nextcloud/apps/deck/lib/Service/OverviewService.php' line 116

juliushaertl:
Thanks for reporting and testing the beta :) I can reproduce this on my personal instance though not yet on my local development setup. Will keep investigating.

PR #3458, opened 2021-11-29
Fix overview card listing

juliushaertl:
Fixes https://github.com/nextcloud/deck/issues/3451

Make sure that the setter behaves as expected when setting an empty array for boards without acl rules in https://github.com/nextcloud/deck/blob/901b8f25060157835157b882859f117200165e8b/lib/Db/BoardMapper.php#L104

Otherwise the new default value of null would make the board object behave as if the acls have not been obtained yet.


eneiluj (approved):
I couldn't reproduce the issue locally but changes look good.

PR #3498, opened 2021-12-28
Update composer config

juliushaertl:
- Avoid pulling in to new psr/container which is no longer compatible with PHP 7.3
- Add allowed plugins for composer 2.2.x

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,13 @@
 	},
 	"config": {
 		"optimize-autoloader": true,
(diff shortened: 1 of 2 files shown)


juliushaertl:
Should make phpunit tests pass again

mejo- (approved):
Looks good to me :)

eneiluj (approved):
Seems a'ight :grin:

Issue #3312, opened 2021-09-09
External CalDAV access no longer works after upgrade to 1.4.4


birefringence:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Describe the bug**
After upgrading to Deck 1.4.4 I can no longer access the automatically created calendars/tasks via CalDAV. I always get a 404 error. Access via the Nextcloud Calendar/Tasks web apps still works.

**To Reproduce**
1. Upgrade to Deck 1.4.4
2. Try to sync Deck tasks with TbSync plugin (Thunderbird) or DAVx5 (Android)
3. PROPFIND returns 404 on each access attempt.

**Expected behavior**
PROPFIND should return 207 and synchronisation should continue.

**Client details:**
 - OS: Linux/Android
 - TbSync 1.24 / DAVx5 3.4-ose
 - Device: PC, Sony Xperia XA2

<details>
<summary>Server details</summary>

**Operating system**: Linux 5.10.52-v7+ #1441 SMP Tue Aug 3 18:10:09 BST 2021 armv7l

**Web server:** nginx/1.14.2 (fpm-fcgi)

**Database:** mysql 10.3.29

**PHP version:** 7.3.29-1~deb10u1
Modules loaded: Core, date, libxml, openssl, pcre, zlib, filter, hash, Reflection, SPL, sodium, session, standard, cgi-fcgi, mysqlnd, PDO, xml, apcu, bcmath, bz2, calendar, ctype, curl, dom, mbstring, fileinfo, ftp, gd, gettext, gmp, iconv, igbinary, imagick, intl, json, exif, mysqli, pdo_mysql, pdo_pgsql, pgsql, apc, posix, readline, redis, shmop, SimpleXML, sockets, sysvmsg, sysvsem, sysvshm, tokenizer, wddx, xmlreader, xmlwriter, xsl, zip, Phar, Zend OPcache

**Nextcloud version:** 21.0.4 - 21.0.4.1

**Where did you install Nextcloud from:** Manual installation from ZIP provided on https://nextcloud.com/

**Signing status:** Array()

```
No errors have been found.
```

**List of activated apps:**

```
Enabled:
 - accessibility: 1.7.0
[...]

PR #3500, opened 2021-12-29
Fix CalDAV blocking and modernize circles API usage

juliushaertl:
- Avoid blocking calendar access if something goes wrong while fetching deck entries
- Move any circles API usage to internal service and make use of new circles API to get the list of circles for a user

* Resolves: #3312

### DIFF
--- a/lib/Db/BoardMapper.php
+++ b/lib/Db/BoardMapper.php
@@ -24,6 +24,7 @@
 namespace OCA\Deck\Db;
 
 use OC\Cache\CappedMemoryCache;
+use OCA\Deck\Service\CirclesService;
 use OCP\AppFramework\Db\DoesNotExistException;
 use OCP\AppFramework\Db\QBMapper;
 use OCP\DB\QueryBuilder\IQueryBuilder;
@@ -38,9 +39,8 @@ class BoardMapper extends QBMapper implements IPermissionMapper {
 	private $stackMapper;
 	private $userManager;
 	private $groupManager;
+	private $circlesService;
 	private $logger;
-
-	private $circlesEnabled;
 
 	/** @var CappedMemoryCache */
 	private $userBoardCache;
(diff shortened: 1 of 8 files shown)


juliushaertl:
@ArtificialOwl https://github.com/nextcloud/deck/blob/81740c9404699cb956803c119c17b946b396f4a5/lib/Service/CirclesService.php#L56 is still using the old API, would the following work fine for replacing that? 

```php
		// Enforce current user condition since we always want the full list of members
		/** @var CirclesManager $circlesManager */
		$circlesManager = \OC::$server->get(CirclesManager::class);
		$circlesManager->startSuperSession();
		return $circlesManager->getCircle($circleId);
```

ArtificialOwl:
> @ArtificialOwl
> 
> https://github.com/nextcloud/deck/blob/81740c9404699cb956803c119c17b946b396f4a5/lib/Service/CirclesService.php#L56
> is still using the old API, would the following work fine for replacing that?
> 
> ```
> 		// Enforce current user condition since we always want the full list of members
> 		/** @var CirclesManager $circlesManager */
> 		$circlesManager = \OC::$server->get(CirclesManager::class);
> 		$circlesManager->startSuperSession();
> 		return $circlesManager->getCircle($circleId);
> ```

It should work, however you need to be sure that the params in `getCircle(string $circleId)` are not fed by a user: https://github.com/nextcloud/deck/blob/81740c9404699cb956803c119c17b946b396f4a5/lib/Service/CirclesService.php#L51

ArtificialOwl (approved):

ArtificialOwl:
not tested, but code looks good. 

The only comment I have is that after migration of `CirclesService::getCircle()` to the new API, the exception returned by the new API (if Circle is out of reach) will be managed by a `catch (Throwable)`

Issue #2554, opened 2020-11-14
Move card shows deleted boards


cloud2018:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Describe the bug**
Move card shows deleted boards

**To Reproduce**
1. Open move card
2. choose a board
4. See inconsistency: Deleted boards are listed in the selection for target, but not in the left sidebar 

**Expected behavior**
Deleted boards are not listed in the selection for target 

**Screenshots**
![d22](https://user-images.githubusercontent.com/34692335/99143004-73676480-265a-11eb-9cb6-cd2a404ccae8.png)

Note:
Please add feature: order the boards in the selection by name

Versions
Deck 1.2.0-beta1
Nextcloud 20.0.0

stefan-niedermann:
Just for clarity: Boards are not physically deleted immediately, but as soon as the next cronjob has been executed. (This behavior is by design to enable "undo").

Can you please verify, that there only appear deleted boards which have been deleted *recently* and the deleted boards do no longer appear after the next cron job has been executed (for example after 30 minutes or so, depends on your configuration).

I guess there is missing some kind of

```sql
WHERE status != 'DELETED'
```

or so in the SQL query to fetch the proposals.

stefan-niedermann:
Additional note: Maybe moving a card to a board which has been deleted should throw an error (some kind of HTTP 400)? This might prevent data loss when one moves a card to a board which has been deleted, but it might also expected, that the moved card is just gone with the board then...? 🤔 cc @juliushaertl for the desored behavior.

cloud2018:
@stefan-niedermann 
"Can you please verify, that there only appear deleted boards which have been deleted recently and the deleted boards do no longer appear after the next cron job ..." 
sorry, I would have tested that before,  but there are problems with my cron job :-(
In the sidebar the list don't show recently deleted boards. If the dialog shows the same list, there is no need to throw an error

stefan-niedermann:
> If the dialog shows the same list, there is no need to throw an error

[...]

PR #3502, opened 2021-12-30
Exclude deleted boards in the selection for target

luka-nextcloud:

* Resolves: #2554
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/src/components/cards/CardMenu.vue
+++ b/src/components/cards/CardMenu.vue
@@ -58,7 +58,7 @@
 				<h3>{{ t('deck', 'Move card to another board') }}</h3>
 				<Multiselect v-model="selectedBoard"
 					:placeholder="t('deck', 'Select a board')"
-					:options="boards"
+					:options="activeBoards"
 					:max-height="100"
 					label="title"
 					@select="loadStacksFromBoard" />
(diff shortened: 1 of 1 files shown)


src/components/cards/CardMenu.vue
```diff
@@ -129,6 +129,9 @@ export default {
 		isCurrentUserAssigned() {
 			return this.card.assignedUsers.find((item) => item.type === 0 && item.participant.uid === getCurrentUser()?.uid)
 		},
+		activeBoards() {
+			return this.$store.getters.boards.filter((item) => item.deletedAt === 0)
```
juliushaertl, line 133: For additional improvement, it could also filter out archived boards here:

```suggestion
			return this.$store.getters.boards.filter((item) => item.deletedAt === 0 && item.archived === false)
```

src/components/navigation/AppNavigationBoard.vue
```diff
@@ -272,9 +272,7 @@ export default {
 							.then(() => {
 								this.loading = false
 								this.deleted = true
-								this.undoTimeoutHandle = setTimeout(() => {
-									this.$store.dispatch('removeBoard', this.board)
-								}, 7000)
+								this.$store.dispatch('removeBoard', this.board)
```
juliushaertl, line 275: I guess this was for making testing easier, but please revert this change as the undo option should still be available.

luka-nextcloud: @juliushaertl When we delete a board, we need to wait for 7s which is too long? Is there any reason for that?

juliushaertl: UX-wise the idea is that delete is a destructive action and since we currently have nothing like a trashbin for deleted boards, the timeout will allow you to still undo this for some seconds. Lets not change that for now, but we could look into the deletion as an enhancement in the future.

juliushaertl (approved):

Issue #3476, opened 2021-12-08
Username when adding file to card 


ArtemDrach:
Hello everyone.

I have a situation like this:
My username = phone number.
Display name = First name + Last name

When adding a file to the card, the usersname is written next to it.

Is it possible to do this so that the user's display name is written? 

Nextcloud 22.2.3

dodedodo:
This might be somewhat related to #3389?

juliushaertl:
Fixed with https://github.com/nextcloud/deck/pull/4036

PR #3510, opened 2022-01-06
Use display name instead of username when adding file to cards

luka-nextcloud:

* Resolves: #3476
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/lib/Service/AttachmentService.php
+++ b/lib/Service/AttachmentService.php
@@ -60,6 +60,7 @@ class AttachmentService {
 	private $activityManager;
 	/** @var ChangeHelper */
 	private $changeHelper;
+	/** @var IUserManager */
 	private IUserManager $userManager;
 	/** @var AttachmentServiceValidator */
 	private AttachmentServiceValidator $attachmentServiceValidator;
@@ -71,6 +72,7 @@ class AttachmentService {
 		ChangeHelper $changeHelper,
 		PermissionService $permissionService,
 		Application $application,
+		ICacheFactory $cacheFactory,
 		AttachmentCacheHelper $attachmentCacheHelper,
 		$userId,
 		IL10N $l10n,
--- a/lib/Capabilities.php
+++ b/lib/Capabilities.php
@@ -53,7 +53,8 @@ class Capabilities implements ICapability {
 				'canCreateBoards' => $this->permissionService->canCreate(),
 				'apiVersions' => [
 					'1.0',
-					'1.1'
+					'1.1',
+					'1.2'
 				]
 			]
 		];
--- a/lib/Capabilities.php
+++ b/lib/Capabilities.php
@@ -53,8 +53,7 @@ class Capabilities implements ICapability {
 				'canCreateBoards' => $this->permissionService->canCreate(),
 				'apiVersions' => [
 					'1.0',
-					'1.1',
-					'1.2'
+					'1.1'
 				]
 			]
 		];
--- a/lib/Controller/AttachmentApiController.php
+++ b/lib/Controller/AttachmentApiController.php
@@ -27,13 +27,16 @@ use OCP\AppFramework\Http;
 use OCP\AppFramework\Http\DataResponse;
 use OCP\IRequest;
 use OCA\Deck\Service\AttachmentService;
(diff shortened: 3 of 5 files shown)


lib/Service/FilesAppService.php
```diff
@@ -94,11 +98,13 @@ public function listAttachments(int $cardId): array {
...
 				return null;
 			}
+			$user = $this->userManager->get($this->userId);
+			$displayName = $user ? $user->getDisplayName() : '';
 			$attachment = new Attachment();
 			$attachment->setType('file');
 			$attachment->setId((int)$share->getId());
 			$attachment->setCardId($cardId);
-			$attachment->setCreatedBy($share->getSharedBy());
+			$attachment->setCreatedBy($displayName);
```
juliushaertl, line 107: The Entity classes like `Attachment` in deck allow to mark a field as resolvable to then later on fill it with a more detailed object if needed. This way we can pass both the user id (uid) as well as the display name to the frontend using a [User object](https://github.com/nextcloud/deck/blob/fc5c34897e601ac96f2b3c482791d075a35d831d/lib/Db/User.php) with something like this:

```php
$attachment->resolveRelation('createdBy', function ($userId) {
	$user = $this->userManager->get($userId);
	if ($user !== null) {
		return new User($user);
	}
	return $user;
});
```

Now the crucial point here is that we have two API endpoints where the result of this method may be used, the `AttachmentController` is the WebUI entrypoint and the `AttachmentApiController` is the public REST API which is for example used by the community android app. For the bug fixing part of this issue I'd say we should only change the result of the createdBy field to the Object to avoid changes in the API results. This would mean to resolve the relation only in https://github.com/nextcloud/deck/blob/199cccf86bd9386e7bb621e5bef724b02f5dee20/lib/Controller/AttachmentController.php#L44

For the REST API we should probably introduce a minor API version bump similar to https://github.com/nextcloud/deck/blob/199cccf86bd9386e7bb621e5bef724b02f5dee20/lib/Controller/AttachmentApiController.php#L47 and introduce a new 1.2 in https://github.com/nextcloud/deck/blob/199cccf86bd9386e7bb621e5bef724b02f5dee20/lib/Capabilities.php#L54

juliushaertl (changes requested):
I've made some comment inline about the approach, let me know if you have any questions in that regard :)

lib/Service/FilesAppService.php
luka-nextcloud: Please check again. :)

juliushaertl: Maybe you can check the description above about the resolveRelation method and moving it to the AttachmentController again. Unfortunately it cannot just be put in the service class, as that would introduce a breaking API change.

If doing that as mentioned there should be a fallback for the older API versions and the version should be bumped.

luka-nextcloud: @juliushaertl Did you have a chance to check it again?

lib/Service/AttachmentService.php
```diff
@@ -127,6 +131,9 @@ public function findAll($cardId, $withDeleted = false) {
 			try {
 				$service = $this->getService($attachment->getType());
 				$service->extendData($attachment);
+				$user = $this->userManager->get($this->userId);
+				$displayName = $user ? $user->getDisplayName() : '';
+				$attachment->setCreatedBy($displayName);
```
juliushaertl, line 136: Maybe we can use `$this->attachmentMapper->mapOwner($attachment)` here as well

juliushaertl:
@luka-nextcloud Please rebase and fix the conflicts while we review and test

marcelklehr:
Is this still relevant since we have extendedData for attachments? https://github.com/nextcloud/deck/blob/master/lib/Service/AttachmentService.php#L407-L424

luka-nextcloud:
> Is this still relevant since we have extendedData for attachments? https://github.com/nextcloud/deck/blob/master/lib/Service/AttachmentService.php#L407-L424

@juliushaertl I noticed that the issue was already resolved on master. This PR is no longer useful.

juliushaertl:
Yep, indeed this was fixed with https://github.com/nextcloud/deck/pull/4036 in the meantime

PR #3528, opened 2022-01-12
Add app config to toggle the default calendar setting as an admin

juliushaertl:
Allow to change the default setting if calendar integration should be enabled through occ:

```
occ config:app:set deck calendar --value="no"
```


eneiluj (approved):
Should it be documented?

miaulalala (approved):

Issue #2642, opened 2020-12-10
Horizontal scrollbar missing (large tables in description)


q-wertz:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Describe the bug**
When having a table that is wider than the sidebar the part that is too big just "vanishes".
No horizontal scrollbar visible

**To Reproduce**
1. Create a Card with a wide table
2. Switch to view mode

**Expected behavior**
A horizontal scrollbar appears so that one can see the right part of the table

**Screenshots**
![image](https://user-images.githubusercontent.com/20948931/101805757-0a64f680-3b0b-11eb-8135-42334acd72a1.png)

**Deck version:** 1.2.2

**Client details:**
 - OS: Ubuntu 20.10
 - Browser: Firefox 83.0
 - Device: Laptop

PR #3531, opened 2022-01-12
Add horizontal scrollbar into the large table inside description

luka-nextcloud:

* Resolves: #2642
* Target version: master 

### Checklist

### DIFF
--- a/src/components/card/Description.vue
+++ b/src/components/card/Description.vue
@@ -262,6 +262,8 @@ export default {
 
 #description-preview {
 	min-height: 100px;
+	width: auto;
+	overflow-x: auto;
 
 	&::v-deep {
 		@import './../../css/markdown';


src/components/card/Description.vue
```diff
@@ -262,6 +262,8 @@ export default {
 
 #description-preview {
 	min-height: 100px;
+	width: max-content;
+	overflow-x: scroll;
```
juliushaertl, line 266: Haven't tried out yet, but would `auto` also work maybe to not always have a scorllbar?

juliushaertl, line 266:
```suggestion
	overflow-x: auto;
```

Let's drop the max-content as that would also avoid long lines to break - and the auto seems to work fine to not always have a scrollbar in some browsers.

juliushaertl (changes requested):
See inline 😉

eneiluj (approved):
:+1:

mejo- (approved):

Issue #2702, opened 2021-01-05
Improve modal view design


juliushaertl:
The current modal view design is basically a larger sidebar and doesn't properly use the available space. When talking about this in the past with @jancborchardt and @Clementine46 the Trello approach of showing only the available UI parts and having a set of "Add to card" options for attachments seemed quite interesting, so maybe we should consider redesigning the card modal view in a similar way.

With that we could give the users direct access to all card content without the need of switching between tabs.

![image](https://user-images.githubusercontent.com/3404133/103642560-97a53a80-4f53-11eb-97d6-60fa15b11a45.png)

@Clementine46 Would this be something you'd be interested in to see how we could redesign the modal in that regard?

Clementine46:
Here is the modal redesign we came up with:

![01 Deck New Card](https://user-images.githubusercontent.com/50834537/110160612-87db9280-7dec-11eb-829b-0b9e8e9ff065.png)

We got rid of the tabs and went for toggle buttons to save some space. Clicking on one of the button will enable the chosen option. That way, only the needed options will be opened. 

Here are several examples:

**Assigning a member**

![deck members](https://user-images.githubusercontent.com/50834537/110173598-207b0e00-7dff-11eb-9402-75b599b48d89.gif)


**Assigning a tag**

![deck tag](https://user-images.githubusercontent.com/50834537/110173624-2a9d0c80-7dff-11eb-940c-93463573d10d.gif)


**Setting a due date**

![deck due date](https://user-images.githubusercontent.com/50834537/110173635-3092ed80-7dff-11eb-853a-e1788427a3a3.gif)


Once enabled, the button switches to blue so the user knows it's activated. The selection process remains the same, the main change is the way it's presented. Once the user is done choosing, the new category appears with the current selection and the "+" icon allows them to add more.


Here is the modal with members, tags and due date selected:
![07 Card Date 2](https://user-images.githubusercontent.com/50834537/110173722-561ff700-7dff-11eb-98ef-0e4c4ac5b4f8.png)


This content is flexible and moves depending on how much space is taken.

Here, the addition of two other tags is pushing the due date below the first two options:
![deck due date 2](https://user-images.githubusercontent.com/50834537/110175389-f4ad5780-7e01-11eb-960c-c128facdfde7.gif)


Lastly, the buttons are sticky and will stay on top of the modal even when scrolled for easy access.

Here is how it looks like when scrolling to the bottom of the modal:
[...]

PR #3557, opened 2022-01-25
WIP: New CardModal UI

luka-nextcloud:

* Resolves: #2702
* Target version: master 

### TODO

- [ ] ...

### Checklist


luka-nextcloud:
Members tab:
image

luka-nextcloud:
image

eneiluj (commented):
I didn't check the code yet.

* The CardModal component is not displayed in a modal but rather at the right sidebar's location.
* The member/user multiselect disappears on mouseout.
* I don't see the tag input like in your screenshot.

Did I do something wrong? Did you forgot to include some commits?

src/components/card/CardModal.vue
```diff
@@ -0,0 +1,370 @@
...
+		</div>
+		<div class="activities">
+			<h2 class="activities-title">
+				<div class="icon-flash-black" /> Activities
+			</h2>
+			<div class="comments">
+				<Avatar :user="currentUser.uid" />
+				<CommentForm v-model="newComment" @submit="createComment" />
+			</div>
+			<div class="activities-logs">
```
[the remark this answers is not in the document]
juliushaertl: For comments itself I think we should then also reuse the existing component for now:

https://github.com/nextcloud/deck/blob/master/src/components/card/CardSidebarTabComments.vue

Basically other than in the mockups, lets go for something like this to start with for the UI structure of the new modal:

```
Comments
----------
CardSidebarTabComments

Activities
----------
CardSidebarTabActivity
```

src/components/card/CardModal.vue
```diff
@@ -0,0 +1,370 @@
...
+	<div class="container">
+		<div class="top">
+			<h1 class="top-title">
+				Example task 3
+			</h1>
+			<p class="top-modified">
+				Modified: 2 days go. Created 3 days ago
+			</p>
+		</div>
+		<div class="tabs">
```
juliushaertl, line 33: I think there might have been a bit of confusion on what the purpose of the top buttons is, they are not tabs in specific, but more toggle buttons for the actual input fields. Let me maybe describe the idea there a bit further:

Taking the due date as an example
- By default the input field is hidden
- The user decides to add a due date by clicking the "Due date" button
- Then the input is shown
- If a card has a due date set, we always show the input field

The same would be the case for the others (Members, Projects, Attachments, Tags)

img/plus.svg
```diff
@@ -0,0 +1 @@
+<svg fill="#26e07f" xmlns="http://www.w3.org/2000/svg"  viewBox="0 0 24 24" width="24px" height="24px"><path fill-rule="evenodd" d="M 11 2 L 11 11 L 2 11 L 2 13 L 11 13 L 11 22 L 13 22 L 13 13 L 22 13 L 22 11 L 13 11 L 13 2 Z"/></svg>
```
juliushaertl, line 1: This should be available already as a global icon class `icon-add` https://docs.nextcloud.com/server/latest/developer_manual/html_css_design/icons.html

Otherwise we can make use of https://github.com/robcresswell/vue-material-design-icons for pulling in icons.

juliushaertl (changes requested):
Looks fine from the overall structure, I left some comments inline and also a longer one regarding the "tabs" behaviour which might be something to adjust further.

eneiluj (changes requested):
Ok got it, "Use bigger card view" must be checked to display the card modal.

It looks nice!

* If the modal content is too long, there is no way to scroll down. A simple way to solve that is to add `overflow: scroll;` to the `.modal-container` but there might be a cleaner approach.
* We could use more partial vue component import like in line 48 of App.vue
* Clicking on tabs does not have any effect on the modal content. What's the purpose of the tabs?
* User select, tag, date and "Add to project" are all displayed on the same line
aaa
* Multiselect disappears on mouseout for users and tags

luka-nextcloud:
@juliushaertl @max-nextcloud Could you review? Thanks.

juliushaertl:
Let me close this since this has quite some heavy conflicts meanwhile, we can pick this up once relevant again

PR #3600, opened 2022-02-24
Migrate AclMapper to QBMapper

juliushaertl:
☝️ See title


juliushaertl:
Was done with https://github.com/nextcloud/deck/pull/3720 now in one go

Issue #2618, opened 2020-11-25
Insert Attachment button is hard to click


reox:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Describe the bug**

Since I think 1.2.0, the insert attachment button is very hard to use. If you try to click it, the tooltip shows up.
So you have to click the button like crazy to activate it.
Sometimes I can click it within a second, but sometimes it is almost not possible to hit it.
Also the button "flickers" when hovering with the mouse.

**To Reproduce**
1. Upload attachment
2. Edit Description
3. Select Attachment
4. try to click button

**Expected behavior**
Button should be easily clickable, just as before.

I think the tooltip should be over the button (just like the other buttons in the description window) and not left from it.
The arrow of the tooltop points up in this case, but It should point down and the box should appear over it.
Why the button freaks out when hovering, I have no idea...

**Screenshots**
I hope this illustrates the problem:
![insert_attachment2](https://user-images.githubusercontent.com/431045/100190752-366b5f80-2eef-11eb-96a5-f23c31841a43.gif)

**Client details:**
 - OS: W10, Centos 7.8
 - Browser: Chrome, Firefox
 - Version 87, 83
 - Device: Desktop

NC 19.0.5 / Deck 1.2.2

juliushaertl:
Yes, something seems odd there with the click target.

juliushaertl:
I cannot reproduce this anymore when testing with #2638

reox:
@juliushaertl is the change already in v1.2.3? I just tested in there and I get the same behavior.

juliushaertl:
Ok, now i can indeed reproduce it with Firefox, strangely it didn't happen on master for me. I'm not entirely sure why that happens yet, the button seems to span the full size but somehow the tooltip is toggled on and off when hovering the lower third.

reox:
I just found out, that this only affects the last entry in the list. If you have several attachments, it works fine for all but the last.

fm-sys:
I have found out something new: It is only happening when the browser is in full-screen. And it is a cross browser bug, as it happens at least for Firefox and Chromium Edge...

reox:
[...]

PR #3612, opened 2022-02-28
Make insert attachment buttom easy to click

luka-nextcloud:

* Resolves: #2618
* Target version: master 

### TODO

- [ ] ...

### Checklist


src/components/card/Description.vue
```diff
@@ -252,6 +252,7 @@ export default {
 		flex-shrink: 1;
 		overflow: scroll;
 		max-height: 50vh;
+		min-height: 50vh;
```
juliushaertl, line 255: I think it might even be better if we just remove any height specification and also the overflow scroll, as currently we have two scrollbars with that:

Removing the overflow also seems to fix the tooltip flickering:

juliushaertl (approved):

PR #3625, opened 2022-03-09
Fix: Check all circle shares for permissions

bink:
* Target version: master 

### Summary

Fixes a bug that appears when a board is shared with multiple circles, causing calendars not to work because of a permission exception.
If a board is shared with a circle, the \OCA\Deck\Service\PermissionService::userCan method would only return the permissions for the first circle and not evaluate other permissions like groups or other circles. This is fixed by only returning true if the permissions are actually present and continuing the check when they are not.

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/lib/Service/PermissionService.php
+++ b/lib/Service/PermissionService.php
@@ -213,7 +213,9 @@ class PermissionService {
 
 			if ($this->circlesService->isCirclesEnabled() && $acl->getType() === Acl::PERMISSION_TYPE_CIRCLE) {
 				try {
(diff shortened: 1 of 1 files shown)


juliushaertl:
Thanks for your pull request and good catch on the bug. 

Could you have a look at the failing DCO check? https://github.com/nextcloud/deck/pull/3625/checks?check_run_id=5639445346 should give you more details about that. Otherwise that is good to get merged from my side 👍

juliushaertl (approved):

juliushaertl:
Let's get this in without the DCO since it is a very minor change code wise.

Issue #3649, opened 2022-03-17
Long activity subject parameters may cause errors


juliushaertl:
If activity subject parameters are too long there may be an undefined array index call on https://github.com/nextcloud/deck/blob/51ae25d6a9ffbe9d77ad4f2a670b2b82580fca4b/lib/Activity/ActivityManager.php#L256 for other activity subjects, like assigning a label. The logic to cut off the description in before/after/card should only apply if those are present

```
 "message": "Undefined index: after at /var/www/html/apps/deck/lib/Activity/ActivityManager.php#256",
  "userAgent": "Mozilla/5.0 (X11; Fedora; Linux x86_64; rv:98.0) Gecko/20100101 Firefox/98.0",
  "version": "23.0.3.1",
  "exception": {
    "Exception": "Error",
    "Message": "Undefined index: after at /var/www/html/apps/deck/lib/Activity/ActivityManager.php#256",
    "Code": 0,
    "Trace": [
      {
        "file": "/var/www/html/apps/deck/lib/Activity/ActivityManager.php",
        "line": 256,
        "function": "onError",
        "class": "OC\\Log\\ErrorHandler",
        "type": "::"
      },
      {
        "file": "/var/www/html/apps/deck/lib/Service/CardService.php",
        "line": 571,
        "function": "triggerEvent",
        "class": "OCA\\Deck\\Activity\\ActivityManager",
        "type": "->"
      },
      {
        "file": "/var/www/html/apps/deck/lib/Controller/CardController.php",
        "line": 143,
        "function": "assignLabel",
        "class": "OCA\\Deck\\Service\\CardService",
        "type": "->"
      },
      {
        "file": "/var/www/html/lib/private/AppFramework/Http/Dispatcher.php",
        "line": 217,
        "function": "assignLabel",
        "class": "OCA\\Deck\\Controller\\CardController",
        "type": "->"
      },
      {
        "file": "/var/www/html/lib/private/AppFramework/Http/Dispatcher.php",
        "line": 126,
        "function": "executeController",
        "class": "OC\\AppFramework\\Http\\Dispatcher",
        "type": "->"
      },
      {
        "file": "/var/www/html/lib/private/AppFramework/App.php",
        "line": 157,
        "function": "dispatch",
        "class": "OC\\AppFramework\\Http\\Dispatcher",
        "type": "->"
      },
      {
        "file": "/var/www/html/lib/private/Route/Router.php",
        "line": 302,
        "function": "main",
[...]

PR #3650, opened 2022-03-17
Handle description shortening more gracefully

juliushaertl:
- Avoid trying to cut of fields that are not there are reported in #3649
- Move the logic to the actual event creating method
- Add some more test cases for the scenario and try to improve the related ActivityManagerTest cases a bit

### DIFF
--- a/lib/Activity/ActivityManager.php
+++ b/lib/Activity/ActivityManager.php
@@ -31,7 +31,6 @@ use OCA\Deck\Db\Acl;
 use OCA\Deck\Db\AclMapper;
 use OCA\Deck\Db\Assignment;
 use OCA\Deck\Db\Attachment;
-use OCA\Deck\Db\AttachmentMapper;
 use OCA\Deck\Db\Board;
 use OCA\Deck\Db\BoardMapper;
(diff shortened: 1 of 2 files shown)


lib/Activity/ActivityManager.php
```diff
@@ -410,12 +397,31 @@ private function createEvent($objectType, $entity, $subject, $additionalParams =
 
 		$subjectParams['author'] = $author === null ? $this->userId : $author;
 
+		$subjectParams = array_merge($subjectParams, $additionalParams);
+		$json = json_encode($subjectParams);
+		if (mb_strlen($json) > 4000) {
```
vinicius73, line 402: Can we use more describable variables? 

Like?

```php
const MAX_SUBJECT_LENGTH = 4000;
const MAX_SUBJECT_CONTENT_LENGTH = 2000;

// ---

if (mb_strlen($json) > self::MAX_SUBJECT_LENGTH) { ...
```

juliushaertl: Sure :) Added.

vinicius73 (approved):

PR #3664, opened 2022-03-22
[stable23] Transfer ownership

juliusknorr:
Backport of #2496 

- [x] Drop PHP 7.2 from CI/Composer as it is not supported on 23 server
- [x] Adjust cases where the query builder is used on master but not stable23
- [x] Manual test of occ and UI


luka-nextcloud:
Looks good to me! 👍

PR #3665, opened 2022-03-22
[stable22] Transfer ownership

juliushaertl:
Backport of https://github.com/nextcloud/deck/pull/2496

- [x] Drop PHP 7.2 from CI/Composer as it is not supported on 22 server
- [x] Adjust cases where the query builder is used on master but not stable22
- [x] Manual test of occ and UI

### DIFF
--- a/src/components/board/SharingTabSidebar.vue
+++ b/src/components/board/SharingTabSidebar.vue
@@ -208,13 +208,13 @@ export default {
 					confirmClasses: 'error',
 					cancel: t('deck', 'Cancel'),
 				},
-				async (result) => {
+				async(result) => {
 					if (result) {
 						try {
 							this.isLoading = true
 							await this.$store.dispatch('transferOwnership', {
 								boardId: this.board.id,
-								newOwner
+								newOwner,
 							})
 							const successMessage = t('deck', 'Transfer the board for {user} successfully', { user: newOwner })
 							showSuccess(successMessage)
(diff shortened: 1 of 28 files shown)


luka-nextcloud:
Looks good to me! 👍

luka-nextcloud (approved):

src/components/board/SharingTabSidebar.vue
```diff
@@ -194,6 +198,38 @@ export default {
...
+					cancel: t('deck', 'Cancel'),
+				},
+				async(result) => {
+					if (result) {
+						try {
+							this.isLoading = true
+							await this.$store.dispatch('transferOwnership', {
+								boardId: this.board.id,
+								newOwner,
+							})
```
luka-nextcloud, line 218: `transferOwnership` action has been called with two params: `boardId` and `newOwner`, but the action expects 3 params: `boardId`, `newOwner` and `owner`.

src/store/main.js
```diff
@@ -494,5 +494,11 @@ export default new Vuex.Store({
 					dispatch('loadBoardById', acl.boardId)
 				})
 		},
+		async transferOwnership({ commit }, { boardId, owner, newOwner }) {
+			await axios.put(generateUrl(`apps/deck/boards/${boardId}/transferOwner`), {
+				owner,
```
juliushaertl, line 499: - [ ] Drop owner since it is not used by the backend anyways

PR #3670, opened 2022-03-25
Properly check for the stack AND setting board permissions

juliushaertl:
Fixes a regression from https://github.com/nextcloud/deck/pull/3541 where the the boardId was taken to check for stack permissions. To properly check this we need to extend the permission checks to validate if the board provided in the boardId is also accessible by the user.

Steps to reproduce:
- Rename a list

Before:
403

After:
works again 

Fixes https://github.com/nextcloud/deck/issues/3633

### DIFF
--- a/lib/Service/StackService.php
+++ b/lib/Service/StackService.php
@@ -290,10 +290,13 @@ class StackService {
 			throw new BadRequestException('order must be a number');
 		}
 
-		$this->permissionService->checkPermission($this->stackMapper, $boardId, Acl::PERMISSION_MANAGE);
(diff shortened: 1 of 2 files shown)


nickvergessen:
So it seems to work in the backend and there is no error reported anymore, but I have to reload the page to see my edit?

juliushaertl:
No the edit is applied right away, at least I cannot see any issue when testing here.

nickvergessen:
Peek 2022-04-11 12-55

Just tried again on this PR and it seems to be "one edit behind"

nickvergessen:
But since this PR is php only it should be unrelated

nickvergessen (approved):

Issue #3677, opened 2022-03-30
File count is not increased after sharing


luka-nextcloud:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Describe the bug**
File count does not increase after sharing.

**To Reproduce**
1. Click “Share from Files”
2. Select a file to share
3. File count does not increase

**Expected behavior**
File count should be increased after sharing a new file.

**Screenshots**
![image](https://user-images.githubusercontent.com/89908051/160811033-102b978e-ccb7-4f7c-8cf8-a474aef4d147.png)
![image](https://user-images.githubusercontent.com/89908051/160811087-3624ac0b-4ca5-4b94-952b-8fe877fc179c.png)
![image](https://user-images.githubusercontent.com/89908051/160811131-90ee52b7-b0b8-4ab5-b33b-20dce890f9fd.png)

**Client details:**
 - Device: [e.g. iPhone6, desktop]

<details>
<summary>Server details</summary>

**Operating system**:

**Web server:**

**Database:**

**PHP version:**

**Nextcloud version:** (see Nextcloud admin page)

**Where did you install Nextcloud from:**

**Signing status:**

```
Login as admin user into your Nextcloud and access
http://example.com/index.php/settings/integrity/failed
paste the results here.
```

**List of activated apps:**

```
If you have access to your command line run e.g.:
sudo -u www-data php occ app:list
from within your Nextcloud installation folder
```

**Nextcloud configuration:**

```
If you have access to your command line run e.g.:
sudo -u www-data php occ config:list system
from within your Nextcloud installation folder

or

Insert your config.php content here
Make sure to remove all sensitive content such as passwords. (e.g. database password, passwordsalt, secret, smtp password, …)
```

**Are you using an external user-backend, if yes which one:** LDAP/ActiveDirectory/Webdav/...

</details>

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
```
Insert your Nextcloud log here
```

#### Browser log
```
Insert your browser log here, this could for example include:

a) The javascript console log
b) The network log
c) ...
```

</details>

PR #3682, opened 2022-03-31
Increase file count after sharing

luka-nextcloud:

* Resolves: #3677
* Target version: master 

### Summary
File count does not increase after sharing.

### Checklist


src/components/card/AttachmentList.vue
```diff
@@ -234,6 +234,7 @@ export default {
 						shareWith: '' + this.cardId,
 					}).then(() => {
 						this.$store.dispatch('fetchAttachments', this.cardId)
+						this.$store.commit('cardIncreaseAttachmentCount', this.cardId)
```
vinicius73, line 237: Maybe updating the value direct into `fetchAttachments`  action will be more consistent.

lib/Service/AttachmentService.php
```diff
@@ -137,16 +137,17 @@ public function findAll($cardId, $withDeleted = false) {
...
 	 */
-	public function count($cardId) {
+	public function count($cardId, $update = false) {
 		if (is_numeric($cardId) === false) {
 			throw new BadRequestException('card id must be a number');
 		}
 
 		$count = $this->cache->get('card-' . $cardId);
-		if (!$count) {
+		if ($update OR !$count) {
```
juliushaertl, line 150:
```suggestion
		if ($update || !$count) {
```

Just to stick to our code style ;)

lib/Sharing/DeckShareProvider.php
```diff
@@ -75,16 +76,29 @@ class DeckShareProvider implements \OCP\Share\IShareProvider {
...
+		CardMapper $cardMapper, 
+		AttachmentService $attachmentService,
+		PermissionService $permissionService, 
+		IL10N $l
+	) {
 		$this->dbConnection = $connection;
 		$this->shareManager = $shareManager;
 		$this->boardMapper = $boardMapper;
 		$this->cardMapper = $cardMapper;
+		$this->attachmentService = $attachmentService;
```
juliushaertl, line 99: Injecting this service seems to cause a cyclic dependency and therefore crashing the php process on the test runs

luka-nextcloud:
@juliushaertl Do we need to add a test for it now?

juliushaertl:
If existing tests pass that should be fine for the backend side I'd say with the integration tests in https://github.com/nextcloud/deck/blob/master/tests/integration/features/sharing.feature

For the frontend I've started collecting the case in https://github.com/nextcloud/deck/issues/3768 so we can add it after the cypress parts are merged.

juliushaertl:
Seems tests are not happy yet unfortunately ;)

juliushaertl (approved):

Issue #3679, opened 2022-03-30
Duplicating a board copies its lists, but not the card


luka-nextcloud:
**Describe the bug**
Duplicating a board copies its lists, but not the card.

**To Reproduce**
No card is cloned when “Clone Board” is used.

**Expected behavior**
Cards should also be cloned when “Clone Board” is used.

q-wertz:
I thought this is what is wanted by copying the board. To just get the "frame" of the board. It would be nice if this behavior could also be somehow kept and duplicating the cards is an optional feature. (Cannot really think of a scenario where I want duplicated cards in different boards and then have the same task two times to do :sweat_smile: but apparently there is)

juliushaertl:
Closing as a duplicate of https://github.com/nextcloud/deck/issues/2797

PR #3691, opened 2022-04-05
Clone cards while cloning boards

luka-nextcloud:

* Resolves: #3679
* Target version: master 

### Summary
Duplicating a board copies its lists, but not the card.

### Checklist

### DIFF
--- a/lib/Service/BoardService.php
+++ b/lib/Service/BoardService.php
@@ -47,9 +47,11 @@ use OCP\EventDispatcher\IEventDispatcher;
 use OCP\IConfig;
 use OCP\IGroupManager;
 use OCP\IL10N;
+use OCA\Deck\Db\Card;
 use OCA\Deck\Db\Board;
 use OCA\Deck\Db\BoardMapper;
 use OCA\Deck\Db\LabelMapper;
+use OCA\Deck\Event\CardCreatedEvent;
 use OCP\IUserManager;
 use OCA\Deck\BadRequestException;
 use OCP\IURLGenerator;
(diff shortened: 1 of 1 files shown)


lib/Service/BoardService.php
```diff
@@ -673,15 +675,39 @@ public function clone($id, $userId) {
...
+		foreach ($cards as $card) {
+			$newCard = new Card();
+			$newCard->setTitle($card->getTitle());
+			$newCard->setStackId($insertedStack->getId());
+			$newCard->setType($card->getType());
+			$newCard->setOrder($card->getOrder());
+			$newCard->setOwner($card->getOwner());
+			$newCard->setDescription($card->getDescription());
+			$newCard->setDuedate($card->getDuedate());
+			$newCard = $this->cardMapper->insert($newCard);
```
vinicius73, line 701: We can isolate this process in a method inside `Card` entity.

```suggestion
			$newCard = new Card();
			$newCard->setFromCard($card);
			$newCard->setStackId($insertedStack->getId());
			$newCard = $this->cardMapper->insert($newCard);
```
or
```suggestion
			$newCard = Card::copyFrom($card);
			$newCard->setStackId($insertedStack->getId());
			$newCard = $this->cardMapper->insert($newCard);
```

luka-nextcloud: Yes. But Card is a RelationalEntity. Only accessors and mutators should be defined in it I think. I wanted to add it in CardService but CardService is already using BoardService so there is a dependency issue. What I think is to add a new layer or put it into CardMapper (but mappers are meant to be the ORM layer I think).

vinicius73: I see, you are right
Looks like we need to think more about it to create a wise solution.

luka-nextcloud: I think we can put the logic in a function just in case we need to isolate later on in a class.

juliushaertl:
As discussed in the call, lets:
- [ ] Add a parameter to make the cloning of cards option in the backend
- [ ] Trigger a dialog on the clone action in the frontend
- [ ] Let the user decide if cards should be cloned as well or not (to still cover the use case of using a board as template)
- [ ] While the clone request is running show a nice loading/progress indicator
- [ ] Once done, show a done message and option to close or go to the cloned board

Issue #3678, opened 2022-03-30
Attachment icon is not shown on the archived cards


luka-nextcloud:
**Describe the bug**
Attachment icon is not shown on the archived cards.

**To Reproduce**
1. Click “Archive all cards” for the list of cards which contain attachment
2. Attachment icon is not shown for the archived list

**Expected behavior**
Attachment icon should be shown on the archived cards.

**Screenshots**
![image](https://user-images.githubusercontent.com/89908051/160812034-9f1d54dd-bad6-4108-a4c1-3d4cf64749ae.png)
![image](https://user-images.githubusercontent.com/89908051/160812070-07df4a3e-d856-4309-b394-6075202027ae.png)

PR #3692, opened 2022-04-05
Fix hidden attachment icon on archived cards

luka-nextcloud:

* Resolves: #3678
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/lib/Service/StackService.php
+++ b/lib/Service/StackService.php
@@ -181,6 +181,7 @@ class StackService {
 				if (array_key_exists($card->id, $labels)) {
 					$cards[$cardIndex]->setLabels($labels[$card->id]);
 				}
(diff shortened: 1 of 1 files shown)


lib/Service/StackService.php
```diff
@@ -181,6 +181,7 @@ public function findAllArchived($boardId) {
 				if (array_key_exists($card->id, $labels)) {
 					$cards[$cardIndex]->setLabels($labels[$card->id]);
 				}
+				$cards[$cardIndex]->setAttachmentCount($this->attachmentService->count($card->getId()));
```
vinicius73, line 184: Looks good, though I guess it will be good update tests.
https://github.com/nextcloud/deck/blob/master/tests/unit/Service/StackServiceTest.php#L123

Issue #3229, opened 2021-08-04
Rest-API: First call with different structure


desperateCoder:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Describe the bug**
#### TL;DR: 
With a fresh Deck installation, if you call the get boards endpoint, the very first call is differently structured than described in the API documentation.

#### Details:
Thanks @Aruscha for finding and investigating this issue.
He installed the Deck Android app and just connected the account as usual. With the first sync, the app crashes because of invalid data from the server. After several tests we found out that the first response to the boards call looks like this:

``` json
{
  "1": {
    "title": "Persönlich",
    "owner": {
      "primaryKey": "admin",
      "uid": "admin",
      "displayname": "admin",
      "type": 0
    },
    "color": "0087C5",
    "archived": false,
    "labels": [
      {
        "title": "Abgeschlossen",
        "color": "31CC7C",
        "boardId": 1,
        "cardId": null,
        "lastModified": 1627986420,
        "id": 1,
        "ETag": "c6bddc9dec21b630436527323bd3e042"
      },
    ],
    "acl": [],
    "permissions": {
      "PERMISSION_READ": true,
      "PERMISSION_EDIT": true,
      "PERMISSION_MANAGE": true,
      "PERMISSION_SHARE": true
    },
    "users": [
      {
        "primaryKey": "admin",
        "uid": "admin",
        "displayname": "admin",
        "type": 0
      }
    ],
    "stacks": [],
    "deletedAt": 0,
    "lastModified": 1627986420,
    "settings": {
      "notify-due": "assigned",
      "calendar": true
    },
    "id": 1,
    "ETag": "c6bddc9dec21b630436527323bd3e042"
  }
}
```
The relevant part is This one:
```json
{
  "1": {
```
While any subsequent call returns the expected structure:
```json
[
  {
    "title": "Pers\u00f6nlich",
    "owner": {
      "primaryKey": "admin",
      "uid": "admin",
      "displayname": "admin",
      "type": 0
    },
    "color": "0087C5",
    "archived": false,
    "labels": [],
    "acl": [],
    "permissions": {
      "PERMISSION_READ": true,
      "PERMISSION_EDIT": true,
[...]

PR #3696, opened 2022-04-07
Setup cypress test

luka-nextcloud:

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist


cypress/integration/boardTitle.js
```diff
@@ -0,0 +1,12 @@
+describe('Files default view', function() {
+	beforeEach(function() {
+		cy.login('admin', 'admin')
+	})
+
+	it('See the board title', function() {
+		cy.get('.board-title h2')
+        .should('have.length', 1)
+        .first()
+        .should('have.text', 'Upcoming cards')
```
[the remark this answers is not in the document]
juliushaertl: Could also add a test case that the Personal board was created for the user automatically and is visible in the sidebar as a good starter.

cypress/support/commands.js
```diff
@@ -0,0 +1,40 @@
...
+	cy.session(user, function () {
+		cy.visit(route)
+		cy.get('input[name=user]').type(user)
+		cy.get('input[name=password]').type(password)
+		cy.get('#submit-wrapper input[type=submit]').click()
+		cy.url().should('include', route)
+	})
+	// in case the session already existed but we are on a different route...
+	cy.visit(route)
+})
```
juliushaertl, line 40: Maybe we add https://github.com/nextcloud/text/blob/master/cypress/support/commands.js#L41-L96 also right away, then the basic test could already use a random user as in https://github.com/nextcloud/text/blob/master/cypress/integration/share.spec.js#L25-L26

juliushaertl:
- [ ] Add GitHub action to run cypress tests automatically on pull requests
- [ ] Add some tests for basic functions
  - [ ] After login check if the "Personal" board is there with the example data
  - [ ] Create a board, open it, check that initial columns are there
  - [ ] Add a card a board

After that we could get that in as a start and then extend iteratively I'd say :)

cypress/integration/boardTitle.js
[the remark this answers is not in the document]
luka-nextcloud: BTW, what does Personal board mean? Is it the board that's not shared with anyone else?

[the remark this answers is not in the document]
juliushaertl: That is basically just the name of the example board that gets added on first use for each user. It has 3 columns and some cards to show users how they can get started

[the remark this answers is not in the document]
luka-nextcloud: Those users created by occ didn't have any default board.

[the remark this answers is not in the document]
juliushaertl: Strange, it should get created once any user loads a deck route for the first time https://github.com/nextcloud/deck/blob/52febb396cc79fd162efdd87550947f35d3e3f84/lib/Middleware/DefaultBoardMiddleware.php#L53-L61

At least here it seems to work fine also for new users created with occ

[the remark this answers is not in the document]
luka-nextcloud: Here's a screenshot:
image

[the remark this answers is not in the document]
stefan-niedermann: Just a guess, but might be related to https://github.com/nextcloud/deck/issues/3229 🤷

.github/workflows/cypress.yml
```diff
@@ -0,0 +1,111 @@
...
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        node-version: [14.x]
+        # containers: [1, 2, 3]
+        php-versions: [ '7.4' ]
+        databases: [ 'sqlite' ]
+        server-versions: [ 'master' ]
```
vinicius73, line 26: **Question:** 
We are trigging this test in `stable*` branches too.
But the server version always is master, there is a way to test stable branches with the stable version of the server?

cypress/support/commands.js
```diff
@@ -0,0 +1,115 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+ 
+ const url = Cypress.config('baseUrl').replace(/\/index.php\/?$/g, '')
+ Cypress.env('baseUrl', url)
+ 
+ Cypress.Commands.add('login', (user, password, route = '/apps/deck/') => {
+	let session = `${user}-${Date.now()}`
```
vinicius73, line 27: **Question:**
Why did you use `Date.now()`?
It will ways generate a new session instead to reuse the previous session

.github/workflows/cypress.yml
```diff
@@ -0,0 +1,112 @@
...
+          php -f index.php
+          php -S 0.0.0.0:8081 &
+          export OC_PASS=1234561
+          php occ user:add --password-from-env user1
+          php occ user:add --password-from-env user2
+          php occ app:enable deck
+          php occ app:list
+          composer install
+          cd apps/deck
+          composer install
```
juliushaertl, line 81:
```suggestion
          composer install
          npm ci
          npm run build
```

It seems to be missing the frontend build steps for deck here in order to run on GitHub actions.

.github/workflows/cypress.yml
```diff
@@ -0,0 +1,114 @@
...
+          npm install
+          npm run build
```
juliushaertl, line 83:
```suggestion
          cd apps/deck
          composer install
          npm ci
          npm run build
```

.github/workflows/cypress.yml
```diff
@@ -0,0 +1,114 @@
...
+          cd ../../
+          curl -v http://localhost:8081/index.php/login
+          cat data/nextcloud.log
```
juliushaertl, line 86:
```suggestion
```

juliushaertl: We dump the log anyways further down

cypress/integration/cardFeatures.js
```diff
@@ -0,0 +1,38 @@
...
+
+        cy.get('.board .stack').eq(0).within(() => {
+            cy.get('button.action-item.action-item--single.icon-add')
+                .first().click()
+
+            cy.get('.stack__card-add form input#new-stack-input-main')
+                .type(card)
+            cy.get('.stack__card-add form input[type=submit]')
+                .first().click()
+            cy.get('.card').first().contains(card).should('be.visible')
```
juliushaertl, line 35: Maybe additionally assert that the card text contains `Card 1`.

cypress/integration/deckDashboard.js
```diff
@@ -0,0 +1,33 @@
...
+	it('Can show the right title on the dashboard', function() {
+		cy.get('.board-title h2')
+        	.should('have.length', 1).first()
+        	.should('have.text', 'Upcoming cards')
+	})
+
+	it('Can see the default "Personal Board" created for user by default', function () {
+		const defaultBoard = 'Personal'
+
+		cy.get('.app-navigation button.app-navigation-toggle').click()
```
juliushaertl, line 24: Maybe we can increase the screen size in https://github.com/nextcloud/deck/pull/3696/files#diff-db1f8bb89a89d5b089b39a2af3c27ab75092f1844a158eca0a76445d513d9044R4-R5 and avoid this additional step for all tests that affect the left sidebar.

cypress/integration/stackFeatures.js
```diff
@@ -0,0 +1,35 @@
...
+
+    before(function () {
+		cy.nextcloudCreateUser(randUser, password)
+        cy.deckCreateBoard({ user: randUser, password }, board)
+	})
+
+	beforeEach(function() {
+        cy.logout()
+		cy.login(randUser, password)
+	})
```
juliushaertl, line 17: Indentation seems a bit off ;)

cypress/integration/stackFeatures.js
```diff
@@ -0,0 +1,35 @@
...
+            .find('a.app-navigation-entry-link')
+            .first().click({force: true})
+
+        cy.get('#stack-add button').first().click()
+        cy.get('#stack-add form input#new-stack-input-main')
+            .type(stack)
+        cy.get('#stack-add form input[type=submit]')
+            .first().click()
+
+        cy.get('.board .stack').eq(0).contains(stack).should('be.visible')
```
juliushaertl, line 33: Can we also check for the stack title in here?

juliushaertl (commented):
A few minor comments, but seems a good start otherwise 👍

cypress/integration/deckDashboard.js
luka-nextcloud: When we increase the width, we still need to open the sidebar so I think we need to do so before each operation.

cypress/integration/cardFeatures.js
luka-nextcloud: Sorry but I didn't get your point. "contains with should" assertion checks the tile in both cases: https://docs.cypress.io/api/commands/contains.
card's content is 'Card 1' and stack's value is 'Stack 1'

cypress/integration/stackFeatures.js
luka-nextcloud: Sorry but I didn't really get your point. "contains with should" assertion checks the tile in both cases: https://docs.cypress.io/api/commands/contains.
card's content is 'Card 1' and stack's value is 'Stack 1'

juliushaertl: Same here, sorry for the noise

juliushaertl (approved):
LGTM 🚀

PR #3712, opened 2022-04-11
Remove unused argument from transfer ownership

juliushaertl:
The owner is not used in the controller arguments https://github.com/nextcloud/deck/blob/23f0b16a5aca738699bfaa396ff4d505f975b4cf/lib/Controller/BoardController.php#L165 as it is inherited from the board that is being transferred.


juliushaertl:
No backports needed, cherry-picked directly into https://github.com/nextcloud/deck/pull/3665 and https://github.com/nextcloud/deck/pull/3664

PR #3720, opened 2022-04-12
Update `DeckMapper` to extend `QBMapper`

Raudius:
* Resolves: https://github.com/nextcloud/deck/issues/3162
* Target version: master 

### Summary
Updates the mapper implementation of `DeckMapper` away from the deprecated `Mapper` to the newer `QBMapper`, which allows for further integration with the query builder. 

### Checklist

### DIFF
--- a/lib/Db/AttachmentMapper.php
+++ b/lib/Db/AttachmentMapper.php
@@ -60,7 +60,7 @@ class AttachmentMapper extends DeckMapper implements IPermissionMapper {
 	public function find($id) {
 		$qb = $this->db->getQueryBuilder();
 		$qb->select('*')
-			->from('deck_attachment')
+			->from($this->getTableName())
 			->where($qb->expr()->eq('id', $qb->createNamedParameter($id, IQueryBuilder::PARAM_INT)));
 
 		return $this->findEntity($qb);
@@ -77,7 +77,7 @@ class AttachmentMapper extends DeckMapper implements IPermissionMapper {
 	public function findByData($cardId, $data) {
 		$qb = $this->db->getQueryBuilder();
 		$qb->select('*')
-				->from('deck_attachment')
+				->from($this->getTableName())
 				->where($qb->expr()->eq('card_id', $qb->createNamedParameter($cardId, IQueryBuilder::PARAM_INT)))
 				->andWhere($qb->expr()->eq('data', $qb->createNamedParameter($data, IQueryBuilder::PARAM_STR)));
 
@@ -92,7 +92,7 @@ class AttachmentMapper extends DeckMapper implements IPermissionMapper {
 	public function findAll($cardId) {
 		$qb = $this->db->getQueryBuilder();
 		$qb->select('*')
-			->from('deck_attachment')
+			->from($this->getTableName())
 			->where($qb->expr()->eq('card_id', $qb->createNamedParameter($cardId, IQueryBuilder::PARAM_INT)))
 			->andWhere($qb->expr()->eq('deleted_at', $qb->createNamedParameter(0, IQueryBuilder::PARAM_INT)));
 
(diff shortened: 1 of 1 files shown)


lib/Db/AclMapper.php
```diff
@@ -29,22 +29,53 @@
 use OCP\IDBConnection;
 
 class AclMapper extends DeckMapper implements IPermissionMapper {
+
+	/**
+	 * @param IDBConnection $db
```
juliushaertl, line 34: I'd vote for dropping phpdoc comments if methods are typed already ;)

lib/Db/AclMapper.php
```diff
@@ -29,22 +29,53 @@
...
+	 * @return bool
+	 * @throws \OCP\DB\Exception
+	 */
 	public function isOwner($userId, $aclId): bool {
-		$sql = 'SELECT owner FROM `*PREFIX*deck_boards` WHERE `id` IN (SELECT board_id FROM `*PREFIX*deck_board_acl` WHERE id = ?)';
-		$stmt = $this->execute($sql, [$aclId]);
-		$row = $stmt->fetch();
-		return ($row['owner'] === $userId);
+		$qb = $this->db->getQueryBuilder();
+		$qb->select('*')
```
juliushaertl, line 66: Might be very minor optimisation, but I think some databases are able to get the information just from the index when querying just an indexed column like the id of the ACL. Since we only use the rowCount that should be enough.

juliushaertl (changes requested):
Very minor comments inline and CI needs some more love, but very nice to have that moved finally.  🚀

lib/Db/AclMapper.php
Raudius: Removed doc blocks for constructors, but kept them for other signed functions since I think think they are still handy for exception propagation. Any thoughts?

lib/Db/AclMapper.php
Raudius: Nice suggestion, added!

lib/Db/AttachmentMapper.php
```diff
@@ -52,54 +51,43 @@ public function __construct(IDBConnection $db, CardMapper $cardMapper, IUserMana
...
+	 * @param int $id
+	 * @return Attachment
+	 * @throws DoesNotExistException
+	 * @throws MultipleObjectsReturnedException
+	 * @throws \OCP\DB\Exception
 	 */
 	public function find($id) {
 		$qb = $this->db->getQueryBuilder();
 		$qb->select('*')
 			->from('deck_attachment')
```
luka-nextcloud, line 63: I think it'd be better to use here: `->from($this->getTableName())` instead of the name of the table.

lib/Db/LabelMapper.php
```diff
@@ -26,48 +26,113 @@
...
+	 * @return Label[]
+	 * @throws \OCP\DB\Exception
+	 */
+	public function findAssignedLabelsForBoard($boardId, $limit = null, $offset = null): array {
+		$qb = $this->db->getQueryBuilder();
+		$qb->select('l.id as id', 'l.title as title', 'l.color as color')
+			->selectAlias('c.id', 'card_id')
+			->from($this->getTableName(), 'l')
+			->innerJoin('l', 'deck_assigned_labels', 'al', 'al.label_id = l.id')
+			->innerJoin('l', 'deck_cards', 'c', 'al.card_id = c.id')
```
luka-nextcloud, line 99: It seems there is a misspelling here 🙂. Instead it should be `->innerJoin('al', 'deck_cards', 'c', 'al.card_id = c.id')` I think.

Raudius: Maybe I misunderstood the phpDoc from `innerJoin` but my interpretation is that in the first argument you need to provide the alias for the `from` parameter.

> ```@param string $fromAlias The alias that points to a from clause.```

lib/Db/AttachmentMapper.php
Raudius: Nice suggestion, added! (also for the rest of finders in this class)

juliushaertl (approved):

PR #3740, opened 2022-04-20
[stable22] Cache most frequent queries

juliushaertl:
Backport of #3444

- Cache card to board id relation
- Avoid fetching board details multiple times
- Keep API results the same as before


juliushaertl:
Let's skip this backport

Issue #2906, opened 2021-03-22
IDs for each card


JS-E:
Hi there,

Love deck, super useful for a little project me and my family have going. Would be awesome to have an easily viewable id for each card so we can refer to each item when we discuss it.

If someone can point me in the right direction i'll happily add it myself.

Thanks very much and great work

Josh

stefan-niedermann:
Hey @JS-E 👋 

the URL (see the address bar in your browser) already contains a unique ID. It looks something like this:

```
https://example.com/index.php/apps/deck/#/board/106/card/463
```

where `463` is the unique ID of the card.

Out of personal interest: I am curious what your workflow is and why the titles of your cards aren't unique? How are you using Deck that you can't talk like "Hey, the card 'hire craftman for repairing the roof' is still in the 'blocked' list. Do you have any news?"?

To be honest, i don't see an use case for the majority of the users which justifies the introdiction of an artifical new ID in the UI and would therefore propose to close this issue in favor of https://github.com/nextcloud/deck/issues/1833
This will bring user defined fields and implicitely provide support for your scenario.

JS-E:
Ahh okay i can see that now, thanks for pointing that out 👍 The issue i'm having is that it might be more awkward for less technical people to know the id is in the address bar, and having that shown per card would be kinda cool in my opinion, or perhaps just a little setting to hide show id in the card view. I do understand your justification though, appreciate you clearing that up.

In terms of our use case, it's more for tracking boxes of items, so having an id would be quicker than going "Oh the box with toys and electronics" but we can just id them ourselves on the card title. 

That #1833 does seem like that will help with potential so i agree. Will that potentially have an autogenerated field for example? So incremented integer or something along those lines?

stefan-niedermann:
> perhaps just a little setting to hide show id in the card view

Yeah 😄 just [a little setting](https://user-images.githubusercontent.com/4741199/111898640-a87c2d00-8a27-11eb-98ae-ecf46f46d607.png) 😄 

> Will that potentially have an autogenerated field for example? So incremented integer or something along those lines?

I doubt that, at least for the first iteration, though I am only the Android guy, so i really can't tell you...

JS-E:
> Yeah 😄 just a little setting
haha fair point, mind you there are ways of making it subtle, unlike wget it seems.

> I am only the Android guy, so i really can't tell you...
Ahh cool, nice work with the app. I'd like to contribute at some stage

pedro-nonfree:
[...]

PR #3757, opened 2022-04-26
feat: #2906 add card ID badge

oneWaveAdrian:
- adds vscode settings file to gitingore
- adds new badge for card ID
- adds card ID to board filter
- adds settings to disable card ID badge

* Resolves: #2906 
* Target version: master 

### Summary
Adds ID Badge to top left corner of card.
image

### Checklist

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -79,7 +79,7 @@ jobs:
 
       - name: Set up PHPUnit
         working-directory: apps/${{ env.APP_NAME }}
-        run: composer i
+        run: composer i --no-dev
 
       - name: Set up Nextcloud
         run: |
--- a/lib/Activity/ActivityManager.php
+++ b/lib/Activity/ActivityManager.php
@@ -367,7 +367,15 @@ class ActivityManager {
 			case self::SUBJECT_CARD_USER_ASSIGN:
 			case self::SUBJECT_CARD_USER_UNASSIGN:
 				$subjectParams = $this->findDetailsForCard($entity->getId(), $subject);
-				break;
+
+				if (isset($additionalParams['after']) && $additionalParams['after'] instanceof \DateTimeInterface) {
+					$additionalParams['after'] = $additionalParams['after']->format('c');
+				}
+				if (isset($additionalParams['before']) && $additionalParams['before'] instanceof \DateTimeInterface) {
+					$additionalParams['before'] = $additionalParams['before']->format('c');
+				}
+
+			break;
 			case self::SUBJECT_ATTACHMENT_CREATE:
 			case self::SUBJECT_ATTACHMENT_UPDATE:
 			case self::SUBJECT_ATTACHMENT_DELETE:
--- a/lib/Activity/DeckProvider.php
+++ b/lib/Activity/DeckProvider.php
@@ -312,12 +312,19 @@ class DeckProvider implements IProvider {
 			$userLanguage = $this->config->getUserValue($event->getAuthor(), 'core', 'lang', $this->l10nFactory->findLanguage());
 			$userLocale = $this->config->getUserValue($event->getAuthor(), 'core', 'locale', $this->l10nFactory->findLocale());
 			$l10n = $this->l10nFactory->get('deck', $userLanguage, $userLocale);
-			$date = new \DateTime($subjectParams['after']);
-			$date->setTimezone(new \DateTimeZone(\date_default_timezone_get()));
+			if (is_array($subjectParams['after'])) {
+				// Unluckily there was a time when we stored jsonSerialized date objects in the database
+				// Broken in 1.8.0 and fixed again in 1.8.1
+				$date = new \DateTime($subjectParams['after']['date']);
+				$date->setTimezone(new \DateTimeZone(\date_default_timezone_get()));
+			} else {
+				$date = new \DateTime($subjectParams['after']);
+				$date->setTimezone(new \DateTimeZone(\date_default_timezone_get()));
+			}
 			$params['after'] = [
 				'type' => 'highlight',
 				'id' => 'dt:' . $subjectParams['after'],
-				'name' => $l10n->l('datetime', $date)
+				'name' => $l10n->l('datetime', $date),
 			];
 		}
 		return $params;
--- a/tests/integration/run.sh
+++ b/tests/integration/run.sh
@@ -19,8 +19,7 @@ else
 	exit 1
 fi
 
-composer install --no-dev
-composer dump-autoload
+composer install
 
 # avoid port collision on jenkins - use $EXECUTOR_NUMBER
 if [ -z "$EXECUTOR_NUMBER" ]; then


oneWaveAdrian:
@juliushaertl can this be marked as to review please?

juliushaertl:
I'd say we can add it to the list of card badges on the bottom left, but would also appreciate feedback from @nimishavijay and @jancborchardt about that and how we should handle this in terms of the extra setting for showing/hiding the number. Note that the id is unique across users/boards so on larger instances this might quickly become a number with >10 digits

oneWaveAdrian:
@juliushaertl that is why I placed it above the title, already tested with up to 12 digits and it works without impact.
@stefan-niedermann I'd appreciate design input, see this as a technical draft, happy to edit to fit in. I felt an additional badge took too much space in the grand scheme, so I looked at  how Jira and Trello do it.
@juliushaertl setting for hiding this feature is already implemented, I am unsure of what you mean.

raimund-schluessler:
> @juliushaertl that is why I placed it above the title, already tested with up to 12 digits and it works without impact.

How does it help a user to quickly see the 12 digit id of a card? That's not really a number you memorize for refering to a card in a conversation, no?

nimishavijay:
Super nice work! :)

> I'd say we can add it to the list of card badges on the bottom left

This seems like a good idea! It could go in the beginning of the badges, something like 
image


> ID = OWSWS-422

This kind of ID seems really technical. Deck is usually compared with Trello, so a simpler ID like that seems more friendly. Currently all cards have unique IDs across all boards. How about cards within one board have unique IDs? For eg. 
Right now:
- Board1 has cards with ID 1, 2, 3, 4, 5
- Board2 has cards with ID 6, 7, 8

Instead we could do 
- Board1 has cards with ID 1, 2, 3, 4, 5
- Board2 has cards with ID 1, 2, 3

This wouldn't be the same as the ID in the URL, but more like a reference number for each card. It also makes sense that when you create a new board and add a card it shows `#1`. What do you think?

> So in everyones best interest we could define that it is disabled by default, so it doesn't impact anyone's current setup and those who need it can enable it.

Agreed, there is an [addon](https://trello.com/power-ups/59c3d177178a761767b49278/card-numbers-by-reenhanced) in Trello for this and it seems quite popular too, so it is okay to have a setting for this since it's a feature that would be useful to power users but maybe not so much to just everyday users.

@jancborchardt any thoughts about this?

oneWaveAdrian:
4 days later - How are we going to proceed here?

oneWaveAdrian:
> However, there’s the big confusion about the usefulness of the ID as @juliushaertl mentioned:
> 
> > Note that the id is unique across users/boards

@jancborchardt I get the idea behind this: you can move a card between boards and external links to the card still work. So the card can be found even if, let's say it moves from the "User feedback" board to the "In development" board.

Hence my suggestion to add a local ID (as @nimishavijay suggested starting with `#1`) and still have the global ID. Therefore nothing it will remain backwards compatible and nothing changes for those users who won't need it.

@nimishavijay you mentioned placing it at the beginning of the badges. What happens if an ID on a boarddoes get up to let's say `#9999`? Will that still work design wise/is there enough space?

nimishavijay:
> What happens if an ID on a boarddoes get up to let's say #9999? Will that still work design wise/is there enough space?

Well, in the worst case when the entire bottom row is filled out there would be the card ID, description badge, attachment badge, assignees, and 3dot menu. If there is no space to show all these, we could start hiding the description and attachment badges, and if the ID is so big that there's _really_ no space even after that it should be okay if we add a row below the ID and move the badges, assignees and 3dot menu there (although if the IDs are that large it would not be of much use anyway like @raimund-schluessler mentioned)

oneWaveAdrian:
Thanks for the design feedback @nimishavijay . So, @juliushaertl, @stefan-niedermann, @raimund-schluessler  what are the next steps here in your opinion? Can the card Model be altered to accommodate a local ID? This is more than just a frontend change, but rather a change in card database architecture, therefore I would like some discourse/agreements before starting to alter things and then the change being rejected.

stefan-niedermann:
@oneWaveAdrian I am not in the position to "make a decision". I am a volunteer contributor and maintainer of the Deck Android app. I can give input and advice, but every else is up to @juliushaertl

oneWaveAdrian:
As another PR has popped up trying to provide the same feature, this seems to be something a lot of users need/want, can we move on with this topic? Even if you decide to close it, leaving it hanging mid-air forever is not an option to work with your community @juliushaertl @nimishavijay @jancborchardt

juliushaertl:
Thinking about this a bit more I'd say even the global unique id is fine, as we also expose it for files with internal links. This might especially be useful if we introduce automatic references as GitHub has them with https://github.com/nextcloud/server/issues/31667, so I'd be fine to merge this with the id badge being moved to the bottom row.

That being said, from a user perspective I could see that a local id might still be useful, but I'd consider that a follow up feature. As you mention it would require a separate field being added to the database to expose a unique id per card and would require careful handling for cases where cards are moved across boards.

lib/Service/ConfigService.php
```diff
@@ -124,6 +130,17 @@ public function isCardDetailsInModal(int $boardId = null): bool {
 		return (bool)$this->config->getUserValue($this->getUserId(), Application::APP_ID, 'board:' . $boardId . ':cardDetailsInModal', $defaultState);
 	}
 
+	public function isCardIdBadgeEnabled(int $boardId  = null): bool 
+	{
+		if ($this->getUserId() === null) {
+			return false;
```
juliushaertl, line 136: This may not be used anyways, but we could take the `$appConfigState` so the admin can set this in case we introduce public shares.

src/components/cards/badges/CardId.vue
```diff
@@ -0,0 +1,58 @@
...
+@import './../../../css/variables';
+	.cardid {
+    position: absolute;
+    left: 0px;
+    top: 0px;
+		font-size: 75%;
+		margin-left: $card-padding;
+    margin-bottom: $card-padding;
+		vertical-align: text-top;
+    opacity: .75;
```
juliushaertl, line 56: Indentation looks off ;)

juliushaertl (commented):
Looks good besides some small inline comments and the design feedback discussed earlier.

lib/Service/ConfigService.php
oneWaveAdrian: @juliushaertl I am unsure of what you need me to change when you say "take the `$appConfigState`". Do you have an example?

src/components/cards/badges/CardId.vue
oneWaveAdrian: @juliushaertl #tabsVSspaces

oneWaveAdrian:
@juliushaertl Thanks for the feedback. 
Just to be sure:
I am placing the ID in the bottom left as suggested by @nimishavijay, even though this could create visual problems if the ID was too long, correct?

> > I'd say we can add it to the list of card badges on the bottom left
> 
> This seems like a good idea! It could go in the beginning of the badges, something like image

ultrasites:
Is this a release candidate now? I need it :)

lib/Service/ConfigService.php
ultrasites: @juliushaertl it is the same procedure is the other methods. It can be written so, in my opinion

.gitignore
```diff
@@ -9,3 +9,4 @@ tests/.phpunit.result.cache
 vendor/
 .php_cs.cache
 \.idea/
+settings.json
```
ultrasites, line 12: why?

oneWaveAdrian: @ultrasites This is a VSCode settings file containing developer preferences which should not be committed.

lib/Service/ConfigService.php
```diff
@@ -124,6 +130,17 @@ public function isCardDetailsInModal(int $boardId = null): bool {
 		return (bool)$this->config->getUserValue($this->getUserId(), Application::APP_ID, 'board:' . $boardId . ':cardDetailsInModal', $defaultState);
 	}
 
+	public function isCardIdBadgeEnabled(int $boardId  = null): bool
```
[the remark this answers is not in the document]
oneWaveAdrian: Correct, resolved in https://github.com/oneWaveAdrian/deck/commit/504e6a81a81b18baacc6eb5c8d1d037e652b9a8d

lib/Service/ConfigService.php
oneWaveAdrian: @juliushaertl @ultrasites either way, if someone could explain what should be changed, I'm happy to do so, otherwise pls resolve this.

juliushaertl:
> I am placing the ID in the bottom left as suggested by @nimishavijay, even though this could create visual problems if the ID was too long, correct?

Yes, let's go with that I'd say. The code seems good otherwise, so we could merge this if you move the CardId to the CardBadges component. 👍

oneWaveAdrian:
@juliushaertl accidentally closed this PR in an attempt to rebase... can you reopen it please? I usually use gitlab and got confused here...

juliushaertl:
I can't it seems as the branch does not contain any other commits compared to master. Maybe you can reset the branch to one of the previous commit ids: https://github.com/nextcloud/deck/commit/bfb35a506779d6d87ef811e3bb11965f9e3a1e66

ultrasites:
@juliushaertl @oneWaveAdrian alternatively could you open a new pr? i want to approve it :+1:

oneWaveAdrian:
I did reset and force push but it seems the Pr somehow didn't get it. Will open a new one.

ultrasites:
> I did reset and force push but it seems the Pr somehow didn't get it. Will open a new one.

But the pr/repo owner could reopen it via github, or?

juliushaertl:
Not if the source branch has no commits:

Screenshot 2022-10-26 at 13 44 03

oneWaveAdrian:
@juliushaertl @ultrasites new PR as suggested: https://github.com/nextcloud/deck/pull/4156/
It is weird, the reason this happened was that I clicked on "Synch Fork" in my branch repo and then it deleted all my edits as it reset to master apparently... will do it through terminal again next time, thought I'd try this github feature to save time, oh well...

Issue #3711, opened 2022-04-11
Selected text in description is not visible/readable


nickvergessen:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Describe the bug**

**To Reproduce**
1. Create a card
2. Add a description
3. Reload
4. Open the card
5. Go to edit description (also not too easy as the modal width is limited and the button is not visible immediately in 24)
6. Select text (it's grey font on grey background)

**Expected behavior**
Still readable text

**Screenshots**
Grey is selection, red is me censoring
The color is not exactly the same that's why making it red was necessary, but it's totally not readable.

![Bildschirmfoto von 2022-04-11 13-04-46](https://user-images.githubusercontent.com/213943/162727718-81572ad3-e866-4436-87d2-4790355848fa.png)

**Client details:**
 - OS: Ubuntu
 - Browser Firefox
 - Version 23+master at least (c.nc.c)
 - Device: Desktop PC

juliushaertl:
I think this is actually an issue of the Modal component from `@nextcloud/vue` since I can reproduce that on other dialogs as well, but need to dive a bit more into.

PR #3761, opened 2022-04-28
Fix text selection in dark mode and modal view

juliushaertl:
* Resolves: #3711
* Target version: master 

### Summary

- Ensure that text in the modal is selectable (as it is blocked by the modal components touch handling)
- Fix selection background color in dark mode

### DIFF
--- a/src/components/card/CardSidebar.vue
+++ b/src/components/card/CardSidebar.vue
@@ -225,6 +225,8 @@ export default {
 		padding: 0 14px;
 		max-height: 100%;
 		overflow: initial;
+		user-select: text;
+		-webkit-user-select: text;
 
 		&::v-deep {
 			.app-sidebar-header {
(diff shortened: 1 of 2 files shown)


src/components/card/CardSidebar.vue
```diff
@@ -225,6 +225,9 @@ export default {
 		max-width: calc(100% - #{$modal-padding*2});
 		padding: 0 14px;
 		max-height: 100%;
+		user-select: text;
+		-webkit-user-select: text;
```
luka-nextcloud, line 229: What do you think about the support for other browsers?

-moz-user-select: text;
-webkit-user-select: text; 
-ms-user-select: text; 
user-select: text;

juliushaertl: All recent major browsers should be fine without the prefix except for safari, according to https://caniuse.com/mdn-css_properties_user-select ;)

luka-nextcloud (approved):

Issue #3268, opened 2021-08-22
`CTRL` + `ENTER` to Save Changes on Card Description


SethFalco:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Is your feature request related to a problem? Please describe.**

In many apps doing `CTRL` + `ENTER` acts as a "finished" action to either send, post, or save whatever the user is typing.

For example:
* To send an email in Tutanota.
* Submitting issues, pull requests, and comments on GitHub.
* Commenting on Reddit.
* Tweeting on Twitter.

**Describe the solution you'd like**

After clicking "Edit description" the user can type to edit the description of a card.
I think if a user does `CTRL` + `ENTER` it should act as if the user clicked "View description". (Save and return to the non-edit view.)

**Describe alternatives you've considered**

N/A

**Additional context**

Moved over from https://github.com/stefan-niedermann/nextcloud-deck/issues/1087

stefan-niedermann:
Related to #46

dodedodo:
I think `CTRL` + `S` would make more sense in this case. You're not sending or submitting a message like in a mail client or twitter, but rather saving a description. 

`CTRL` + `ENTER` would make sense for comments.

SethFalco:
I think `CTRL` + `ENTER` makes more sense, since `CTRL` + `S` usually doesn't affect the view.
I'm not asking to only save the content of the description, I'm suggesting this should kick the user out of the "Edit description" view entirely.

This is likely subjective, however. In the end, I'm down for whatever most users are into.

Ben-Ro:
Hey all,
this missing feature annoyed me for some time, so i decided to implement it. 

It uses Keylisteners on `easymde.codemirror`. Once the CMD/CTRL is pressed, it sets `keyExitState = 1` until the key is released. If Enter is pushed while this, it calls `hideEditor()` (just like the Action Button) and removes all Keylisteners.

I just create a Pull Request, i hope the code quality is ok :)

PR #3774, opened 2022-04-30
feat: #3268 CTRL/CMD + ENTER to Save Changes on Card Description

Ben-Ro:
CTRL/CMD + ENTER to Save Changes on Card Description
fixes nextcloud/deck#3268

Signed-off-by: ben <ben@ro.tt>

* Resolves: #3268
* Target version: master 

### Summary
Allows saving and leaving the Card Description Editor via `CTRL`/`CMD` + `ENTER`. It's the same shortcut as with Twitter, Trello, Github and lot's of other Webapps

### TODO

- [ ] ...

### Checklist

 Haven't found any documentation of Keyboard Shortcuts, if there is any place to add this shortcut just let me know :)

### DIFF
--- a/src/components/card/Description.vue
+++ b/src/components/card/Description.vue
@@ -60,6 +60,7 @@
 			ref="markdownEditor"
 			v-model="description"
 			:configs="mdeConfig"
+			@initialized="addKeyListeners"
 			@update:modelValue="updateDescription"
 			@blur="saveDescription" />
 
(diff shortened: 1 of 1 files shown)


src/components/card/Description.vue
```diff
@@ -175,8 +176,33 @@ export default {
 			}
 			this.descriptionEditing = true
 			this.description = this.card.description
+
+			// Has to start after the Editor is fully loaded. This shouldn't take longer than 1/4 second
+			setTimeout(() => {
```
juliushaertl, line 181: To avoid the setTimeout you could also register this on a callback of the initialized event that the VueEasyMDE component emits

juliushaertl (changes requested):
Thanks a lot for the contribution. I've added a small suggestion to avoid the timeout inline, otherwise this looks good to me 👍

src/components/card/Description.vue
Ben-Ro: Thanks for the feedback, i just added a method that's called via `@initialized`. The Code inside the function is exactly the same as in the previous version :)

juliushaertl (approved):

SethFalco:
I'm using Deck 1.8.3 but it seems that `CTRL` + `ENTER` doesn't do what is described in this PR. :thinking: 

Can someone confirm if they have the same issue or if this broke at some point?

Ben-Ro:
> I'm using Deck 1.8.3 but it seems that `CTRL` + `ENTER` doesn't do what is described in this PR. 🤔
> 
> Can someone confirm if they have the same issue or if this broke at some point?

Hey, which OS and Browser are you usign? If you're on a mac the shortcut would be `CMD + ENTER` :)

SethFalco:
> which OS and Browser are you usign

I'm on Debian 11 (Linux), worth noting I use this same shortcut everywhere else (including to post this comment) and it works fine. :sob:

Issue #3472, opened 2021-12-05
Add Custom Fields to Cards


GuyInCorner:
I would like the ability to add custom fields in a card.  Kind of like you you can add a custom field to a contact.  This should probably be at the board level.

I tried the custom properties app but that didn't add anything to the deck.  I see no way to add custom fields.

Example fields I would add to some of my boards if I could:

- value of my card
- time I spent on my card
- a version of SW that my card relates to
-

stefan-niedermann:
Duplicate of https://github.com/nextcloud/deck/issues/1833

Maybe you want to start a Pull Request 😉🚀?

GuyInCorner:
I don't have the knowledge to contribute to this or I would.  Sorry for the duplicate.  Didn't find the other issue.  Will anxiously follow it.

KaiBuskirk:
ADVANCED CUSTOM FIELDS / CARDS
i cant work this out if anyone feels this would or could be GREAT!
being able to add to the card basics would be exceptional...
is there a group or person dedicated to this?
tia,
kai

PR #3775, opened 2022-05-01
Feature/counting cards

bkrith:
* Resolves: # 
* Target version: master 

### Summary

Add value field on deck cards and display sum/avg of each list. Is the idea Counting cards and created for nextGov hackathon 2022. 

### TODO

- [ ] ...

### Checklist


florisvangeel:
This pull request came out of the [NextGov 2022](https://nextgov-hackathon.eu/) hackathon

**Abstract**:
The ability to count cards has been very beneficial in the past, in deck we introduce a sum of custom fields that allow you to rate and score new leads deals and truly collaborate with your teams.

[Presentation slides](https://docs.google.com/presentation/d/1JGZoDvtZXxIQI46TJhbD5y4ryotaW5zj)

[Demo video](https://www.youtube.com/watch?v=W6vkQt-6jLI)

florisvangeel:
**Abstract**
The ability to count cards has been very beneficial in the past, in deck we introduce a sum of custom fields that allow you to rate and score new leads deals and truly collaborate with your teams.

Presentation slides: https://docs.google.com/presentation/d/1JGZoDvtZXxIQI46TJhbD5y4ryotaW5zj
Presentation video: https://www.youtube.com/watch?v=W6vkQt-6jLI

matbgn:
Any news about the merge/review of this cool feature?

juliushaertl:
@bkrith @matbgn Thanks for your contribution and I'm very sorry for taking so long to reply. In general this seems like a nice feature, however the implementation is quite specific to a use case and from the product perspective we would rather have aimed for a more generic way of adding additional card metadata (like in #3472). 

@jancborchardt @nimishavijay I'd like to hear what you think about this. Should we introduce such a specific field and what do you think design wise?

In general the code looks pretty solid from a first look, but there would be a few things that needs to get addressed before this could be merged:

- It should expose this to the API endpoints and also properly document the API changes
- Having some tests would be very nice, for the backend in terms of integration tests and frontend wise some basic e2e tests that ensure that those features continue working in the future
- From my POV the unit is something that should be rather a board-specific setting as for some projects you might measure currency values while for other you might track hours or story points

putt1ck:
Time prefix sounds particularly interesting for projects, as can be used both as reporting "how long did it take" and estimating "how long will this take". So if the unit can be free text (with a low char limit, maybe 4?) then it can be used for many purposes, including ones we haven't thought of -  where limiting it to selection restricts use cases.

Ideally the unit setting would be per list but that might be asking a lot :)

juliushaertl:
> Time prefix sounds particularly interesting for projects, as can be used both as reporting "how long did it take" and estimating "how long will this take". So if the unit can be free text (with a low char limit, maybe 4?) then it can be used for many purposes, including ones we haven't thought of - where limiting it to selection restricts use cases.

That seems like a case where more generic custom fields would be more suitable already for tracking more than one value. Same as for the different units per list, there it could just be different value fields and I feel like having a very specific implementation may then lead to users not making use of it and rather requesting another specific feature set.

putt1ck:
I think the high value part is the running total for a list rather than the bit in the card itself

nimishavijay:
Agreed with @juliushaertl, a generic "Add new property" field would be more flexible. We can allow people to set the label and possibly a type as well (text, numbers, time, date, etc. ) :) If it's numerical field then we can show the sum or average like in this example with currency.

matbgn:
@bkrith Would you be OK to update your MR in this way?

PR #3782, opened 2022-05-04
Add CardDetail serialization model

Raudius:
* Target version: master 

### Summary
Adds `CardDetail` + `BoardSummary` models which centralises serialisation of Card database models. This will be useful for the upcoming changes in the overview page (#2807) which will require attaching additional information to the Card (namely: board name) as it will not need to be added in multiple parts of the code.

### Checklist

### DIFF
--- a/lib/Db/Board.php
+++ b/lib/Db/Board.php
@@ -23,6 +23,14 @@
 
 namespace OCA\Deck\Db;
 
+/**
+ * @method int getId()
+ * @method string getTitle()
+ * @method int getShared()
+ * @method bool getArchived()
+ * @method int getDeletedAt()
+ * @method int getLastModified()
+ */
 class Board extends RelationalEntity {
 	protected $title;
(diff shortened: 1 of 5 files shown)


juliushaertl (approved):
Nice, good approach I think to move the representation logic out of the actual Db entity layer. I'd be fine with the Model naming as well as its a common term here 👍

juliushaertl:
@Raudius Mind to squash the commits a bit (especially the different cs fix ones)?

Raudius:
> But, what do you think about renaming Models to Presenters or something similar @Raudius ?
@vinicius73 

I copied the idea (and name) from the [Collectives](https://gitlab.com/collectivecloud/collectives/-/tree/main/lib/Model) app. So I think I will leave it as `Model` for consistency, at least within the Office apps.

vinicius73 (approved):

PR #3803, opened 2022-05-11
SCSS cleanup

juliushaertl:
Contributes to https://github.com/nextcloud/server/issues/32072

Fixes https://github.com/nextcloud/deck/issues/2225

- [x] Move deck.scss over to plain css
- [x] Get rid of globalstyles.scss
  - [x] Move icons to svg icons 
  - [x] Move print.scss to be imported through web pack

PR #3807, opened 2022-05-11
Move to vue-material-design-icons

juliushaertl:
- Get rid of any server-side scss compilation
- Migrate icons to vue-material-design-icons

Resolves issues after https://github.com/nextcloud/server/pull/32326


juliushaertl:
Whoops, accidentally pushed two branches. Correct one is at https://github.com/nextcloud/deck/pull/3803

PR #3858, opened 2022-06-09
Stable24

orglube:
* Resolves: # 
* Target version: master 

### Summary
I use Deck to manage projects and I love it.

However the relation between deck and tasks is confusing and (to me) quite useless as it is implemented now. Let me explained :

1. each stack in deck currently appears in tasks as a “read only” list that can't be removed nor hidden in tasks : this overloads the task dashboard and is moreover quite useless from a task/deck management point of view. Even worse : DAVX fails to synchonize those deck lists on smartphone which produces useless notifications.

2. on the other hand, It would be really usefull to be able to create todolists **within deck cards** in the same way that they are created in tasks. I tried to create tick boxes in my deck cards or to link existing tasks within cards, but I didn’t manage to do it. I thought that a very easy way to implement this would be to have te possibility to link task lists to cards in the "details" panel : what do you think ? Is there any chance to see this implemented ?

### TODO

Suggestion :

- [ ] add the possibility to link task lists to related project via the "details" panel

### Checklist


juliushaertl:
Please open an issue instead of a pull request, for more details see https://docs.github.com/en/issues/tracking-your-work-with-issues/about-issues

PR #3862, opened 2022-06-10
Inserted required property in the rename list field, to prevent the l…

mstolf:
When creating a list on the Deck, there is a validation requiring the name of the list to be informed

But when renaming a list, it was possible to keep it without a name, so I put the required field validation

* Resolves: #3861
* Target version: master 

### Checklist

### DIFF
--- a/lib/Service/StackService.php
+++ b/lib/Service/StackService.php
@@ -214,7 +214,7 @@ class StackService {
 	 * @throws BadRequestException
 	 */
 	public function create($title, $boardId, $order) {
-		if ($title === false || $title === null) {
+		if ($title === false || $title === null || mb_strlen($title) === 0) {
 			throw new BadRequestException('title must be provided');
 		}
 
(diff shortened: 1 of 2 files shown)


stefan-niedermann:
Usually my opinion is, that this is a *validation*, and validations belong into the backend (at least). Other clients will otherwise still be able to create / rename lists to `blank`.
On the other hand this was a breaking change of the REST-API and should be planned well therefore.

mstolf:
@stefan-niedermann In this case I followed the list creation pattern, from what I saw it was treated on the front-end too, I understand that in this case the validation would be better on the back-end because it would not open room for errors

juliushaertl (approved):
Thanks for the pull request, I pushed a fix to also enforce this on the backend, so LGTM 👍

stefan-niedermann:
@juliushaertl 

1. Will there be a database migration? (e. g. converting <kbd>Blank</kbd> to <kbd>Space</kbd>?)
2. How will the API respond to a request that contains <kbd>Blank</kbd>? `400`? `422`? Any proper error message?

juliushaertl:
@stefan-niedermann The API will return a 400 Bad Request response as if the title was not set at all with the error message from https://github.com/nextcloud/deck/pull/3862/files#diff-8a98def6b971c75496456ec632f1310cd132642aa188861682457694b47b5096R218

PR #3876, opened 2022-06-20
basic notify_push usage with session handling

alangecker:
* Touches: #256
* Target version: master 

### Summary
keeps track of current active sessions on a board and shows them in the UI, including real time updates via notify_push.

screen

it is not very useful in itself, but acts as a base and is probably most of the work towards getting live updates for changes #256 finally, because the server needs to know to whom it sends these updates.

If this step goes through, I'm also happy to work on the actual live updates :)

Part of the code is thankfully taken from @juliushaertl sessions in [text](https://github.com/nextcloud/text)

### Requirement
it needs [notify_push](https://github.com/nextcloud/notify_push) installed and running.

### Checklist


alangecker:
tests and analysis currently still fail with errors, which seem totally unrelated to my changes.
Is `master` broken or do I oversee somthing?


- `ERROR: UndefinedMethod - lib/Service/CirclesService.php:61:21 - Method OCA\Circles\CirclesManager::startSuperSession does not exist`

- ```
  1) OCA\Deck\Db\AssignmentMapperTest::testFind
  OCA\Deck\NoPermissionException: Creating boards has been disabled for your account.
  
  /home/runner/work/deck/deck/apps/deck/lib/Service/BoardService.php:316
  /home/runner/work/deck/deck/apps/deck/tests/integration/database/AssignmentMapperTest.php:98
  /home/runner/work/deck/deck/apps/deck/tests/integration/database/AssignmentMapperTest.php:93
  ```

- ```
  Error: Your lock file does not contain a compatible set of packages. Please run composer update.
  ```

lib/AppInfo/Application.php
```diff
@@ -218,4 +224,26 @@ protected function registerCollaborationResources(IProviderManager $resourceMana
...
+		try {
+			$queue = $container->get(IQueue::class);
+		} catch (\Exception $e) {
+			// most likely notify_push is not installed.
+			return;
+		}
+		
+		$eventDispatcher->addListener(SessionCreatedEvent::class, function (SessionCreatedEvent $event) use ($sessionService, $queue) {
+			$sessionService->notifyAllSessions($queue, $event->getBoardId(), "DeckBoardUpdate", $event->getUserId(), [
+				"id" => $event->getBoardId()
```
juliushaertl, line 239:
```suggestion
				'id' => $event->getBoardId()
```

To fit our coding style but this should be auto-fixable by using php-cs-fixer running `composer run cs:fix`

lib/AppInfo/Application.php
```diff
@@ -218,4 +224,26 @@ protected function registerCollaborationResources(IProviderManager $resourceMana
...
+		$container = $this->getContainer();
+		try {
+			$queue = $container->get(IQueue::class);
+		} catch (\Exception $e) {
+			// most likely notify_push is not installed.
+			return;
+		}
+		
+		$eventDispatcher->addListener(SessionCreatedEvent::class, function (SessionCreatedEvent $event) use ($sessionService, $queue) {
+			$sessionService->notifyAllSessions($queue, $event->getBoardId(), "DeckBoardUpdate", $event->getUserId(), [
```
juliushaertl, line 238: Would be nice to use constants for the events like DeckBoardUpdate and maybe stick to snake case as that seems to be common with https://github.com/nextcloud/notify_push/blob/main/DEVELOPING.md#sending-custom-events: `deck_board_update`

lib/AppInfo/Application.php
```diff
@@ -218,4 +224,26 @@ protected function registerCollaborationResources(IProviderManager $resourceMana
...
+	protected function registerSessionListener(IEventDispatcher $eventDispatcher, SessionService $sessionService): void {
+		$container = $this->getContainer();
+		try {
+			$queue = $container->get(IQueue::class);
+		} catch (\Exception $e) {
+			// most likely notify_push is not installed.
+			return;
+		}
+		
+		$eventDispatcher->addListener(SessionCreatedEvent::class, function (SessionCreatedEvent $event) use ($sessionService, $queue) {
```
juliushaertl, line 237: Maybe this could be moved to a dedicated Listener class that is registered using `$context->registerEventListener()` within the register method.

lib/Controller/SessionController.php
```diff
@@ -0,0 +1,88 @@
...
+	 * notifies the server that the session is still active
+	 * @NoAdminRequired
+	 * @param $boardId
+	 */
+	public function sync($boardId, $token): DataResponse {
+		$this->permissionService->checkPermission($this->boardMapper, $boardId, Acl::PERMISSION_READ);
+		try {
+			$this->sessionService->syncSession($boardId, $token);
+			return new DataResponse([]);
+		} catch (DoesNotExistException $e) {
```
juliushaertl, line 72: Would be good to annotate what gets thrown in the service method.

lib/Event/SessionClosedEvent.php
```diff
@@ -0,0 +1,45 @@
...
+
+namespace OCA\Deck\Event;
+
+use OCP\EventDispatcher\Event;
+
+class SessionClosedEvent extends Event {
+	private $boardId;
+	private $userId;
+	
+	public function __construct($boardId, $userId) {
```
juliushaertl, line 32:
```suggestion
	public function __construct(int $boardId, string $userId) {
```

lib/Event/SessionCreatedEvent.php
```diff
@@ -0,0 +1,46 @@
...
+
+namespace OCA\Deck\Event;
+
+use OCP\EventDispatcher\Event;
+
+class SessionCreatedEvent extends Event {
+	private $boardId;
+	private $userId;
+	
+	public function __construct($boardId, $userId) {
```
juliushaertl, line 33:
```suggestion
	public function __construct(int $boardId, string $userId) {
```

lib/Db/SessionMapper.php
```diff
@@ -0,0 +1,62 @@
+<?php
```
juliushaertl, line 1:
```suggestion
<?php

declare(strict_types=1);
```

Would be good to add that to all new classes.

lib/Event/SessionCreatedEvent.php
juliushaertl, line 33:
```suggestion
	public function __construct(int $boardId, string $userId) {
```

src/components/Controls.vue
```diff
@@ -263,6 +267,9 @@ export default {
 		labelsSorted() {
 			return [...this.board.labels].sort((a, b) => (a.title < b.title) ? -1 : 1)
 		},
+		isNotifyPushEnabled() {
```
juliushaertl, line 270: Can be just a data attribute as there is no need for reactivity and updates.

src/components/board/Board.vue
```diff
@@ -124,13 +126,51 @@ export default {
 		},
 	},
 	watch: {
-		id: 'fetchData',
+		id(newValue, oldValue) {
+			if (oldValue) {
+				// close old session
+				sessionApi.closeSession(oldValue, this.token)
```
juliushaertl, line 132: Would it make sense to wait for the async function before initialising the new session here to avoid race conditions?

src/components/board/Board.vue
```diff
@@ -124,13 +126,51 @@ export default {
 		},
 	},
 	watch: {
-		id: 'fetchData',
+		id(newValue, oldValue) {
+			if (oldValue) {
```
juliushaertl, line 130: This may be skipped if notify push is not enabled

src/components/board/Board.vue
```diff
@@ -124,13 +126,51 @@ export default {
...
+			this.ensureSession()
+		}
+
 		this.fetchData()
+
+		if (isNotifyPushEnabled()) {
+			// regularly let the server know that we are still here
+			this.sessionInterval = setInterval(() => {
+				this.ensureSession()
+			}, 25 * 1000)
```
juliushaertl, line 156: This is a main concern I have with the session pinging approach, as this will add quite a lot of additional load to the Nextcloud server, especially if we starting pinging the session every 25 seconds. I don't have a good alternative in mind, but maybe we can lower this and also make sure that we don't send requests if the browser tab is not in the foreground.

src/components/board/Board.vue
```diff
@@ -145,6 +185,32 @@ export default {
...
+					const res = await sessionApi.createSession(boardId)
+					this.token = res.token
+				} catch (err) {
+					showError(err)
+				}
+			}
+		},
+
+		async refreshData() {
+			await this.$store.dispatch('refreshBoard', this.id)
```
juliushaertl, line 211: Might be a bit more lightweight to just fetch the sessions in the same request as it is renewed

juliushaertl:
Maybe to also kick of some discussion for the next steps, I'm not sure if you have thought about the live update process already, but there is probably a need for an improved approach from the refreshBoard that is currently happening. We already track and expose last modified dates for different elements so that might be an option so that the notify_push queue could advertise the last change date and the vuex store could refresh the content of just the affected items then.

alangecker:
@juliushaertl
thanks for all the feedback 🙂 
I resolved now already some of the smaller points, will continue on it but maybe won't find much time in the next 2 weeks

juliushaertl:
Thanks for the update, looking forward to that :)

lib/Db/SessionMapper.php
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	public function find($boardId, $userId, $token): Session {
+		$qb = $this->db->getQueryBuilder();
+		$result = $qb->select('*')
+			->from($this->getTableName())
+			->where($qb->expr()->eq('board_id', $qb->createNamedParameter($boardId)))
+			->andWhere($qb->expr()->eq('user_id', $qb->createNamedParameter($userId)))
+			->andWhere($qb->expr()->eq('token', $qb->createNamedParameter($token)))
+			->andWhere($qb->expr()->gt('last_contact', $qb->createNamedParameter(time() - SessionService::SESSION_VALID_TIME)))
```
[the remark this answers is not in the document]
alangecker: I am not very familiar with database ind exes, would an index across all columns be the most efficient way to optimize this query?

```php
$table->addIndex(['board_id', 'user_id', 'token', 'last_contact'], 'deck_sessions_find_index');
```

or rather individual indexes for these columns?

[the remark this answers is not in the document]
juliushaertl: If the token is unique we might be able to just query by token and just check the time, board and user id afterwards in php. Then it only needs an index on the token and smaller indexes require less space on disk/ram of the db then

src/components/board/Board.vue
alangecker: in 8f698e8c015a7c98898a514c7f0fd31203079d7d i've changed now the way sessions are created and closed, which should avoid any potential race condition :)

src/components/board/Board.vue
alangecker: with https://github.com/nextcloud/deck/commit/8f698e8c015a7c98898a514c7f0fd31203079d7d the session create&close methods are still being called, but basically noop's when notify push is not enabled

src/components/board/Board.vue
alangecker: since 8f698e8c015a7c98898a514c7f0fd31203079d7d the session closing is much more reliable, so the timeout and pinging is less important and can happen much less often.
with 434b297755a5fe2498b95ade65908a0c31d285c8 it is now set to 90 seconds

lib/Db/SessionMapper.php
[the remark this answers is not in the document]
alangecker: with 6332f8bd this should now be applied as you have proposed :)

thanks for the explanations!

src/components/board/Board.vue
alangecker: the periodic refreshing was there, to catch cases where some update events got lost like with suspended tabs.

with the recent updates we should catch almost everything, so I've removed it, so this should be obsolete :)

matbgn:
@juliushaertl any plan for the review/merge of this highly awaited MR?

alangecker:
I've added now some integration tests and would say that it is now ready to be reviewed from my side 🙂

juliushaertl:
Thanks, I'll have another look next week.

matbgn:
@juliushaertl kindly reminder 😸 ❤️

juliushaertl:
Sorry for the delay, has been busy times 🙈  I'll schedule it next week again

lib/Service/SessionService.php
```diff
@@ -0,0 +1,101 @@
...
+	/**
+	 * @throws DoesNotExistException
+	 */
+	public function syncSession(int $boardId, string $token) {
+		$session = $this->sessionMapper->find($boardId, $this->userId, $token);
+		$session->setLastContact($this->timeFactory->getTime());
+		$this->sessionMapper->update($session);
+	}
+
+	public function closeSession(int $boardId, string $token): void {
```
juliushaertl, line 81: It would be good to have some cleanup (maybe as a background job) to get rid of old session entries.

lib/Service/SessionService.php
```diff
@@ -0,0 +1,101 @@
...
+			$this->sessionMapper->delete($session);
+		} catch (DoesNotExistException $e) {
+		}
+		$this->eventDispatcher->dispatchTyped(new SessionClosedEvent($boardId, $this->userId));
+	}
+
+	public function notifyAllSessions(IQueue $queue, int $boardId, $event, $excludeUserId, $body) {
+		$activeSessions = $this->sessionMapper->findAllActive($boardId);
+
+		foreach ($activeSessions as $session) {
```
juliushaertl, line 93: What would be nice is if we can somehow pass the current session id over on each request from the frontend so that we do not push notifications for the sessions own changes.

Maybe we could wrap axios in some way so that the session (if present) is added as a header to all requests through https://axios-http.com/docs/config_defaults

lib/Service/SessionService.php
```diff
@@ -0,0 +1,101 @@
...
+			$queue->push("notify_custom", [
```
juliushaertl, line 94:
```suggestion
			$queue->push('notify_custom', [
```

lib/Service/SessionService.php
```diff
@@ -0,0 +1,101 @@
...
+	public function closeSession(int $boardId, string $token): void {
+		try {
+			$session = $this->sessionMapper->find($boardId, $this->userId, $token);
+			$this->sessionMapper->delete($session);
+		} catch (DoesNotExistException $e) {
+		}
+		$this->eventDispatcher->dispatchTyped(new SessionClosedEvent($boardId, $this->userId));
+	}
+
+	public function notifyAllSessions(IQueue $queue, int $boardId, $event, $excludeUserId, $body) {
```
juliushaertl, line 90: Just because I was curious how things behave, I added a small patch that would also notify on content changes (which is the next step but seems to work quite nicely already). In case you need that as it is kind of the next step:

```diff
--- a/lib/Db/ChangeHelper.php
+++ b/lib/Db/ChangeHelper.php
@@ -23,10 +23,14 @@

 namespace OCA\Deck\Db;

+use OCA\Deck\NotifyPushEvents;
+use OCA\Deck\Service\SessionService;
+use OCA\NotifyPush\Queue\IQueue;
 use OCP\ICacheFactory;
 use OCP\ICache;
 use OCP\IDBConnection;
 use OCP\IRequest;
+use OCP\Server;

 class ChangeHelper {
        public const TYPE_BOARD = 'boardChanged';
@@ -49,12 +53,22 @@ class ChangeHelper {
                $this->userId = $userId;
        }

-       public function boardChanged($boardId) {
+       public function boardChanged(int $boardId) {
                $time = time();
                $etag = md5($time . microtime());
                $this->cache->set(self::TYPE_BOARD . '-' . $boardId, $etag);
                $sql = 'UPDATE `*PREFIX*deck_boards` SET `last_modified` = ? WHERE `id` = ?';
                $this->db->executeUpdate($sql, [$time, $boardId]);
+               try {
+                       $queue = \OCP\Server::get(IQueue::class);
+                       \OCP\Server::get(SessionService::class)
+                               ->notifyAllSessions($queue, $boardId, NotifyPushEvents::DeckBoardUpdate, $this->userId, [
+                                       'id' => $boardId
+                               ]);
+               } catch (\Exception $e) {
+                       // most likely notify_push is not installed.
+                       return;
+               }
        }

        public function cardChanged($cardId, $updateCard = true) {
```

src/sessions.js
```diff
@@ -0,0 +1,128 @@
...
+ *
+ * @param triggeredBoardId
+ */
+export function triggerDeckReload(triggeredBoardId) {
+	const currentBoardId = store.state.currentBoard?.id
+
+	// only handle update events for the currently open board
+	if (triggeredBoardId !== currentBoardId) return
+
+	store.dispatch('refreshBoard', currentBoardId)
```
juliushaertl, line 53: This is currently not enough to also update the stack/card content as the stacks are currently fetched in Board.vue.

But I think we could just combine the two calls from https://github.com/nextcloud/deck/blob/1168255f171f99c4b3873b3d1d51ec2c10d59220/src/components/board/Board.vue#L152 for that

psalm.xml
```diff
@@ -42,6 +42,7 @@
 				<referencedClass name="OC\*" />
 				<referencedClass name="OC" />
 				<referencedClass name="OC\Security\CSP\ContentSecurityPolicyNonceManager" />
+				<referencedClass name="OCA\NotifyPush\Queue\IQueue" />
```
juliushaertl, line 45: You can add a copy of the Queue interface to tests/stub.phpstub so that psalm can also analyse it.

appinfo/routes.php
```diff
@@ -81,6 +81,11 @@
 		['name' => 'label#update', 'url' => '/labels/{labelId}', 'verb' => 'PUT'],
 		['name' => 'label#delete', 'url' => '/labels/{labelId}', 'verb' => 'DELETE'],
 
+		// sessions
+		['name' => 'Session#create', 'url' => '/session/create', 'verb' => 'PUT'],
```
juliushaertl, line 85: Would be good if that could be an OCS Api as that makes it possible to be used by external clients and the web UI at the same time.

Steps to do that in summary:
- Move the routes to the ocs section in routes.php
- Move the Controller to extend from OCSController
- Use [generateOcsUrl](https://nextcloud.github.io/nextcloud-router/modules/_index_.html#generateocsurl) in the frontend to build the url for the request
- The response data is wrapped now under `ocs.data` in the JSON response

lib/Controller/SessionController.php
```diff
@@ -0,0 +1,92 @@
...
+			'token' => $session->getToken(),
+		]);
+	}
+
+	/**
+	 * notifies the server that the session is still active
+	 * @NoAdminRequired
+	 * @param $boardId
+	 */
+	public function sync($boardId, $token): DataResponse {
```
juliushaertl, line 70:
```suggestion
	public function sync(int $boardId, string $token): DataResponse {
```

lib/Controller/SessionController.php
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	/**
+	 * delete a session if existing
+	 * @NoAdminRequired
+	 * @NoCSRFRequired
+	 * @param $boardId
+	 * @return bool
+	 */
+	public function close($boardId, $token) {
```
juliushaertl, line 87:
```suggestion
	public function close(int $boardId, string $token) {
```

lib/Migration/Version10900Date202206151724222.php
```diff
@@ -0,0 +1,66 @@
...
+				'length' => 64,
+			]);
+			$table->addColumn('board_id', Types::INTEGER, [
+				'notnull' => false,
+			]);
+			$table->addColumn('token', Types::STRING, [
+				'notnull' => true,
+				'length' => 64,
+			]);
+			$table->addColumn('last_contact', Types::INTEGER, [
```
juliushaertl, line 55: Might need an index for queries getting outdated tokens during the cleanup

package.json
```diff
@@ -32,13 +32,14 @@
     "@babel/runtime": "^7.19.0",
     "@juliushaertl/vue-richtext": "^1.0.1",
     "@nextcloud/auth": "^2.0.0",
-    "@nextcloud/axios": "^2.0.0",
+    "@nextcloud/axios": "^1.11.0",
```
juliushaertl, line 35: Seems this needs a rebase to not downgrade the version ;)

juliushaertl (commented):
Very nice progress @alangecker I left some more comments but think it would be good to get this in after that round.

src/sessions.js
alangecker: I had actually rather imagined extending refreshBoard and only triggering an additional reload of the stacks if necessary based on the ETag (see code below)

but for this PR I do not intend to implement actual updates for the stacks/cards, thought to limit it to the active session avatar list as the base and work on the live updates separately :)

```diff
diff --git a/src/store/main.js b/src/store/main.js
index 52c26ab7..ebe19dd9 100644
--- a/src/store/main.js
+++ b/src/store/main.js
@@ -334,10 +334,15 @@ export default new Vuex.Store({
 			commit('setAssignableUsers', board.users)
 		},
 
-		async refreshBoard({ commit }, boardId) {
+		async refreshBoard({ commit, dispatch }, boardId) {
 			const board = await apiClient.loadById(boardId)
+			const etagHasChanged = board.ETag !== this.state.currentBoard.ETag
 			commit('setCurrentBoard', board)
 			commit('setAssignableUsers', board.users)
+			if (etagHasChanged) {
+				dispatch('loadStacks', boardId)
+			}
 		},
 
 		toggleShowArchived({ commit }) {

```

lib/Service/SessionService.php
alangecker: probably good to be aware about what comes next :)

So far I've created a listener ([Application.php](https://github.com/nextcloud/deck/pull/3876/files#diff-4eb824426a8cffb8b33b62b57744c69f32ea30ea755249a7a6170d9402885deb) -> [LiveUpdateListener.php](https://github.com/nextcloud/deck/pull/3876/files#diff-9037a36df135b63c0274adf1d2b46b28b51fd1f16067c8bc1acdbef4b1a23df6)) which is listening for OCP events and forwarding them then to the related sessions.

This seems cleaner for me , because I don't need to hook into all the individual methods for changing things and it's a more standardized way than the `ChangeHelper` methods.

@juliushaertl or would you prefer the `ChangeHelper` way?

package.json
alangecker: since the [dependency issue with the notify_push module](https://github.com/icewind1991/notify_push-client/pull/5) is now fixed, I've reverted the [commit with the workaround](https://github.com/nextcloud/deck/commit/5aa36aa5ecb84cb576a3c226c0368a73a1ccfa3a) and therefore not downgrading the `@nextcloud/axios` anymore ;)

matbgn:
What an impressive job @alangecker ! 😍 

I would be glad to buy you a coffee (yes maybe a big one)... Do you have any sponsorship link? Or any PayPal account?

alangecker:
@matbgn thanks a lot ❤️ 

you've (hopefully) got a mail 🙂 

(If used the only address I could easily find now: mat.gh (at) bqn.**)

matbgn:
> @matbgn thanks a lot ❤️
> 
> you've (hopefully) got a mail 🙂
> 
> (If used the only address I could easily find now: mat.gh (at) bqn.**)

You've done it right! The order is gone, as I said it's not a big amount but I hope you will enjoy your next coffee and again: THANK YOU!

alangecker:
@matbgn thank you very much!

as I already said in the mail, I'm always really happy that people donate for free software, even without asking - like you ❤️

matbgn:
> but for this PR I do not intend to implement actual updates for the stacks/cards, thought to limit it to the active session avatar list as the base and work on the live updates separately :)

And what I can promise you is that when you will release the live updates part I will just double it! 😉 

Happy coding!

appinfo/routes.php
alangecker: with your instructions it was now straight forward to do this :)

-> https://github.com/nextcloud/deck/pull/3876/commits/9b21cef40d8a67d357d3a773fff04554ddd4dd21

lib/Service/SessionService.php
alangecker: I've added this now as proposed: https://github.com/nextcloud/deck/pull/3876/commits/ae6774f252f38c461cf7cac7a0bd1d0fe0b268d1

It feels a bit hacky, but couldn't think about a more beautiful way. maybe someone has suggestions? 🙂

alangecker:
so, from my side I don't see anything to do anymore.
Would be great if someone (@juliushaertl ?) could do another review (especially the recent changes with passing on the session token)

I have also started working on the actual live updates, which seems already fully working 🙂 
https://github.com/alangecker/nextcloud-deck/compare/sessions...live-updates
...PR incoming!

shoetten:
Sorry for the spam, but.. hell yeah! Really excited for this feature! 🎉

matbgn:
I can only agree, really excited to see @juliushaertl's feedback

lib/Listeners/LiveUpdateListener.php
```diff
@@ -0,0 +1,85 @@
...
+		IRequest $request,
+		SessionService $sessionService
+	) {
+		try {
+			$this->queue = $container->get(IQueue::class);
+		} catch (\Exception $e) {
+			// most likely notify_push is not installed.
+			return;
+		}
+		$this->logger = $container->get(LoggerInterface::class);
```
marcelklehr, line 57: nitpick: this could be injected via Dependency injection, no?

marcelklehr:
From looking through the code, it looks really good, apart from the two nitpicks above! I'll also have to setup a test env to properly test it, though.

src/sessions.js
```diff
@@ -0,0 +1,137 @@
...
+
+			// we must assume that the websocket connection was
+			// paused and we have missed updates in the meantime.
+			store.dispatch('refreshBoard', store.state.currentBoard?.id)
+
+			// restart session refresh interval
+			interval = setInterval(ensureSession, SESSION_INTERVAL * 1000)
+		}
+	}
+	document.addEventListener('visibilitychange', visibilitychangeListener)
```
marcelklehr, line 128: closing the session when the tab is not active might be a bit confusing, even if it's technically a cool feature. Google Docs also doesn't do this afaik. @juliushaertl / @jancborchardt Do we want this?

marcelklehr:
I was able to successfully test this. Functionality works when notify_push is enabled and nothing breaks when it's not installed. Great stuff!

nimishavijay:
Design-wise this is great first step towards achieving live updating boards! It looks great :) Really nice work! 🚀

marcelklehr:
@nimishavijay 

> closing the session when the tab is not active might be a bit confusing, even if it's technically a cool feature. Google Docs also doesn't do this afaik. @juliushaertl / @jancborchardt Do we want this?

nimishavijay:
Ah, missed that, sorry!

No I would say closing the session immediately after the tab is unfocused would be confusing. If there was a timeout of say, 15 mins or something that would make more sense :)

shoetten:
@juliushaertl @marcelklehr @nimishavijay Can we merge this sooner than later and work out any feature wishes you might have from a company perspective later? As far as i understand there are no technical blockers anymore and this PR is now almost half a year in review.

juliushaertl (approved):

juliushaertl:
@alangecker Could you maybe rebase to latest master to resolve the conflict? Would be fine to get this in then from my side.

I'd generally agree that the session close after the tab is in the background should rather be deferred for some time. This may also help to avoid additional load for creating and removing the extra sessions.

juliushaertl:
Pushed a commit to make the static analysis happy. Good to go from my side once CI is green 👍

alangecker:
I've just created an issue for further discussions about the session closing/creation topic https://github.com/nextcloud/deck/issues/4354
maybe someone is motivated to put some thinking effort into it? 🙃 

@juliushaertl nice, big thanks for the commit, reviewing and everything else :)

juliushaertl:
Thanks a lot for your awesome contribution. So great to see the first part being merged. Keep up the great work. ❤️

PR #3884, opened 2022-06-28
Fix z-index for deck sidebar

Raudius:
The sidebar would previously render above the user menu (logout, settings, etc)

### Checklist


eneiluj (approved):
:+1: Works fine.

I wonder why it was set to such a high value. Maybe this broke because the user menu z-index was changed in the server's css.

Issue #2594, opened 2020-11-20
Card opens detail view when dragging to other stack


reox:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Describe the bug**

When dragging a card to another stack, it opens up the detail view.
This problem seems to be exclusive to Firefox. It works flawlessly in Edge and Chrome.
Fails on both Windows 10 and CentOS with Firefox 82.0.3 and 68.12esr respectively.

edit: It looks like in *some* cases it is possible to drag the card without opening the details, but I can not figure out when this happens (I tried both touchpad and mouse and the behaviour is the same in both cases)

**To Reproduce**
1. Drag a card to another stack
2. detail view (modal or sidebar) opens up

**Expected behavior**

The detail view should only open on a click not a drag

**Client details:**
 - OS: Windows 10, Centos 7.8
 - Browser Firefox 
 - Version: 82.0.3, 68.12esr
 - Device: Desktop

NC 19.0.5 Deck 1.2.1

Console
```
move card to stack 7 3 Stack.vue:199
jQuery is deprecated: The global jQuery is deprecated. It will be updated to v2.4 in Nextcloud 20 and v3.x in Nextcloud 21. In later versions of Nextcloud it might be removed completely. Please ship your own. 8 globals.js:61:15
Could not find capabilities initial state fall back to _oc_capabilities 3 index.js:14:12
```

leadermontanus:
Can reproduce on Firefox for Windows.

Does not occur on Firefox for Linux (Manjaro)

DragonDev1906:
Can reproduce on Firefox, Linux (Arch), too, but the behavior depends on where the dragging is started (Mouse down):

- If on the title: Card details open
- If on the tag: Filter for this tag is enabled
- If on the Time: Card details open
- Somewhere else on the card: Expected behavior

Same behavior in compact mode, but the tag behaves like expected (probably because there is no click event on the tag in compact mode)

timkrief:
It seems to also happen when dragging stacks by grabbing them from the title of the stack. When dropped, the stack title will enter edition mode. I'm also using firefox.

PR #3916, opened 2022-07-12
Prevent opening card and applyLabelFilter on card drag end

eneiluj:
refs #2594

Problem: A click event is triggered on the element we drag (when dragging a card) when it's dropped.

This issue was mentioned in vue-smooth-dnd: https://github.com/kutlugsahin/vue-smooth-dnd/issues/50 but no solution/fix was provided.

I can't find another way to solve this.

Suggested solution: Thanks to drag-start and drag-end events, we know if we are currently dragging a card or not. We pass this information to the cards. The `applyLabelFilter` and `openCard` methods of the CardItem component don't do anything when the card is being dragged.

### DIFF
--- a/src/components/board/Stack.vue
+++ b/src/components/board/Stack.vue
@@ -99,12 +99,14 @@
 			non-drag-area-selector=".dragDisabled"
 			:drag-handle-selector="dragHandleSelector"
 			@should-accept-drop="canEdit"
+			@drag-start="draggingCard = true"
+			@drag-end="draggingCard = false"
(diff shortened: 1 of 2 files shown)


juliushaertl (approved):
Tested and works 👍

juliushaertl:
PHP tests skipped as no change affected PHP -> Merging

juliushaertl:
@eneiluj Just a quick hint, using `resolves #2594` or `closes #2594` will make sure that the linked issues get closed automatically. `refs` doesn't 😉

PR #3991, opened 2022-08-20
Update to @nextcloud/vue@6.0.0-beta.4

juliushaertl:
- Bump to @nextcloud/vue@6.0.0-beta.3 with renaming all used components to the Nc prefix
- Update vue to 2.7.9
- Small styling fix for the controls bar


src/CardCreateDialog.vue
```diff
@@ -84,17 +84,17 @@
...
-	</Modal>
+	</NcModal>
 </template>
 
 <script>
 import { generateUrl } from '@nextcloud/router'
-import Modal from '@nextcloud/vue/dist/Components/Modal'
-import Multiselect from '@nextcloud/vue/dist/Components/Multiselect'
-import EmptyContent from '@nextcloud/vue/dist/Components/EmptyContent'
+import NcModal from '@nextcloud/vue/dist/Components/Modal'
```
marcelklehr, line 95: The import file probably needs changing too, right?

juliushaertl: Good catch, missed that in my regex replacing

src/components/cards/AvatarList.vue
```diff
@@ -70,14 +70,14 @@
 </template>
 
 <script>
-import { Avatar, PopoverMenu, Tooltip } from '@nextcloud/vue'
+import { NcAvatar, NcPopoverMenu, Tooltip } from '@nextcloud/vue'
```
marcelklehr, line 73: Does Tooltip stay without prefix?

marcelklehr:
Looks good! I found only two visual issues

Peek 2022-08-26 11-05

In the board overview:
image

src/components/cards/AvatarList.vue
juliushaertl: Yes, only components require the two word naming enforced by eslint.

juliushaertl:
Can't reproduce the second, the first should be fixed upstream:
Screenshot 2022-08-26 at 12 56 44

marcelklehr:
> Can't reproduce the second

Maybe because I'm on firefox?

juliushaertl:
Hm, same here. But I'll file issues for follow up fixes then to further investigate.

marcelklehr (approved):

marcelklehr:
Cypress is also still failing

PR #4031, opened 2022-09-06
Implement card reference widget

### LINKED ISSUE #4018: Add reference widget for deck cards

eneiluj:
closes #4018 

This adds a reference widget for deck cards.
It also fixes the DeckIcon.vue component so it uses the current color by default (which fixes an issue with dark themes).

Such URLs will be handled by `@nextcloud/vue-richtext`: `https://nextcloud.local/index.php/apps/deck/#/board/2/card/11`

It is necessary to use Talk master to try this. :warning: Warning: to compile Talk master, make sure you are using Node v14.x and NPM v7.x

Then send a card link in Talk, the displayed message should be like this

https://user-images.githubusercontent.com/11291457/188685151-73ce184e-9293-4178-baf0-0a4a79ca5629.mp4

Bonus question, where does the music come from?


eneiluj:
I also fixed almost all eslint warnings and brought back eslint on compilation. Hope it's fine to do it in this branch. Necessary sooner or later IMO.

eneiluj:
Nevermind the static analysis failure. I did just like https://github.com/juliushaertl/integration_youtube/blob/5867a872bc9d54fc85d07fcee412f2c7fb5ac547/lib/Reference/YoutubeReferenceProvider.php#L48
so this will most likely be fixed on in server.

karlitschek:
this looks amazing. thank you

AndyScherzinger (approved):

src/views/CardReferenceWidget.vue
```diff
@@ -0,0 +1,235 @@
...
+				target="_blank"
+				class="link">
+				{{ board.title }}
+			</a>
+			<a v-tooltip.top="{ content: stackTooltip }"
+				:href="boardLink"
+				target="_blank"
+				class="link">
+				/ {{ stack.title }}
+			</a>
```
nickvergessen, line 37: It feels a bit weird at the moment I think.

Attached is a screenshot of:
1. Link to a card
2. Link to a board
3. A card shared into a conversation

Bildschirmfoto vom 2022-09-12 16-33-23

I think case 1 and 3 should look similar, but the information (Title vs. Subline) is exactly the opposite?!

nickvergessen: Okay nevermind. you are showing much more details when a redis is configured. So for now I guess this is fine and we try to extend the old bit (card sharing) in the future.

eneiluj: 2 is not supported so it tries with the "link reference provider" which returns nothing. That's expected.

Ok I can make it look more like the card shared in the conversation.

jancborchardt (dismissed):
Very cool! :) Design details:
- [x] If the card is left-aligned with the rest of the text it would look a bit nicer (we don’t center anything in there atm)
- [x] I’d also left-align the board and column name

eneiluj:
@jancborchardt @nickvergessen 

image

PR #4036, opened 2022-09-07
Fix attachment creator name: show display name

### LINKED ISSUE #4035: AttachmentList.vue uses uid instead of display name
### How to use GitHub

**Describe the bug**
The AttachmentList.vue component uses the UID instead of the Display Name

Especially for configurations that use LDAP UUIDs this is weird.

https://github.com/nextcloud/deck/blob/f8a255a9f066bb5aa9c7c93cc8056bc5f52516fa/src/components/card/AttachmentList.vue#L66

**To Reproduce**

1. Share a deck card with an attachment
2. Look at the attachments list

**Expected behavior**
The display name is shown

**Screenshots**
image

**Client details:**
 - Browser : Chrome

On c.nc.com

eneiluj:
closes #4035 

The user id was displayed.

Not sure if it's the perfect way to implement that. I did't want to make changes to the Attachment class so the display name goes in the `extendedData` which was originally used for the file info only. Now it also holds the info about the attachment creator.

So this PR adds a `attachmentCreator` field in the `extendedData` (named like that so it can't be confused with the file creator).

We might want to backport this to stable24 to have it at least in the latest NC enterprise.

### DIFF
--- a/lib/Service/AttachmentService.php
+++ b/lib/Service/AttachmentService.php
@@ -39,6 +39,7 @@ use OCA\Deck\StatusException;
 use OCP\AppFramework\Db\IMapperException;
 use OCP\AppFramework\Http\Response;
 use OCP\IL10N;
+use OCP\IUserManager;
 
 class AttachmentService {
 	private $attachmentMapper;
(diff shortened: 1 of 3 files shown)


Raudius (approved):
:+1:

PR #4059, opened 2022-09-13
feat: add validators to check values in services

luka-nextcloud:

* Resolves: # 
* Target version: master 

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/lib/Service/LabelService.php
+++ b/lib/Service/LabelService.php
@@ -29,6 +29,7 @@ use OCA\Deck\Db\Acl;
 use OCA\Deck\Db\LabelMapper;
 use OCA\Deck\StatusException;
 use OCA\Deck\BadRequestException;
+use OCA\Deck\Validators\LabelServiceValidator;
 
 class LabelService {
 
@@ -40,12 +41,21 @@ class LabelService {
 	private $boardService;
 	/** @var ChangeHelper */
 	private $changeHelper;
+	/** @var LabelServiceValidator */
+	private LabelServiceValidator $labelServiceValidator;
 
-	public function __construct(LabelMapper $labelMapper, PermissionService $permissionService, BoardService $boardService, ChangeHelper $changeHelper) {
+	public function __construct(
+		LabelMapper $labelMapper,
+		PermissionService $permissionService,
+		BoardService $boardService,
+		ChangeHelper $changeHelper,
+		LabelServiceValidator $labelServiceValidator
+		) {
 		$this->labelMapper = $labelMapper;
 		$this->permissionService = $permissionService;
 		$this->boardService = $boardService;
 		$this->changeHelper = $changeHelper;
+		$this->labelServiceValidator = $labelServiceValidator;
 	}
 
 	/**
@@ -76,17 +86,7 @@ class LabelService {
 	 * @throws BadRequestException
 	 */
 	public function create($title, $color, $boardId) {
-		if ($title === false || $title === null) {
-			throw new BadRequestException('title must be provided');
-		}
-
-		if ($color === false || $color === null) {
-			throw new BadRequestException('color must be provided');
-		}
-
-		if (is_numeric($boardId) === false) {
-			throw new BadRequestException('board id must be a number');
-		}
+		$this->labelServiceValidator->check(compact('title', 'color', 'boardId'));
 
 		$this->permissionService->checkPermission(null, $boardId, Acl::PERMISSION_MANAGE);
 
@@ -119,9 +119,7 @@ class LabelService {
 	 * @throws BadRequestException
 	 */
 	public function delete($id) {
-		if (is_numeric($id) === false) {
-			throw new BadRequestException('label id must be a number');
-		}
+		$this->labelServiceValidator->check(compact('id'));
 
 		$this->permissionService->checkPermission($this->labelMapper, $id, Acl::PERMISSION_MANAGE);
 		if ($this->boardService->isArchived($this->labelMapper, $id)) {
@@ -144,17 +142,7 @@ class LabelService {
 	 * @throws BadRequestException
 	 */
 	public function update($id, $title, $color) {
-		if (is_numeric($id) === false) {
-			throw new BadRequestException('label id must be a number');
-		}
-
-		if ($title === false || $title === null || $title === "") {
-			throw new BadRequestException('title must be provided');
-		}
-
-		if ($color === false || $color === null) {
-			throw new BadRequestException('color must be provided');
-		}
+		$this->labelServiceValidator->check(compact('title', 'color', 'id'));
 
 		$this->permissionService->checkPermission($this->labelMapper, $id, Acl::PERMISSION_MANAGE);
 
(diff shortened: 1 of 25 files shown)


stefan-niedermann:
Any breaking changes in here? I didn't see one at the first look, but just to make sure 😉

lib/Validators/BaseValidator.php
```diff
@@ -0,0 +1,136 @@
...
+	 */
+	public function check(array $data)
+	{
+		return $this->validate($data);
+	}
+
+	// TODO: check length base on the database field length
+	private function checkLength ($value) {
+		if (!is_string($value)) return true;
+		return strlen($value) <= self::MAX_LENGTH;
```
nickvergessen, line 71: might always want to compare `mb_strlen` here, to not restrict languages like russian too much.
```php
var_dump(strlen('ипсум')); // int(10)
var_dump(mb_strlen('ипсум')); // int(5)
```

When the DB is configured with utf8mb4 support it is also only 5 chars on the db length.

juliushaertl: This method seems to be unused anyways and mb_strlen is used in the getSize method, so I think you could just drop that.

lib/Validators/BaseValidator.php
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	/**
+	 * @param array $data
+	 * @return bool
+	 */
+	public function check(array $data)
+	{
+		return $this->validate($data);
+	}
```
juliushaertl, line 66:
```suggestion
	public function check(array $data): bool {
		return $this->validate($data);
	}
```

lib/Validators/AssignmentServiceValidator.php
```diff
@@ -0,0 +1,14 @@
+<?php
+declare(strict_types=1);
+
```
juliushaertl, line 3: Copyright header is missing ;)

lib/Service/StackService.php
```diff
@@ -291,22 +285,12 @@ public function delete($id) {
 	 * @throws BadRequestException
 	 */
 	public function update($id, $title, $boardId, $order, $deletedAt) {
+		$this->stackServiceValidator->check(compact('title', 'boardId', 'order'));
+
 		if (is_numeric($id) === false) {
```
juliushaertl, line 290: Can we also validate the id through the validator?

lib/Service/LabelService.php
```diff
@@ -144,17 +144,12 @@ public function delete($id) {
 	 * @throws BadRequestException
 	 */
 	public function update($id, $title, $color) {
+		$this->labelServiceValidator->check(compact('title', 'color', 'boardId'));
+
 		if (is_numeric($id) === false) {
```
juliushaertl, line 149: Can we also validate the id through the validator here?

lib/Validators/BaseValidator.php
juliushaertl: Actually it seems to throw so maybe rather use void and annotate the exception that is being thrown here.

lib/Validators/BaseValidator.php
```diff
@@ -0,0 +1,136 @@
...
+
+	// TODO: check length base on the database field length
+	private function checkLength ($value) {
+		if (!is_string($value)) return true;
+		return strlen($value) <= self::MAX_LENGTH;
+	}
+
+	private function numeric ($value) {
```
juliushaertl, line 74:
```suggestion
	private function numeric ($value): bool {
```

lib/Validators/BaseValidator.php
```diff
@@ -0,0 +1,136 @@
...
+		return is_numeric($value);
+	}
+
+	private function bool ($value) {
```
juliushaertl, line 78:
```suggestion
	private function bool ($value): bool {
```

lib/Validators/BaseValidator.php
```diff
@@ -0,0 +1,136 @@
...
+		return is_bool($value);
+	}
+
+	private function not_false ($value) {
```
juliushaertl, line 82:
```suggestion
	private function not_false ($value): bool {
```

lib/Validators/BaseValidator.php
```diff
@@ -0,0 +1,136 @@
...
+		return ($value !== false) && ($value !== 'false');
+	}
+
+	private function not_null ($value) {
```
juliushaertl, line 86:
```suggestion
	private function not_null ($value): bool {
```

lib/Validators/BaseValidator.php
```diff
@@ -0,0 +1,136 @@
...
+		return !is_null($value);
+	}
+
+	private function not_empty ($value) {
```
juliushaertl, line 90:
```suggestion
	private function not_empty ($value): bool {
```

lib/Validators/BaseValidator.php
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	/**
+     * Get the size of an attribute.
+     *
+     * @param  mixed  $value
+     * @return mixed
+     */
+    protected function getSize($value)
+    {
```
juliushaertl, line 114:
```suggestion
	/**
     * Get the size of an attribute.
     *
     * @param  mixed  $value
     */
    protected function getSize($value): int {
```

juliushaertl:
@luka-nextcloud Indentation seems off quite a bit in some places. Maybe double check your editor settings to use tab indentation. php-cs-fixer should however be able to auto fix those with `composer run cs:fix`

luka-nextcloud:
@juliushaertl Please check again.

lib/Validators/BoardServiceValidator.php
```diff
@@ -0,0 +1,44 @@
...
+declare(strict_types=1);
+
+namespace OCA\Deck\Validators;
+
+class BoardServiceValidator extends BaseValidator {
+	public function rules() {
+		return [
+			'id' => ['numeric'],
+			'mapper' => ['not_empty', 'not_null', 'not_false'],
+			'title' => ['not_empty', 'not_null', 'not_false', 'max:100', 'min:5'],
```
juliushaertl, line 34:
```suggestion
			'title' => ['not_empty', 'not_null', 'not_false', 'max:100'],
```

This is causing the test failures it seems as it is a new restriction

lib/Validators/BoardServiceValidator.php
```diff
@@ -0,0 +1,44 @@
+<?php
+/**
+ * @copyright Copyright (c) 2016 Julius Härtl <jus@bitgrid.net>
+ *
+ * @author Julius Härtl <jus@bitgrid.net>
+ * @author Maxence Lange <maxence@artificial-owl.com>
```
juliushaertl, line 6: Make sure to add you to the copyright, as you created that file :)

juliushaertl (approved):

juliushaertl:
- stable23 https://github.com/nextcloud/deck/pull/4175
- stable24 https://github.com/nextcloud/deck/pull/4176

PR #4069, opened 2022-09-22
Reference widget adjustments for Text

eneiluj:
Style context is different between Talk and Text. Here are some adjustments making it better in Text.

Requires https://github.com/nextcloud/text/pull/2977 to try it in Text.

In Talk, all good:
image

In Text, same card, the labels are too wide:
image

Remaining issues:
* [x] Labels overflow
* [x] Description overflow when it contains lists for example

There might be a single cause for those issues. If so, this would make it easier to solve all style issues at once :grin:.


eneiluj:
Oh and I don't know why the spacing between links and widget is wrong for me.

eneiluj:
All good now.

juliushaertl (approved):

PR #4092, opened 2022-10-02
Migrate to nextcloud/OCP package in stable25

nickvergessen:
Signed-off-by: Joas Schilling <coding@schilljs.com>


juliushaertl (approved):

come-nc:
I don’t get the array typing error, the array type is actually correct in the phpdoc block

juliushaertl:
> I don’t get the array typing error, the array type is actually correct in the phpdoc block

Added to the baseline as this seems like a false-positive

PR #4113, opened 2022-10-13
Fix sorting stacks

nickvergessen:
Currently stacks can not be reordered as wished.

The problem is `findAll` always returns sorted by `id`, not by `order`.
So when you move a second column, everything else will reset

* Create a new Deck board with stacks named 1-5
* Move stack 5 to the beginning, expected: 5,1,2,3,4
* Reload the page ✅ 
* Move stack 1 behind 3, expected: 5,2,3,1,4
* Reload the page 💥, actual: 2,3,1,4,5

Happens on c.nc.c as well


skjnldsv (approved):

nickvergessen:
Deployed on company and now it works properly

juliushaertl (approved):

juliushaertl:
Failure seems unrelated, fix is in https://github.com/nextcloud/deck/pull/4114

PR #4115, opened 2022-10-13
Fix issue with duedate format

### LINKED ISSUE #4112: Breaking API change in `1.8.0-beta.2`: card `duedate` has wrong date format
### How to use GitHub

**Describe the bug**
The `duedate` property had the format `2021-03-09T18:35:00+00:00"` in version `1.7.1`, while Deck `1.8.0-beta.2` returns a date with format `2022-10-20 22:00:00`.

**To Reproduce**
1. Use Deck Android to connect to an instance running Deck `1.8.0-beta.2` (like cloud.nextcloud.com). Ensure there is a card with a due date.

**Expected behavior**
Date format did not change compared to previous Deck versions.

<details>
<summary>Server details</summary>

See cloud.nextcloud.com

</details>

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)

Not available because I am only using a guest account.

</details>

Originally reported at https://github.com/stefan-niedermann/nextcloud-deck/issues/1355

nickvergessen:
With this change new items work again.

However the actual problem is that entries in the database have the wrong datetime format already. So at least the timezone is forgotten for every entry.
Also I didn't find the motivation to test all kind of migrations with and without duedate, with past and future, in case there is something related that would need handling.

Fix #4112

### DIFF
--- a/lib/Db/Card.php
+++ b/lib/Db/Card.php
@@ -108,7 +108,7 @@ class Card extends RelationalEntity {
 		$this->addType('archived', 'boolean');
 		$this->addType('notified', 'boolean');
 		$this->addType('deletedAt', 'integer');
-		$this->addType('duedate', 'string');
+		$this->addType('duedate', 'datetime');
 		$this->addRelation('labels');
 		$this->addRelation('assignedUsers');
 		$this->addRelation('attachments');
@@ -126,20 +126,6 @@ class Card extends RelationalEntity {
 		$this->databaseType = $type;
 	}
 
-	public function getDueDateTime(): ?DateTime {
-		return $this->duedate ? new DateTime($this->duedate) : null;
-	}
-
-	public function getDuedate($isoFormat = false): ?string {
-		$dt = $this->getDueDateTime();
-		$format = 'c';
-		if (!$isoFormat && $this->databaseType === 'mysql') {
-			$format = 'Y-m-d H:i:s';
-		}
-
-		return $dt ? $dt->format($format) : null;
-	}
-
 	public function getCalendarObject(): VCalendar {
 		$calendar = new VCalendar();
 		$event = $calendar->createComponent('VTODO');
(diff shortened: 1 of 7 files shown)


stefan-niedermann:
> Also I didn't find the motivation to test all kind of migrations with and without duedate, with past and future, in case there is something related that would need handling.

Could you pleasw clarify the fix for the due dates in the database?

Is it
1. basically there but not tested for each environment?
2. not there at all and everyone who had installed the beta needs to manually fix the database or restore a backup from before the beta 2?

nickvergessen:
So I didn't really test it, but basically the fields where written into the database without the timezone.
But from the one test that I did it basically recovers on read so that the state is acceptable. But the users timezone is gone for good, so for a CEST person all events modified lost the 2h offset and all cards have to be modified again to fix the time.

nickvergessen:
Rebasing for #4114

juliushaertl:
I'm not entirely sure if I remember that correctly but we used to store the due date in etc anyways.

@nickvergessen do you know how the db abstraction handles timezones? According to https://www.sqlite.org/datatype3.html#date_and_time_datatype it does not store those at all.

juliushaertl:
Turns out doctrine never stores the timezone, so having the UTC time in the db is expected and what we also send over from the frontend. 

Ref https://github.com/doctrine/dbal/issues/3916

I tested this with sqlite/mariadb/postgres and could not see a change in the date times. @nickvergessen Was that something you had encountered?

juliushaertl (approved):

juliushaertl:
https://github.com/doctrine/dbal/blob/3.4.x/src/Platforms/AbstractPlatform.php#L4131-L4134

Seems to be used for all our backends https://sourcegraph.com/search?q=context:global+repo:doctrine/dbal+getDateTimeFormatString%28%29+file:Platform&patternType=standard

nickvergessen:
> Was that something you had encountered?

Yes i have some recurring tickets which I do weekly and every week the time changed 2 hours forward. They ended up at 4am until i started also adjusting the time

nickvergessen:
On company server that is, so mariadb

Issue #4162, opened 2022-10-27
wrong alignment of task-list-checkboxes and there labels in card #description-preview


jszeibert:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Describe the bug**
The alignment of task-list-checkboxes and there labels in the card description is off.
(see screenshot below.)

On inspection a `height` of `36px` is set via the nextcloud core style `inputs.scss:203`. Here's an excerpt:
```
input:not([type='range']):not(.input-field__input):not([type='submit']):not([type='button']):not([type='reset']):not(.multiselect__input):not(.select2-input):not(.action-input__input),
select,
div[contenteditable=true],
textarea {
	...
	height: 36px;
	...
```

It seems to be an issue with css rule binding.
Setting `height` to `auto` fixes the issue and indeed there is such a rule in [components/card/Description.vue#L301](https://github.com/nextcloud/deck/blob/3af54d7186a74550debb05792115dc28a687c733/src/components/card/Description.vue#L301)

```
#description-preview {
...
	&::v-deep input {
		min-height: auto;
	}
```

and in inputs.scss:151 as well:

```
/* Specific override */
input {
	...
	&[type='radio'],
	&[type='checkbox'],
	&[type='file'],
	&[type='image'] {
		height: auto;
		width: auto;
	}
```

but they are not applied!

**To Reproduce**
1. create a Card
2. edit the description and add a tasklist
3. switch to view mode

**Expected behavior**
Checkboxes and there labels should be aligned

**Screenshots**
<img width="374" alt="Bildschirmfoto 2022-10-27 um 11 57 58" src="https://user-images.githubusercontent.com/22236337/198254983-cb992b41-2dda-4e9a-99e5-7fe120a6e6d1.png">

**Client details:**
 - OS: Ubuntu 18.04 lts
 - Browser: Firefox 106.0.1 (64 bit)
 - Version: Nextcloud 25.0.0; Deck 1.8.0
 - Device: MacBook Pro

jszeibert:
Just confirmed, ... 
same issue on Safari Version 15.5 (17613.2.7.1.8)

jszeibert:
and here's the commit that broke it:
[...]

PR #4180, opened 2022-11-03
minor style fixes

jszeibert:
* Resolves: #4162
* Target version: master 

### Summary
- changes in nextcloud core style inputs.scss broke the input[type=checkbox] style
=> changed checkbox height from fixed `36px` to `auto`
- somehow the behavior of the core preview provider changed:
	- param `a` that seems to be `crop` (according to [IPreview.php](https://github.com/nextcloud/server/blob/master/lib/public/IPreview.php#L90)) is negated (-> see [core/Controller/PreviewController.php#L144](https://github.com/nextcloud/server/blob/master/core/Controller/PreviewController.php#L144))
	- preview images are upscaled to the next power of 2
=> removed param `a` to receive cropped images

### Checklist

- [x] Documentation (manuals or wiki) has been updated or **is not required**

### DIFF
--- a/src/components/card/AttachmentList.vue
+++ b/src/components/card/AttachmentList.vue
@@ -173,7 +173,7 @@ export default {
 			}
 		},
 		attachmentPreview() {
-			return (attachment) => (attachment.extendedData.fileid ? generateUrl(`/core/preview?fileId=${attachment.extendedData.fileid}&x=64&y=64&a=true`) : null)
+			return (attachment) => (attachment.extendedData.fileid ? generateUrl(`/core/preview?fileId=${attachment.extendedData.fileid}&x=64&y=64`) : null)
 		},
 		attachmentUrl() {
 			return (attachment) => generateUrl(`/apps/deck/cards/${attachment.cardId}/attachment/${attachment.id}`)
--- a/src/components/card/Description.vue
+++ b/src/components/card/Description.vue
@@ -298,7 +298,7 @@ export default {
 	}
 
 	&:deep(input) {
-		min-height: auto;
+		height: auto;
 	}
 
 	&:deep(a) {


src/components/card/AttachmentList.vue
```diff
@@ -173,7 +173,7 @@ export default {
 			}
 		},
 		attachmentPreview() {
-			return (attachment) => (attachment.extendedData.fileid ? generateUrl(`/core/preview?fileId=${attachment.extendedData.fileid}&x=64&y=64&a=true`) : null)
+			return (attachment) => (attachment.extendedData.fileid ? generateUrl(`/core/preview?fileId=${attachment.extendedData.fileid}&x=64&y=64`) : null)
```
juliushaertl, line 176: Can you elaborate more on this change? Having the non-cropped image seems to be expected and I don't see that this was changed on the server side.

jszeibert: I can`t find a change in the server code as well, but I know that images where sized correctly befor I updated my setup to nextcloud 25.

I don't quite understand why the crop flag is negated on fetchPreview in the first placed. One might expect, that calling getPreview with crop=true would result in a cropped image, but it's doesn't

As the preview images in the attachment list are square, I think it's better to fetch them in the correct aspect ratio rather than hiding the overflow.

juliushaertl: I cannot see any issue with this on my test setup, so maybe that is why I don't get it:

Screenshot 2022-11-07 at 20 06 21

Maybe you can split the changes into a separate issue for discussion and possibly also post a system report there so we can rather try to address the root cause.

jszeibert: The issue I wanted to address was the repetition of images as they are set as backgroud-images. I marked two instances I can see in your screenshot.

200393715-7c995338-20ac-4291-b8f7-6ed5530c6cb6

I know that cropping an images might be debatable as parts of an image are cut away. Specialy with very long or very wide images this might give a false impression of the previewed attachment. But repeating the image, does as well. The alternative would be to preserve the aspect ratio and size the preview thumbnails accordingly. But I don't thing that this would look good. So basically this is a style decission. I thought that cropping would be the way to go as other apps like activities or the files app do the same by default.

juliushaertl: Oh, my bad 🙈 Let me check that again then

juliushaertl (approved):

Issue #1472, opened 2020-01-19
Tag creation not working inside the tag selector


wdfee:
**Describe the bug**
Entered tags do not show up / are not saved.

**To Reproduce**
1. Go to a card's details view
2. Enter a new own tag into tags field
3. click enter
4. an empty label is added to tags field, no tag is added to card

**Expected behavior**
An label with the entered tag is created and added to the card.

wdfee:
Uh, I just figured out that tags have to be created in the board settings, not in the input field of a card. So this is no real bug, but maybe it needs some attention to avoid some confusion like this. There should not be an empty tag created at all. 
It would be nice if a tag could be created in the card's input field, too. But if not, there should be some advice for users who try that.

deljones:
I've just flagged the same thing. On the Android app you can create tags, new tags, it even randomly pics a colour for you. The newly created tag is sync'ed to the web version but it appears you cant create a new tag the same way from the web browser. Unless I've missed somethig?

storyjesse:
> Uh, I just figured out that tags have to be created in the board settings

**Where are the board setting??** Can someone please tell me. I can't find them anywhere. :cry:

xdaftcraft:
Hello,

Follow these steps:
Click here
![image](https://user-images.githubusercontent.com/17741063/95458340-51e7ce80-0972-11eb-9a7c-1fccd3264b31.png)
After "Details..."
![image](https://user-images.githubusercontent.com/17741063/95458432-6fb53380-0972-11eb-8a23-37406532c94f.png)
And on the right
![image](https://user-images.githubusercontent.com/17741063/95458529-93787980-0972-11eb-8cee-a709d697ddd8.png)

You can create tag here. When you create a tag it is assigned and usable only to its own table.

Iceman0073:
Does somebody know if the creation of new Tags is now only avail within board settings or is it a bug that I can't create a new Tag on a specific card?

Many thanks for your help.

osm-frasch:
>     * I also went to the Nextcloud Settings, and under the 'Administration' 'Basic Settings' section found 'Collaborative tags'. I have created several 'Collaborative tags' and set them to be 'Public'. I can assign these tags to Document files but they are not available anywhere in Deck.

==> to create 'Collaborative tags' from the  'Administration' 'Basic Settings' section **so that all registered Users can use this tags in the Deck-App too**. This would be a great enhancement for the Deck-App. So Process-Management is more straight forwards.

[...]

PR #4327, opened 2022-12-18
Tag creation from card view

Shadow-Devil:
* Resolves: #1472 
* Target version: master 

### Summary
Implement addLabelToBoardAndCard to also create labels from the CardSidebar. When a label is searched in the label picker for a card and it isn't found, create a new label with a random color and add it to the current card. Tested locally.

### TODO
- [ ] Does this need extra Tests/Documentation?
- [ ] Currently, the placeholder to create the new label is just empty. Maybe we should display a "+" or some text like "Create new label".

### Checklist

### DIFF
--- a/src/components/card/CardSidebarTabDetails.vue
+++ b/src/components/card/CardSidebarTabDetails.vue
@@ -36,7 +36,8 @@
 					label="title"
 					track-by="id"
 					@select="addLabelToCard"
-					@remove="removeLabelFromCard">
+					@remove="removeLabelFromCard"
+					@tag="addLabelToBoardAndCard">
 					<template #option="scope">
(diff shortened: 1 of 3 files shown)


juliushaertl (approved):
Thanks for the pull request. When testing I noticed that the labels were not marked as selected in the multi select after creating. I did a few adjustments locally and will merge those together with your commit as I couldn't push to your branch directly.

juliushaertl:
Will merge with https://github.com/nextcloud/deck/pull/4344 once CI passes over there.

Having a label would be a good follow up for this but I could not find a proper way to set the tag-placeholder with the nextcloud-vue component wrapper NcMultiselect.

Shadow-Devil:
Isn't `tag-placeholder` just working? I thought, NcMultiSelect is just a wrapper for vue-multiselect, where you can set this like described here: https://vue-multiselect.js.org/#sub-tagging

PR #4344, opened 2022-12-30
Tag creation from card view

juliushaertl:
Replacement #4327 for merging with a fixup


juliushaertl (commented):
Approved as part of the old PR

PR #4350, opened 2023-01-02
fix(Sidebar): Only close sidebar on v-click-outside for specific targets

juliushaertl:
Fixes a regression from https://github.com/nextcloud/deck/pull/2272#issuecomment-1368710518

### DIFF
--- a/cypress/e2e/boardFeatures.js
+++ b/cypress/e2e/boardFeatures.js
@@ -1,10 +1,12 @@
 import { randUser } from '../utils/index.js'
 const user = randUser()
+const recipient = randUser()
 
 describe('Board', function() {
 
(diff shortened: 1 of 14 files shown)


juliushaertl:
Pushed a bit of cypress coverage for the sharing sidebar and a follow up fix that allows to clean up cypress a bit :)

marcelklehr (approved):
Sorry for this oversight. Tested and works. Although clicking on the navigation panel doesn't close the sidebar anymore.

juliushaertl:
Thanks for the fast review. I actually did that intentional as it would preserve the board sidebar in open state when switching boards, which seemed more useful than always closing it.

PR #4399, opened 2023-01-27
feat(Description): Use text as editor if available

### LINKED ISSUE #1561: Integrate Nextcloud Text as a markdown editor

juliushaertl:
Resolves #1561 

### Summary

- [x] 🚫 Requires https://github.com/nextcloud/text/pull/3615
- [x] 🐛 Scrolling is not properly working in the modal

Screenshot 2023-02-02 at 07 57 31

### Checklist

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -33,6 +33,11 @@ jobs:
       - name: Set up npm7
         run: npm i -g npm@7
 
+      - name: Register text Git reference
+        run: |
+          text_app_ref="$(if [ "${{ matrix.server-versions }}" = "master" ]; then echo -n "main"; else echo -n "${{ matrix.server-versions }}"; fi)"
+          echo "text_app_ref=$text_app_ref" >> $GITHUB_ENV
+
       - name: Checkout server
         uses: actions/checkout@v3
         with:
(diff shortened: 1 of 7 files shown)


marcelklehr (approved):
Code looks good. Didn't test.

nico-arnold:
Nice to see that! Where can I change the editor? I sadly didn't find any point to change the editor to NC Text. From the "merged" tag here, I guess it's already implemented in Nextcloud Hub 7 (28.0.1)?

PR #4422, opened 2023-02-01
Improve reference provider and add reference widgets

julien-nc:
The card reference provider is now discoverable and searchable. It supports all the available search providers (boards+cards and comments).

There is a minor flaw (i mean just a weird aspect) of the reference provider design. In this case the provider used to resolve and render cards is supporting all the search providers so we can have one single provider entry in the link picker. I'm fine with that but that can be confusive.

New reference widgets for boards and comments.

There would be a lot of factorization to be done between the card and the comment widget. Just waiting we decide what should be their respective content before factorizing.

Also worth mentioning: I changed the CommentService a bit to make it possible to get a comment from a cardId and a commentId. This lead to a small refactoring of the service.

image

TODO:
* [ ] clarify which icons are used in the widgets


lib/Search/CardSearchResultEntry.php
```diff
@@ -26,13 +26,22 @@
...
 use OCA\Deck\Db\Card;
 use OCA\Deck\Db\Stack;
 use OCP\Search\SearchResultEntry;
 
 class CardSearchResultEntry extends SearchResultEntry {
 	public function __construct(Board $board, Stack $stack, Card $card, $urlGenerator) {
-		parent::__construct('', $card->getTitle(), $board->getTitle() . ' » ' . $stack->getTitle(), $urlGenerator->linkToRouteAbsolute('deck.page.index') . '#/board/' . $board->getId() . '/card/' . $card->getId(), 'icon-deck');
+		parent::__construct(
+			$urlGenerator->getAbsoluteURL(
+				$urlGenerator->imagePath(Application::APP_ID, 'card.svg')
```
juliushaertl, line 39: Wondering if this icon is a better match:

https://pictogrammers.com/library/mdi/icon/card-bulleted-outline/

julien-nc: :+1: We had the same idea. It's the one I'm using in the reference widget.

juliushaertl:
Seems the composer ocp packages are not automatically bumped for deck anymore since we renamed the default branch to main. Filed https://github.com/nextcloud/office/issues/25 for this but feel free to just update that composer package manually for this pr.

marcelklehr (approved):
Code looks good. Didn't test.

juliushaertl:
Integration test failures seem related to search of comments ;)

julien-nc:
> Integration test failures seem related to search of comments ;)

I changed a search provider ID and didn't adjust the tests... All good now.

julien-nc:
Rebased on main

juliushaertl (commented):
Looks good :+1: Please merge after updating nextcloud/ocp if tests pass :)

julien-nc:
`__construct has more required parameters than parent method` for `OCA\Deck\Model\BoardSummary` and `OCA\Deck\Model\CardDetails`.

It does not sound like an issue to me :grin:. Should we disable/ignore this psalm check?

juliushaertl:
Yep, let's ignore that for this PR.

juliushaertl (approved):

PR #4488, opened 2023-02-27
fix: Avoid mutating the due date when calculating days

### LINKED ISSUE #4483: v26 all due date times are ignored
1. Set a Due date with a time e.g. 10am CET
2. Due time shows well
3. F5
4. Time is shown as 1:00am (CET time of UTC 12:00pm)

juliushaertl:
DateTime is mutable so we need to take a copy for comparing the days between due date and now. 

Fixes #4483

### DIFF
--- a/lib/Db/Card.php
+++ b/lib/Db/Card.php
@@ -159,16 +159,17 @@ class Card extends RelationalEntity {
 	}
 
 	public function getDaysUntilDue(): ?int {
-		$today = new DateTime();
-		$match_date = $this->getDuedate();
-		if ($match_date === null) {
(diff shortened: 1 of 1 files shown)


nickvergessen (approved):

nickvergessen:
Applied as a patch and works nicely again.

I see it correct that no data was corrupted? Only if you manually saved the faulty time?

juliushaertl:
> I see it correct that no data was corrupted? Only if you manually saved the faulty time?

Yes, even saving should have set the correct time, just the API response always returned 00:00 as the time in UTC.

PR #4452, opened 2023-02-15
Performance boost

juliushaertl:
Improvements are split into individual commits for easier review. 

Overall this adds additional caching layers and groups queries where we performed them for n results into one.

## ToDo

- [x] Get in #3793 first / might fix OverviewService to perform queries per board
- [x] StackMapper::find could use some caching on request level
- [x] BoardService::find does not need full details (users/settings/sessions) when we set the relative board on cards 
- [x] Group queries for related stacks
- [x] Make sure that in queries do not exceed the oracle limit: More than 1000 expressions in a list are not allowed on Oracle.

## Follow up issues
- [ ] CardMapper findBoardId cache should probably use something like WithLocalCache to avoid heavy round trips to redis (which happens at least once for every file share)
- [ ] Enrichment of board data could be grouped queries (might only have a noticeable effect on hundreds of boards a user has)

### Board list 

- Removes 1 query per board the user has access to
- Removes 1 additional query

Screenshot 2023-02-17 at 09 50 43

### Overview

- Removes 2-4 queries per card that is listed in the overview (extra queries still needed if a card has comments to fetch the unread counter)

Screenshot 2023-02-17 at 09 51 52

### Single board request

- removes 1 query for each sharing rule that a board has

Screenshot 2023-02-17 at 09 53 45

### Stack list

- removes 2-4 queries per card that is on the board
- reduce query to fetch stack details from the amount of cards to the number of stacks (40 cards, 4 stacks -> 40 to 4 queries)

Screenshot 2023-02-17 at 09 55 15

### Propfind on Deck folder (3200 incoming deck file shares)

Screenshot 2023-02-17 at 09 59 12

## Testing

<details>
<summary>Script to generate random board content</summary>

```
<?php
/*
 * @copyright Copyright (c) 2021 Julius Härtl <jus@bitgrid.net>
 *
 * @author Julius Härtl <jus@bitgrid.net>
 *
 * @license GNU AGPL version 3 or any later version
 *
 * This program is free software: you can redistribute it and/or modify
 * it under the terms of the GNU Affero General Public License as
 * published by the Free Software Foundation, either version 3 of the
 * License, or (at your option) any later version.
 *
 * This program is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
 * GNU Affero General Public License for more details.
 *
 * You should have received a copy of the GNU Affero General Public License
[...]

PR #4493, opened 2023-03-01
fix: Always return sorted index array to make sure a json array is the result

juliushaertl:
This has been lost in https://github.com/nextcloud/deck/pull/4452/files#diff-4975b048acd756b9fe8e6d63b016021d70ade5c1f79a5797ebe4dc6f2e87d34fL182 but the place was not that obvious. We need to use the array_values of the array_unique result to make sure that indexes are still properly in line and the later conversion to JSON results in an array not an object.


juliushaertl:
Will see if we can have a test case for this in a follow up PR.

PR #4514, opened 2023-03-06
fix(references): Mute NoPermissionException as it is expected to happen for references

juliushaertl:
Avoid throwing and logging permission exceptions which may be expected and should be handled by just returning no reference for deck references.

### DIFF
--- a/lib/Reference/CommentReferenceProvider.php
+++ b/lib/Reference/CommentReferenceProvider.php
@@ -27,6 +27,7 @@ use OCA\Deck\Db\Assignment;
 use OCA\Deck\Db\Attachment;
 use OCA\Deck\Db\Label;
 use OCA\Deck\Model\CardDetails;
+use OCA\Deck\NoPermissionException;
 use OCA\Deck\Service\BoardService;
 use OCA\Deck\Service\CardService;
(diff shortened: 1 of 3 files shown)


marcelklehr (approved):

lib/Reference/CommentReferenceProvider.php
```diff
@@ -85,9 +85,14 @@ public function resolveReference(string $referenceText): ?IReference {
...
 				[$boardId, $cardId, $commentId] = $ids;
 
-				$card = $this->cardService->find($cardId)->jsonSerialize();
-				$board = $this->boardService->find($boardId)->jsonSerialize();
-				$stack = $this->stackService->find((int) $card['stackId'])->jsonSerialize();
+				try {
+					$card = $this->cardService->find($cardId)->jsonSerialize();
+					$board = $this->boardService->find($boardId)->jsonSerialize();
+					$stack = $this->stackService->find((int) $card['stackId'])->jsonSerialize();
+				} catch (NoPermissionException $e) {
```
marcelklehr, line 92: NoPermissionException still needs to be imported

juliushaertl: Fixed

Issue #4517, opened 2023-03-08
`api/v1.0/boards?details=true` does not return the labels and stacks anymore, breaking the Android app


nickvergessen:
### 26
```json
$ curl -k -u admin:admin https://nextcloud26.local/index.php/apps/deck/api/v1.0/boards?details=true | jq
[
  {
    "title": "Personal",
    "owner": {
      "primaryKey": "admin",
      "uid": "admin",
      "displayname": "Laura Adams",
      "type": 0
    },
    "color": "0087C5",
    "archived": false,
    "labels": [],
    "acl": [],
    "permissions": {
      "PERMISSION_READ": true,
      "PERMISSION_EDIT": true,
      "PERMISSION_MANAGE": true,
      "PERMISSION_SHARE": true
    },
    "users": [],
    "shared": 0,
    "stacks": [],
    "activeSessions": [],
    "deletedAt": 0,
    "lastModified": 1678217691,
    "settings": [],
    "id": 1,
    "ETag": "76e5a47a517169505cf3fb9f4f8b9af5"
  }
]
```

### 25

```json
$ curl -k -u admin:admin https://nextcloud25.local/index.php/apps/deck/api/v1.0/boards?details=true | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1510  100  1510    0     0   2425      0 --:--:-- --:--:-- --:--:--  2423
[
  {
    "title": "Persönlich",
    "owner": {
      "primaryKey": "admin",
      "uid": "admin",
      "displayname": "Laura Adams",
      "type": 0
    },
    "color": "0087C5",
    "archived": false,
    "labels": [
      {
        "title": "Abgeschlossen",
        "color": "31CC7C",
        "boardId": 1,
        "cardId": null,
        "lastModified": 1678216421,
        "id": 1,
        "ETag": "fbb635f4624dd7dd79b0a7f4dc627fa7"
      },
      {
        "title": "Zu überprüfen",
        "color": "317CCC",
        "boardId": 1,
        "cardId": null,
        "lastModified": 1678216421,
        "id": 2,
        "ETag": "fbb635f4624dd7dd79b0a7f4dc627fa7"
      },
      {
        "title": "Handlung erforderlich",
        "color": "FF7A66",
        "boardId": 1,
        "cardId": null,
        "lastModified": 1678216421,
        "id": 3,
[...]

PR #4518, opened 2023-03-08
fix(API): Fix board API details parameter to work as expected

nickvergessen:
* Resolves: #4517 
* Target version: main

### Summary

Before `?details=true` in the URL resulted in `string "true"` and therefore did not match `$details === true` which was brought in by https://github.com/nextcloud/deck/commit/b19b7794bc773d824275f18d58263c382532790f but it matched the old `$details` logic.

Defining the parameter as `bool` (like it was done in the other touched controller methods) gives the actual boolean to the controller fixing the comparison

### Checklist

### DIFF
--- a/lib/Controller/BoardApiController.php
+++ b/lib/Controller/BoardApiController.php
@@ -60,9 +60,11 @@ class BoardApiController extends ApiController {
 	 * @NoCSRFRequired
 	 *
 	 * Return all of the boards that the current user has access to.
+	 *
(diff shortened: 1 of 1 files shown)


nickvergessen:
Also applied it on our instance as a patch and the Android app works again :)

stefan-niedermann (approved):

juliushaertl (approved):

Issue #4523, opened 2023-03-08
Create card from message in talk doesn't work


pschopen:
When trying to create a card from a message in talk the window to select a board and a list appears but when submitting the screen gets white and nothing happens. However, a card is created.

Instead a message should appear telling that creating the card was successful.

<img width="654" alt="grafik" src="https://user-images.githubusercontent.com/65334274/223806538-d86d75a7-e673-48f4-af2b-ef7f8793f938.png">

pschopen:
still doesn't work with 26.0.0

PR #4561, opened 2023-03-27
fix: Adapt NcEmptyContent usages to new slots

juliushaertl:
Fixes https://github.com/nextcloud/deck/issues/4523

### Summary

NcEmptyContent has new slots that we should use to properly show the content

### Checklist

### DIFF
--- a/src/CardCreateDialog.vue
+++ b/src/CardCreateDialog.vue
@@ -71,12 +71,22 @@
 				</div>
 			</div>
 			<div v-else id="modal-inner">
-				<NcEmptyContent v-if="creating" icon="icon-loading">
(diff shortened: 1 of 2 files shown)


src/CardCreateDialog.vue
```diff
@@ -71,12 +71,22 @@
 				</div>
 			</div>
 			<div v-else id="modal-inner">
-				<NcEmptyContent v-if="creating" icon="icon-loading">
-					{{ t('deck', 'Creating the new card …') }}
+				<NcEmptyContent v-if="creating">
```
marcelklehr, line 74: Not sure if this counts as abuse but it certainly works :D

marcelklehr (approved):

Issue #3464, opened 2021-11-30
Nextcloud 23 not showing Shares - Shared by me


apg1980:
mariadb 10.5.13
php8.0.13
Apache
ubuntu 20.04 lts

![grafik](https://user-images.githubusercontent.com/9514965/144066397-2652dae9-41c0-4311-9ebb-bed860b39d90.png)
....with a tons of shares

juliushaertl:
FYI @mejo- @azul For the collectives exception trace. Though it then looks unrelated to the issue that is disappearing when deck is disabled. Could you see if there is any other error message when loading the Shared with me page?

juliushaertl:
Also please report which version of deck you are using and maybe check the just release 1.6.0 to see if this also shows up the issue.

azul:
Let's format the trace a bit:
```json
{
   "app" : "no app in context",
   "exception" : {
      "Code" : 0,
      "CustomMessage" : "--",
      "Exception" : "Exception",
      "File" : "/var/www/nextcloud/lib/private/AppFramework/Http/Dispatcher.php",
      "Line" : 158,
      "Message" : "OCA\\Collectives\\Db\\CollectiveShareMapper::deleteByToken(): Argument nextcloud/server#1 ($token) must be of type string, null given, called in /var/www/nextcloud/apps/collectives/lib/Listeners/ShareDeletedListener.php on line 30 in file '/var/www/nextcloud/apps/collectives/lib/Db/CollectiveShareMapper.php' line 125",
      "Previous" : {
         "Code" : 0,
         "Exception" : "TypeError",
         "File" : "/var/www/nextcloud/apps/collectives/lib/Db/CollectiveShareMapper.php",
         "Line" : 125,
         "Message" : "OCA\\Collectives\\Db\\CollectiveShareMapper::deleteByToken(): Argument nextcloud/server#1 ($token) must be of type string, null given, called in /var/www/nextcloud/apps/collectives/lib/Listeners/ShareDeletedListener.php on line 30",
         "Trace" : [
            {
               "class" : "OCA\\Collectives\\Db\\CollectiveShareMapper",
               "file" : "/var/www/nextcloud/apps/collectives/lib/Listeners/ShareDeletedListener.php",
               "function" : "deleteByToken",
               "line" : 30,
               "type" : "->"
            },
            {
               "class" : "OCA\\Collectives\\Listeners\\ShareDeletedListener",
[...]

PR #4566, opened 2023-03-28
Gracefully handle not found card for a share

mejo-:
Fixes: #3464
Fixes: #4565

* Resolves: # 
* Target version: main

### DIFF
--- a/lib/Sharing/ShareAPIHelper.php
+++ b/lib/Sharing/ShareAPIHelper.php
@@ -30,8 +30,10 @@ use OCA\Deck\Db\Acl;
 use OCA\Deck\Db\CardMapper;
 use OCA\Deck\NoPermissionException;
 use OCA\Deck\Service\PermissionService;
+use OCP\AppFramework\Db\DoesNotExistException;
 use OCP\AppFramework\OCS\OCSNotFoundException;
 use OCP\AppFramework\Utility\ITimeFactory;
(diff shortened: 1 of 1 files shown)


mejo-:
> shouldn't a share to a non existing card be deleted and not ignored?

We should rather make sure that shares get deleted along with the cards, but that's a separate topic. Also, since cards don't get deleted straight away but only get marked as deleted, we'd have to clean up shares in the cronjob that actually deletes the cards.

mejo-:
@juliushaertl the PR is tested and ready for review now.

juliushaertl:
Still curious how this could happen as the DeleteCron should also drop the share through the attachment implementation https://github.com/nextcloud/deck/blob/1f66c66ad3a5e830c17a7f2fc43f0ec6901d492b/lib/Service/FilesAppService.php#L276 Any idea about that from your investigations @mejo- ?

Nevertheless the change makes sense

juliushaertl (approved):

mejo-:
>  Any idea about that from your investigations @mejo- ?

Yep. Whenever you soft-delete a card, it will no longer be found by `cardMapper->find()`, resulting in this error. So any soft-deleted card that has an attachment shared with it led to this bug in my understanding. Probably the problem got solved by `DeleteCron` when it really deleted the card and its related attachments.

PR #4604, opened 2023-04-20
Better display of card dates (creation and change dates)

Jerome-Herbinet:
Signed-off-by: Jérôme Herbinet <33763786+Jerome-Herbinet@users.noreply.github.com>

### After (inverted + strong and space before ":") : 

Note : 2 lines are inverted in tooltip as well ... so the "created" line is now above.

### Checklist

### DIFF
--- a/src/components/card/CardSidebar.vue
+++ b/src/components/card/CardSidebar.vue
@@ -147,10 +147,10 @@ export default {
 			return this.$store.getters.cardById(this.id)
 		},
 		subtitle() {
-			return '<strong>' + t('deck', 'Created') + ' :</strong> ' + this.relativeDate(this.currentCard.createdAt * 1000) + '<br><strong>' + t('deck', 'Modified') + ' :</strong> ' + this.relativeDate(this.currentCard.lastModified * 1000)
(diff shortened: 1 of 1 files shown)


juliushaertl:
I personally don't like the line break, since we have plenty of space there, but requesting some input from our designers :)

nimishavijay:
Great point! It could definitely have better separation!

We use the middle dot symbol (⸱) as a separator in many instances, that could be more fitting here :) That way we can also remove the colon and bolding so that it reads more naturally, like this:

`Modified 2 days ago ⸱ Created 1 month ago`

What do you think?

Jerome-Herbinet:
> Great point! It could definitely have better separation!
> 
> We use the middle dot symbol (⸱) as a separator in many instances, that could be more fitting here :) That way we can also remove the colon and bolding so that it reads more naturally, like this:
> 
> `Modified 2 days ago ⸱ Created 1 month ago`
> 
> What do you think?

Done @nimishavijay 
Cc @juliushaertl

juliushaertl (approved):

Issue #4618, opened 2023-04-24
Deck calendar always visible in calendar app


schiessle:
When userA shares a Deck board with userB, the calendar for the corresponding deck board is always visible for userB, userB can't hide it.

When userB tries to hide the calendar by clicking on it in the calendar app, they will get the message that a error happened while changing the visibility of the calendar. This is the response in the network console of the browser:

`<?xml version="1.0" encoding="utf-8"?>
<d:error xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">
  <s:exception>Sabre\DAV\Exception\NotFound</s:exception>
  <s:message>Node with name 'app-generated--deck--board-3' could not be found</s:message>
</d:error>`

Only if userA gives userB the "can manage" permission for the Deck board, userB can hide it corresponding calendar.

The option to hide/show calendars should be independent from the permission a user has for the corresponding deck board. They should always be able to configure their personal view.

PR #4622, opened 2023-04-25
allow user to toggle visibility of the calendar for a deck board

schiessle:
* Resolves:  https://github.com/nextcloud/deck/issues/4618

* Target version: main

### Summary

The calendar object needs to be exposed with "write-properties" in order to allow users to hide/show the corresponding calendar in the calendar app. It doesn't has any affects on the general permissions for the whole board or the individual cards. As soon as you select a calendar item you jump to the deck app where the normal deck permissions are applied.

As far as I understand it, caldav knows two kind if write permissions:

* "{DAV:}write" for editing calendar entries
* "{DAV:}write-properties" to only edit properties (like the visibility) of the calendar.

Looking at how we treat shared regular calendars, I think the solution should be fine: https://github.com/nextcloud/server/blob/master/apps/dav/lib/CalDAV/Calendar.php#L181

### Checklist

@juliushaertl This was a quick try to fix a issue I discovered, does it makes sense?

### DIFF
--- a/lib/DAV/Calendar.php
+++ b/lib/DAV/Calendar.php
@@ -188,12 +188,18 @@ class Calendar extends ExternalCalendar {
 			foreach ($properties as $key => $value) {
 				switch ($key) {
 					case '{DAV:}displayname':
+						if (!$this->backend->checkBoardPermission($this->board->getId(), Acl::PERMISSION_MANAGE)) {
+							throw new Forbidden('no permission to change the displayname');
+						}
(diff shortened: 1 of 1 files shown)


schiessle:
@juliushaertl I added the additional checks.

juliushaertl (approved):

juliushaertl:
Tested and works. I quickly fixed the php code style error.

juliushaertl:
DCO check seems confused

PR #4623, opened 2023-04-26
Grouped dependency update

juliushaertl:
- chore: Bump nextcloud-vue-collections
- chore: Bump @nextcloud/vue @nextcloud/webpack-vue-config and remove @nextcloud/vue-richtext
- Move to material icons as icon class got broken with the nextcloud vue library update for sidebar tabs


datenangebot (approved):
no remarks

PR #3793, opened 2022-05-09
Optimise upcomming overview creation

Raudius:
* Target version: master 

### Summary
By pre-sorting the cards by due date in the backend we can speed-up the view creation, particularly for instances where many cards have to be loaded.

### Checklist

- [ ] Code is properly formatted
- [ ] Sign-off message is added to all commits
- [ ] Tests (unit, integration, api and/or acceptance) are included
- [ ] Documentation (manuals or wiki) has been updated or is not required

PR #4660, opened 2023-05-09
fix: Bring back overdue column by removing faulty condition

juliushaertl:

* Resolves: # 
* Target version: main

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/src/components/overview/Overview.vue
+++ b/src/components/overview/Overview.vue
@@ -31,7 +31,7 @@
 		</div>
 
 		<div v-else-if="isValidFilter" class="overview">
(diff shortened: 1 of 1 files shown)


datenangebot (approved):
Code is okay for me, so approved.
But whats the reason for the change in nc26 or was is a general bug?
Do we want to show the column even if there are no cards?

juliushaertl:
This was an oversight when I adapted the frontend bits for the performance optimizations done in https://github.com/nextcloud/deck/pull/3793, therefore only stable26. We probably want some more sophisticated empty content behaviour but that is a larger topic with the design mockups we have at https://github.com/nextcloud/deck/issues/2807, so something for a different time. ;)

PR #4688, opened 2023-05-17
Faster frontend developing with HMR

juliushaertl:
I used to have HMR on deck already but the config was a bit fragile and not really used as far as I was aware. I took some inspiration from talk here so this should ease frontend development and might be a good idea to bring to other office team apps.


datenangebot (approved):
Code looks good to me,
did not test it yet.

PR #4718, opened 2023-05-21
Refactors controllers by using PHP8's constructor property promotion.

fsamapoor:
### Summary
There are lots of refactoring opportunities and this PR aims to see if it's welcome to make such improvements to the code base.

### DIFF
--- a/lib/Controller/CardApiController.php
+++ b/lib/Controller/CardApiController.php
@@ -38,21 +38,22 @@ use OCP\IRequest;
  * @package OCA\Deck\Controller
  */
 class CardApiController extends ApiController {
-	private $cardService;
-	private $userId;
-	private $assignmentService;
 
 	/**
 	 * @param string $appName
 	 * @param IRequest $request
 	 * @param CardService $cardService
+	 * @param AssignmentService $assignmentService
 	 * @param $userId
 	 */
-	public function __construct($appName, IRequest $request, CardService $cardService, AssignmentService $assignmentService, $userId) {
+	public function __construct(
+		string $appName,
+		IRequest $request,
+		private CardService $cardService,
+		private AssignmentService $assignmentService,
+		private $userId,
+	) {
 		parent::__construct($appName, $request);
-		$this->cardService = $cardService;
-		$this->userId = $userId;
-		$this->assignmentService = $assignmentService;
 	}
 
 	/**
--- a/lib/Controller/PageController.php
+++ b/lib/Controller/PageController.php
@@ -44,37 +44,23 @@ use OCP\IRequest;
 use OCP\IURLGenerator;
 
 class PageController extends Controller {
-	private PermissionService $permissionService;
 	private IInitialStateService $initialState;
-	private ConfigService $configService;
-	private IEventDispatcher $eventDispatcher;
-	private CardMapper $cardMapper;
-	private IURLGenerator $urlGenerator;
-	private CardService $cardService;
-	private IConfig $config;
 
 	public function __construct(
 		string $AppName,
 		IRequest $request,
-		PermissionService $permissionService,
+		private PermissionService $permissionService,
 		IInitialStateService $initialStateService,
-		ConfigService $configService,
-		IEventDispatcher $eventDispatcher,
-		CardMapper $cardMapper,
-		IURLGenerator $urlGenerator,
-		CardService $cardService,
-		IConfig $config
+		private ConfigService $configService,
+		private IEventDispatcher $eventDispatcher,
+		private CardMapper $cardMapper,
+		private IURLGenerator $urlGenerator,
+		private CardService $cardService,
+		private IConfig $config,
 	) {
 		parent::__construct($AppName, $request);
 
-		$this->permissionService = $permissionService;
 		$this->initialState = $initialStateService;
-		$this->configService = $configService;
-		$this->eventDispatcher = $eventDispatcher;
-		$this->cardMapper = $cardMapper;
-		$this->urlGenerator = $urlGenerator;
-		$this->cardService = $cardService;
-		$this->config = $config;
 	}
 
 	/**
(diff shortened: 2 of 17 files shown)


lib/Controller/PageController.php
```diff
@@ -44,37 +44,23 @@
...
-	private IURLGenerator $urlGenerator;
-	private CardService $cardService;
-	private IConfig $config;
 
 	public function __construct(
 		string $AppName,
 		IRequest $request,
-		PermissionService $permissionService,
+		private PermissionService $permissionService,
 		IInitialStateService $initialStateService,
```
juliushaertl, line 53: Any reason to not refactor the `initialStateService`  ?

lib/Controller/StackApiController.php
```diff
@@ -39,18 +39,20 @@
...
  */
 class StackApiController extends ApiController {
-	private $boardService;
-	private $stackService;
 
 	/**
 	 * @param string $appName
 	 * @param IRequest $request
 	 * @param StackService $stackService
+	 * @param BoardService $boardService
```
juliushaertl, line 47: i'd remove the phpdoc directly since we have it as type hints now and it doesn't add additional information

lib/Controller/StackApiController.php
```diff
@@ -39,18 +39,20 @@
...
 	 */
-	public function __construct($appName, IRequest $request, StackService $stackService, BoardService $boardService) {
+	public function __construct(
+		$appName,
+		IRequest $request,
+		private StackService $stackService,
+		private BoardService $boardService
+		) {
```
juliushaertl, line 54: I'd personally prefer to not indent the closing round brackets, but since the php-cs check passes I'm also fine with that.

lib/Controller/StackController.php
```diff
@@ -30,12 +30,13 @@
 use OCP\AppFramework\Controller;
 
 class StackController extends Controller {
-	private $userId;
-	private $stackService;
-	public function __construct($appName, IRequest $request, StackService $stackService, $userId) {
+	public function __construct(
+		$appName,
```
juliushaertl, line 34: Maybe we can directly type hint this one as well:

```suggestion
		string $appName,
```

lib/Controller/CardApiController.php
```diff
@@ -38,21 +38,22 @@
...
 	/**
 	 * @param string $appName
 	 * @param IRequest $request
 	 * @param CardService $cardService
+	 * @param AssignmentService $assignmentService
 	 * @param $userId
 	 */
-	public function __construct($appName, IRequest $request, CardService $cardService, AssignmentService $assignmentService, $userId) {
+	public function __construct(
+		$appName,
```
juliushaertl, line 50:
```suggestion
		string $appName,
```

juliushaertl (commented):
Thanks a lot. I've left a few more comments that would make sense from my perspective if we touch the lines anyways :) Otherwise this would be fine to get in.

lib/Controller/StackApiController.php
fsamapoor: For me personally, It would be fine either way. I'm just always looking for consistency. I saw other controllers indented this way, and I just wanted to unite them.

juliushaertl: Then let's keep it as it is for now

lib/Controller/StackApiController.php
fsamapoor: I'm all for removing redundant phpdocs altogether when it does not add any complementary value. I just figured I should limit the scope of changes in this PR. However, I can remove all redundant phpdocs in this PR as well. I just needed to know if it's okay and we're on the same page on this matter.

lib/Controller/PageController.php
fsamapoor: Again, it was just about limiting the scope of changes in this PR. The `initialStateService` gets assigned to the `$initialState` in the constructor. I just wasn't sure if that is a specific convention in the NC ecosystem or not.

So basically, I should change all the occurrences of the `$initialState` with the new `$initialStateService` and that will make things cleaner in my opinion. Would you agree with that?

fsamapoor:
Hello @juliushaertl.

Although not related to changes being made in this PR, I have fixed an error regarding the failed static analysis workflow. After that, I ran the psalm script in my local environment and it finds loads of errors in the project.

In the Controller namespace alone, there are 149 errors found containing only "UndefinedClass" and "UndefinedDocblockClass" errors. I'm new to using Psalm, but to my understanding, those are autoloading-related issues that can be fixed by changing the Psalm configuration file, which is well beyond the scope of this PR.

So, are there any specific steps I need to take in this PR regarding the failed static analysis workflow?

lib/Controller/StackApiController.php
juliushaertl: Let's keep this separate then :)

juliushaertl:
>  In the Controller namespace alone, there are 149 errors found containing only "UndefinedClass" and "UndefinedDocblockClass" errors. I'm new to using Psalm, but to my understanding, those are autoloading-related issues that can be fixed by changing the Psalm configuration file, which is well beyond the scope of this PR.

I'm not sure about those, I don't see them locally. For static analysis we pull in the nextcloud/ocp composer package which is a copy of the public API provided by the Nextcloud server. It might be that this is either missing on your setup, depending how you run psalm. Other than that I also had troubles in the past where  disabling the psalm caching worked: `composer run psalm -- --no-cache`

juliushaertl:
Would be good to merge from my perspective, can you just do a rebase to latest master to get rid of the merge commit?

fsamapoor:
> Would be good to merge from my perspective, can you just do a rebase to latest master to get rid of the merge commit?

Sorry about that. All good now.

juliushaertl:
Failures are likely unrelated as we are just having a test run for more CI workers in the GitHub organization, will trigger them again later

juliushaertl:
Would you mind to rebase and squash the commits into one for a clean history?

juliushaertl (approved):

fsamapoor:
@juliushaertl Done.

juliushaertl:
Awesome. Thanks a lot and sorry for taking so long :)

PR #4772, opened 2023-06-09
fix: Allow dynamic autoloading for classes added during upgrade

juliushaertl:
Follow https://github.com/nextcloud/documentation/pull/10578


datenangebot (approved):
Like everywhere 🙃

PR #4816, opened 2023-06-15
fix: crash when leaving out system parameter

max-nextcloud:
### Summary

The `--system` parameter can be supplied via command line or selected afterwards.

However if none was provided the command would crash with `TypeError: Cannot assign null to property $system`.

Handle that gracefully and make the type spec more precise for the setSystem function.

* Target version: main

### Checklist

- no tests as we're not testing `occ` commands yet.
- no documentation as this is a bug fix.

### DIFF
--- a/lib/Service/Importer/BoardImportService.php
+++ b/lib/Service/Importer/BoardImportService.php
@@ -143,11 +143,13 @@ class BoardImportService {
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


max-nextcloud:
I'm not sure how far this would need to be backported. The type check that causes this to fail might be new. On the other hand I don't think it can do any harm.

PR #4906, opened 2023-07-13
fix: Split query to fetch board ids to avoid slow query join

juliushaertl:
### Summary

Some over optimization for query count caused this slow query, which does seem to do a full table scan on the boards table (left part of the join):

```
SELECT DISTINCT b.id FROM oc_deck_boards b 
	LEFT JOIN oc_deck_board_acl acl ON b.id = acl.board_id 
	WHERE (owner = 'user1')
	OR ((acl.participant = 'user1') AND (acl.type = 0)) 
	OR ((acl.participant IN ('group', 'group1')) AND (acl.type = 1));
```

Doing two queries instead still seems more reasonable as I also couldn't think of a more performant option.

### DIFF
--- a/lib/Db/BoardMapper.php
+++ b/lib/Db/BoardMapper.php
@@ -106,28 +106,36 @@ class BoardMapper extends QBMapper implements IPermissionMapper {
 	}
 
 	public function findBoardIds(string $userId): array {
+		// Owned by the user
 		$qb = $this->db->getQueryBuilder();
 		$qb->selectDistinct('b.id')
 			->from($this->getTableName(), 'b')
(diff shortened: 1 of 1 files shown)


mejo- (commented):
To be honest I don't understand the change. If I read the code changes correctly, one SQL query is split into two now, but the second new query looks exactly like the old one to me.

The first new query doesn't have the "Shared to [user/groups/circles]" condition, so it returns only the owned boards. But the second still has both conditions: owned boards and shared boards.

Maybe you meant to remove the "Owned by the user" condition from the second query? :thinking: 

It's well possible that I just don't understand the code. DB queries are certainly not my field of expertise :grimacing:

mejo- (approved):
Code changes look good to me now :blush:

PR #4938, opened 2023-07-17
Fix small bugs around export/import

juliushaertl:
- fix: Properly export cards as a child element of the related stack
- fix: Avoid failing due to uninitialized acces of systemInstance

### TODO

- [ ] ...

### Checklist


juliushaertl:
Let's keep this in https://github.com/nextcloud/deck/pull/4939

Issue #4962, opened 2023-07-27
Create new deck card from smart picker


juliushaertl:
Currently we have a combined smart picker to reference a board/card or comment in deck. As discussed during feature planning we want to add the option to create a new card directly from the picker and insert it.

@jancborchardt We would need some mockup here how to best integrate it, especially since creating a card will always require defining a board and list to put the new card in.

Following are a few screenshots of the current picker, maybe it makes sense to consider redesigning parts of it as well. It is all based on the unified search providers but this would be an opportunity to redesign parts of it as well.

<img width="389" alt="Screenshot 2023-07-27 at 20 43 47" src="https://github.com/nextcloud/deck/assets/3404133/1f9852ce-9d92-428c-86d7-72595159441e">
<img width="634" alt="Screenshot 2023-07-27 at 20 42 23" src="https://github.com/nextcloud/deck/assets/3404133/bb699899-d3b5-4da3-bcf9-2754303726da">
<img width="637" alt="Screenshot 2023-07-27 at 20 42 30" src="https://github.com/nextcloud/deck/assets/3404133/cd808994-902b-44f6-9a1d-a018e88a9938">
<img width="633" alt="Screenshot 2023-07-27 at 20 42 50" src="https://github.com/nextcloud/deck/assets/3404133/fa57a3bc-c62a-4acd-8661-ef23bdefeb64">

juliushaertl:
From the discussion with @jancborchardt 

Use the existing card dialog layout with some extensions/adjustments:
- Separate smart picker entry for creating new cards
- Only show card details view without the other tabs
- Add input for the title
- Board/stack selector below the title as multiselect
  - Pre-select first list if board is changed

juliushaertl:
Quick mockup:

![Screenshot 2023-08-01 at 08 30 36](https://github.com/nextcloud/deck/assets/3404133/f9fbb068-f3ff-4f5d-9e6f-c97a02157961)

- We can probably make the modal a bit narrower so that it has the maximum width of the text area
- Should there be labels or icons added to the board and list selector?

@jancborchardt Can you double check that before we dive into implementation?

jancborchardt:
@juliushaertl bit narrower sounds good, otherwise seems nice! :)

PR #4969, opened 2023-08-04
feat: create new card from smart picker

luka-nextcloud:
* Resolves: #4962
* Target version: main

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/src/init-reference.js
+++ b/src/init-reference.js
@@ -19,7 +19,7 @@
  * along with this program. If not, see <http://www.gnu.org/licenses/>.
  */
 
-import { registerWidget } from '@nextcloud/vue/dist/Components/NcRichText.js'
+import { registerWidget, registerCustomPickerElement, NcCustomPickerRenderResult } from '@nextcloud/vue/dist/Components/NcRichText.js'
 import { Tooltip } from '@nextcloud/vue'
 import Vue from 'vue'
 import CardReferenceWidget from './views/CardReferenceWidget.vue'
@@ -35,6 +35,11 @@ Vue.prototype.n = translatePlural
 Vue.prototype.OC = window.OC
 Vue.prototype.OCA = window.OCA
 Vue.directive('tooltip', Tooltip)
+Vue.directive('focus', {
+	inserted(el) {
+		el.focus()
+	},
+})
 
 registerWidget('deck-card', (el, { richObjectType, richObject, accessible }) => {
 	// trick to change the wrapper element size, otherwise it always is 100%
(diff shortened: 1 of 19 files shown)


juliushaertl:
As discussed during the call:


  
- [ ] Put card title to the top
- [ ] Board picker and list picker side by side
- [ ] Make `CardSidebarTabDetails` sections reusable 
  - [ ] Migrate `CreateNewCardCustomPicker` to use a boilerplate card object as well instead of individual duedate/selectedLabels/.. data attributes
  - [ ] Extract tag multiselect section with icon from `CardSidebarTabDetails` to a sepeate component `TagSelector`
  		- those get a card object passed `<DueDateSelector :card="card" />`
  	- [ ] Use the `TagSelector` in `CardSidebarTabDetails`
  	- [ ] Use the `TagSelector` in `CreateNewCardCustomPicker`
  - [ ] Do the same for `AssignmentSelector`, `DueDateSelector`
  - [ ] Use Description component in `CreateNewCardCustomPicker`
  
As the last step:		
		
- [ ] Merge CreateCardDialog and CreateNewCardCustomPicker (only use the new one with all options)
  - Old create card dialog can just have a prop to show the "Card was added to the board" info or emit the event for the smart picker

src/views/CreateNewCardCustomPicker.vue
```diff
@@ -0,0 +1,301 @@
...
+					:can-edit="!loading && selectedBoard"
+					@select="onSelectUser"
+					@remove="onRemoveUser" />
+			</div>
+
+			<div class="section-details">
+				<DueDateSelector :card="card" :can-edit="canEdit" @change="updateCardDue" />
+			</div>
+
+			<textarea v-model="pendingDescription" :disabled="loading || !selectedStack" :placeholder="t('deck', 'Card description')" />
```
juliushaertl, line 65: This should switch to use the `Description` component so we also have the rich text editor

src/views/CreateNewCardCustomPicker.vue
```diff
@@ -0,0 +1,301 @@
...
+						:placeholder="t('deck', 'Select a list')"
+						:options="stacksFromBoard"
+						:max-height="100"
+						:disabled="loading || !selectedBoard"
+						class="multiselect-list"
+						label="title" />
+				</div>
+			</div>
+
+			<TagSelector :card="card"
```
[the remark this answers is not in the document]
juliushaertl: > maybe we can add a CardDetailRow component which has a slot for the selector and a slot for the icon to render them like this:

I think we can even just add the icons directly to the individual components, probably no need for a separate wrapper then

src/views/CreateNewCardCustomPicker.vue
```diff
@@ -0,0 +1,340 @@
+<template>
+	<div class="modal-scroller">
+		<div v-if="!creating && !created" id="modal-inner" :class="{ 'icon-loading': loading }">
+			<h3>{{ t('deck', 'Create a new card') }}</h3>
+
+			<input v-model="card.title"
```
juliushaertl, line 6: This input should probably have an autofocus :)

src/init-reference.js
```diff
@@ -82,3 +82,19 @@ registerWidget('deck-comment', (el, { richObjectType, richObject, accessible })
 		},
 	}).$mount(el)
 })
+
+registerCustomPickerElement('create-new-deck-card', async (el, { providerId, accessible }) => {
```
juliushaertl, line 86: Let's also pass `normal` for the size here to match the requirements from https://github.com/nextcloud/deck/issues/4962#issuecomment-1661721341 

Reference: https://github.com/nextcloud-libraries/nextcloud-vue/blob/8dbc4451dfef7b0edf7ed438f26b2064ffddd113/src/components/NcRichText/NcReferencePicker/customPickerElements.js#L28

luka-nextcloud:
@juliushaertl I've done the following things:

- add icons before selectors input
- add prop autofocus for title input
- pass size param 'normal' while calling registerCustomPickerElement()
- fix closing create card modal

juliushaertl (approved):

juliushaertl:
@luka-nextcloud FYI I pushed a bit more polishing, to reuse styles and also use the same component for creating files from the talk app 

Latest state:

Screenshot 2023-08-09 at 13 35 00

PR #5035, opened 2023-08-18
Card Cover Images

jszeibert:
* Resolves: #1635
* Target version: main

### Summary
This is my simple implementation of card cover images, inspired by the CoverImage feature of deck Android app:

I placed the option alongside compact mode and it enabled/disables the option globaly (even for the upcoming cards view):
CardCover-option

The normal board view displays up to 3 images sorted by id (same as in the AttachmentList) to get the latest. This is the same with the Android app.
CardCover-boardView

In compact mode I'm deviating from the android app because of a different column width. Adding one cover immage to the left of the card (like proposed in https://github.com/nextcloud/deck/pull/2858) reduces space for the title dramatically.
CardCover-compactView

I opted to allow all mime-types that generate a preview to be displayed as card covers. This may include mp3 files with album covers and textfiles. With textfiles this could result in a seemingly blank preview image, as they are centered and if there are to few lines they might be cropped.

Last mentionable thing about my implementation would be that the core preview provider behaves a bit strange providing much larger than needed images when cropping is enabled. Therefore I use the option a=1 to not crop the image and let css handle the overflow.

Finaly, this is the mobile view:
CardCover-mobileView

### TODO

- [ ] translations

### Checklist

### DIFF
--- a/src/components/cards/CardCover.vue
+++ b/src/components/cards/CardCover.vue
@@ -65,7 +65,9 @@ export default {
 		cardId: {
 			immediate: true,
 			handler() {
-				this.fetchAttachments(this.cardId)
+				if (this.$store.getters.cardById(this.cardId)?.attachmentCount > 0) {
+					this.fetchAttachments(this.cardId)
+				}
 			},
 		},
 	},
--- a/src/components/cards/CardCover.vue
+++ b/src/components/cards/CardCover.vue
@@ -1,0 +1,100 @@
+<!--
+  - @copyright Copyright (c) 2023 Johannes Szeibert <johannes@szeibert.de>
+  -
+  - @author Johannes Szeibert <johannes@szeibert.de>
+  -
+  - @license GNU AGPL version 3 or any later version
+  -
+  - This program is free software: you can redistribute it and/or modify
+  - it under the terms of the GNU Affero General Public License as
+  - published by the Free Software Foundation, either version 3 of the
+  - License, or (at your option) any later version.
+  -
+  - This program is distributed in the hope that it will be useful,
+  - but WITHOUT ANY WARRANTY; without even the implied warranty of
+  - MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  - GNU Affero General Public License for more details.
+  -
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+  -
+  -->
+
+<template>
+	<div v-if="cardId && ( attachments.length > 0 )" class="card-cover">
+		<div v-for="(attachment, index) in attachments"
+			:key="attachment.id"
+			:class="['image-wrapper', { 'rounded-left': index === 0 }, { 'rounded-right': index === attachments.length - 1 }]"
+			:style="{ backgroundImage: `url(${attachmentPreview(attachment)})` }" />
+	</div>
+</template>
+<script>
+import { mapActions } from 'vuex'
+import { generateUrl } from '@nextcloud/router'
+export default {
+	name: 'CardCover',
+	props: {
+		cardId: {
+			type: Number,
+			required: true,
+		},
+	},
+	computed: {
+		attachments() {
+			return [...this.$store.getters.attachmentsByCard(this.cardId)]
+				// Filter deleted and hasPreview
+				.filter(attachment => attachment.deletedAt >= 0 && attachment.extendedData.hasPreview)
+				// sort by id (same as in AttachmentList) to get Newest
+				.sort((a, b) => b.id - a.id)
+				// limit to 3 like with android Deck app
+				.slice(0, 3)
+		},
+		attachmentPreview() {
+			// FIXME find a better way to get the stack-width
+			const stackWidth = getComputedStyle(document.documentElement).getPropertyValue('--stack-width').trim()
+			const x = Math.ceil(parseInt(stackWidth) / this.attachments.length) | 260
+			const y = 100
+			return attachment => (
+				// The core preview provider is a bit strange at times, providing much larger than needed images
+				// when cropping is enabled. Therefore use a=1 to not crop the image and let css handle the overflow
+				attachment.extendedData.fileid ? generateUrl(`/core/preview?fileId=${attachment.extendedData.fileid}&x=${x}&y=${y}&a=1`) : null
+			)
+		},
+	},
+	watch: {
+		cardId: {
+			immediate: true,
+			handler() {
+				this.fetchAttachments(this.cardId)
+			},
+		},
+	},
+	methods: {
+		...mapActions([
+			'fetchAttachments',
+		]),
+	},
+}
+</script>
+
+<style lang="scss" scoped>
+@import '../../css/variables';
+
+.card-cover {
+	height: 100px;
+	display: flex;
+	.image-wrapper {
+		flex: 1;
+		position: relative;
+		background-size: cover;
+		background-repeat: no-repeat;
+		background-position: center center;
+		&.rounded-left {
+			border-top-left-radius: 10px;
+		}
+		&.rounded-right {
+			border-top-right-radius: 10px;
+		}
+	}
+}
+</style>
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -208,6 +208,12 @@
 						<ArrowCollapseVerticalIcon slot="icon" :size="20" decorative />
 						{{ t('deck', 'Toggle compact mode') }}
 					</NcActionButton>
+					<NcActionButton @click="toggleShowCardCover">
+						<template #icon>
+							<ImageIcon :size="20" decorative />
+						</template>
+						{{ showCardCover ? t('deck', 'Hide card cover images') : t('deck', 'Show card cover images') }}
+					</NcActionButton>
 				</NcActions>
 				<!-- FIXME: NcActionRouter currently doesn't work as an inline action -->
 				<NcActions>
@@ -226,6 +232,7 @@ import { mapState, mapGetters } from 'vuex'
 import { NcActions, NcActionButton, NcAvatar, NcButton, NcPopover, NcModal } from '@nextcloud/vue'
 import labelStyle from '../mixins/labelStyle.js'
 import ArchiveIcon from 'vue-material-design-icons/Archive.vue'
+import ImageIcon from 'vue-material-design-icons/ImageMultiple.vue'
 import FilterIcon from 'vue-material-design-icons/Filter.vue'
 import FilterOffIcon from 'vue-material-design-icons/FilterOff.vue'
 import ArrowCollapseVerticalIcon from 'vue-material-design-icons/ArrowCollapseVertical.vue'
@@ -245,6 +252,7 @@ export default {
 		NcPopover,
 		NcAvatar,
 		ArchiveIcon,
+		ImageIcon,
 		FilterIcon,
 		FilterOffIcon,
 		ArrowCollapseVerticalIcon,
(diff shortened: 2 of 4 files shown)


juliushaertl:
Thanks for your pull request, this looks very nice already. I'll try to get to a full review during the week.

For the UI, let me already pull in @nextcloud/designers for some review on the screenshots.

Just a quick note, could you use rebase instead of merge for updating the branch to the latest main branch for a cleaner history? :)

juliushaertl:
Translations do not need to be done in the pull request as long as all code parts use the proper `t()` or `n()` functions. Once merged the new strings will be added to transifex and be handled by translators automatically.

src/components/cards/CardCover.vue
```diff
@@ -0,0 +1,99 @@
...
+			return [...this.$store.getters.attachmentsByCard(this.cardId)]
+				// Filter deleted and hasPreview
+				.filter(attachment => attachment.deletedAt >= 0 && attachment.extendedData.hasPreview)
+				// sort by id (same as in AttachmentList) to get Newest
+				.sort((a, b) => b.id - a.id)
+				// limit to 3 like with android Deck app
+				.slice(0, 3)
+		},
+		attachmentPreview() {
+			// FIXME find a better way to get the stack-width
```
juliushaertl, line 53: Seems like a reasonable way to me, I'd be ok with that. We cannot use the current element as it might not be shown at time of calculation

src/components/cards/CardCover.vue
```diff
@@ -0,0 +1,99 @@
...
+				// when cropping is enabled. Therefore use a=1 to not crop the image and let css handle the overflow
+				attachment.extendedData.fileid ? generateUrl(`/core/preview?fileId=${attachment.extendedData.fileid}&x=${x}&y=${y}&a=1`) : null
+			)
+		},
+	},
+	watch: {
+		cardId: {
+			immediate: true,
+			handler() {
+				this.fetchAttachments(this.cardId)
```
juliushaertl, line 68: This is quite bad for performance as it will issue one request per card.

Maybe we need to extend the existing card listing request to return the attachments but might also need some care to not run into performance regressions

juliushaertl (changes requested):
I left some comments inline, generally this seems very nice also from a code perspective, with the performance flaw that we'd still need to figure out. I'll think a bit about that.

One small bug found during testing:
- [ ] When uploading an image attachment to a card the CardCover component loads but no image is shown

marcoambrosini (approved):
Very nice design wise @jszeibert. I don't have comments to add to this.

As an additional feature, it would be nice to be able to select one specific image as a card cover, but something for the future maybe.

juliushaertl:
Mostly for reference, I did a quick patch to list the attachments with the initial request but that has quite some impact due to the required file system setup.



<details>
<summary>Patch (probably not the way to go)</summary>

```diff
~/repos/nextcloud/server/apps-extra/deck enh/CardCover✦ ➜ git diff
diff --git a/lib/Service/CardService.php b/lib/Service/CardService.php
index 45559b87..deaed044 100644
--- a/lib/Service/CardService.php
+++ b/lib/Service/CardService.php
@@ -126,6 +126,7 @@ class CardService {
                        $this->cardMapper->mapOwner($card);
 
                        $card->setAttachmentCount($this->attachmentService->count($cardId));
+                       $card->setAttachments($this->attachmentService->findAll($cardId));
 
                        // TODO We should find a better way just to get the comment count so we can save 1-3 quer
ies per card here
                        $countComments = $this->commentsManager->getNumberOfCommentsForObject('deckCard', (string
)$card->getId());
diff --git a/lib/Service/StackService.php b/lib/Service/StackService.php
index ce3e1195..9eaa97eb 100644
--- a/lib/Service/StackService.php
+++ b/lib/Service/StackService.php
@@ -129,6 +129,7 @@ class StackService {
                                $assignedUsers = $this->assignedUsersMapper->findAll($card->getId());
                                $card->setAssignedUsers($assignedUsers);
                                $card->setAttachmentCount($this->attachmentService->count($card->getId()));
+                               $card->setAttachments($this->attachmentService->findAll($card->getId()));
 
                                return new CardDetails($card);
                        },
(END)

```
</details>

Blackfire profile for comparison https://blackfire.io/profiles/compare/b491c795-67e9-444a-ba75-ca5cbc866436/graph

src/components/cards/CardCover.vue
juliushaertl: Mostly for reference, I did a quick patch to list the attachments with the initial request but that has quite some impact due to the required file system setup.



<details>
<summary>Patch (probably not the way to go)</summary>

```diff
~/repos/nextcloud/server/apps-extra/deck enh/CardCover✦ ➜ git diff
diff --git a/lib/Service/CardService.php b/lib/Service/CardService.php
index 45559b87..deaed044 100644
--- a/lib/Service/CardService.php
+++ b/lib/Service/CardService.php
@@ -126,6 +126,7 @@ class CardService {
                        $this->cardMapper->mapOwner($card);
 
                        $card->setAttachmentCount($this->attachmentService->count($cardId));
+                       $card->setAttachments($this->attachmentService->findAll($cardId));
 
                        // TODO We should find a better way just to get the comment count so we can save 1-3 quer
ies per card here
                        $countComments = $this->commentsManager->getNumberOfCommentsForObject('deckCard', (string
)$card->getId());
diff --git a/lib/Service/StackService.php b/lib/Service/StackService.php
index ce3e1195..9eaa97eb 100644
--- a/lib/Service/StackService.php
+++ b/lib/Service/StackService.php
@@ -129,6 +129,7 @@ class StackService {
                                $assignedUsers = $this->assignedUsersMapper->findAll($card->getId());
                                $card->setAssignedUsers($assignedUsers);
                                $card->setAttachmentCount($this->attachmentService->count($card->getId()));
+                               $card->setAttachments($this->attachmentService->findAll($card->getId()));
 
                                return new CardDetails($card);
                        },
(END)

```
</details>

Blackfire profile for comparison https://blackfire.io/profiles/compare/b491c795-67e9-444a-ba75-ca5cbc866436/graph

juliushaertl: Might be ok if we just do it as a separate request then as card cover images would not be default, so users could opt in.

jszeibert:
> Thanks for your pull request, this looks very nice already. I'll try to get to a full review during the week.
> 
> For the UI, let me already pull in @nextcloud/designers for some review on the screenshots.
> 
> Just a quick note, could you use rebase instead of merge for updating the branch to the latest main branch for a cleaner history? :)

Sorry for that. I used the github feature to sync main. Would you like me to rebase and force push to correct this?

src/components/cards/CardCover.vue
jszeibert: I wasn't sure if there was a better way, ...
If its ok with you I would remove the comment then.

src/components/cards/CardCover.vue
```diff
@@ -0,0 +1,99 @@
...
+  - MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  - GNU Affero General Public License for more details.
+  -
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+  -
+  -->
+
+  <template>
+	<div v-if="cardId && ( attachments.length > 0 )" class="card-cover">
```
jszeibert, line 24:
```suggestion
	<div v-if="cardId && (attachments.length > 0)" class="card-cover">
```

juliushaertl:
> Sorry for that. I used the github feature to sync main. Would you like me to rebase and force push to correct this?

No worries, yes that would be appreciated.

Github also has this function nowadays but well hidden and no option to make this the default per repository unfortunately:

Screenshot 2023-08-22 at 12 49 33

Suleman-Elahi:
Will this be merged anytime soon 😅

jszeibert:
I was finally able to do some digging about the attachment thumbnails or cover images sometimes not showing when uploading an image attachment to a card. This seems to be a nextcloud server core issue that is known since version 25.

see https://github.com/nextcloud/server/issues/36463

The error message above is misleading as this isn't a permission issue but rather an incorrect error handling. The folder for the preview image can't be created, as it is already present. I added some debug output to the core to trace where the real error occurs (please disregard line numbers below as they might have shifted due to my debugging):

```
{ ...
   "method":"GET",
   "url":"/core/preview?fileId=558&x=260&y=100&a=1",
   "message":"mkdir(): File exists at lib/private/Files/Storage/Local.php#110",
...}
```

```
0 lib/private/Files/Storage/Wrapper/Wrapper.php(84): OC\Files\Storage\Local->mkdir('appdata_oc491gy...') 
#1 lib/private/Files/View.php(1148): OC\Files\Storage\Wrapper\Wrapper->mkdir('appdata_oc491gy...') 
#2 lib/private/Files/View.php(243): OC\Files\View->basicOperation('mkdir', '/appdata_oc491g...', Array) 
#3 lib/private/Files/Node/Folder.php(161): OC\Files\View->mkdir('/appdata_oc491g...') 
#4 lib/private/Files/AppData/AppData.php(149): OC\Files\Node\Folder->newFolder('1/b/b/9/1/f/7/5...') 
#5 lib/private/Preview/Storage/Root.php(74): OC\Files\AppData\AppData->newFolder('1/b/b/9/1/f/7/5...') 
#6 lib/private/Preview/Generator.php(643): OC\Preview\Storage\Root->newFolder('558') 
#7 lib/private/Preview/Generator.php(139): OC\Preview\Generator->getPreviewFolder(Object(OC\Files\Node\File)) 
#8 lib/private/Preview/Generator.php(116): OC\Preview\Generator->generatePreviews(Object(OC\Files\Node\File), Array, 'image/jpeg') 
#9 lib/private/PreviewManager.php(192): OC\Preview\Generator->getPreview(Object(OC\Files\Node\File), 260, 100, false, 'fill', NULL) 
#10 core/Controller/PreviewController.php(144): OC\PreviewManager->getPreview(Object(OC\Files\Node\File), 260, 100, false, 'fill') 
#11 core/Controller/PreviewController.php(113): OC\Core\Controller\PreviewController->fetchPreview(Object(OC\Files\Node\File), 260, 100, true, true, 'fill') 
#12 lib/private/AppFramework/Http/Dispatcher.php(230): OC\Core\Controller\PreviewController->getPreviewByFileId(558, 260, 100, true, true, 'fill') 
#13 lib/private/AppFramework/Http/Dispatcher.php(137): OC\AppFramework\Http\Dispatcher->executeController(Object(OC\Core\Controller\PreviewController), 'getPreviewByFil...') 
...
```

In this case `OC\Files\Storage\Local->mkdir` returns `false` which is passed down the stack to `OC\Files\Node\Folder->newFolder` throwing the misleading `OCP\\Files\\NotPermittedException` (see https://github.com/nextcloud/server/blob/master/lib/private/Files/Node/Folder.php#L162)

----

The issue is not persistent, as a refresh triggers the preview generation again and succeeds. Furthermore, I can't reproduce the issue on my new dev environment running nextcloud 28 server master build.

juliushaertl:
I think there is some race condition as two different requests attempt to get the preview and probably both try to generate it for the first time.

Screenshot 2023-09-12 at 15 46 35

We could probably avoid this by having a blocking preview fetch request during the attachment upload before we add the attachment to the vuex store as a temporary fix. Other than that this is probably something to rather investigate in https://github.com/nextcloud/server/ to have some locking around that to avoid this kind of case in the preview generator.

src/components/cards/CardCover.vue
juliushaertl: I pushed a small commit to only request attachments if there are any, which is information we already have around.

Let me merge this already then since it is opt in and we can follow up on backend side optimization of the requests.

juliushaertl (approved):

Issue #5031, opened 2023-08-17
Remember last board/list for new card dialog


juliushaertl:
Small enhancement task for the create card dialog:

- We should remember the last board and stack used in the create card dialog
- If none is previously remembered we fall back to the least recently edited one
- Store this in the browser localStorage

luka-nextcloud:
@juliushaertl 
> If none is previously remembered we fall back to the least recently edited one

Do you mean recently edited board or card?

juliushaertl:
Recently edited board and list, the lastModified property should be good to use as it always updates if something's changes.

PR #5046, opened 2023-08-22
feat: remember last board, list for new card dialog

luka-nextcloud:
* Resolves: #5031
* Target version: main

### TODO

- [ ] ...

### Checklist


src/views/CreateNewCardCustomPicker.vue
```diff
@@ -295,6 +298,31 @@ export default {
 		openNewCard() {
 			window.location = generateUrl('/apps/deck') + `#/board/${this.selectedBoard.id}/card/${this.newCard.id}`
 		},
+		preSelectBoard() {
+			const selectedBoardId = Number(localStorage.getItem('selectedBoardId'))
```
juliushaertl, line 302: Please prefix the storage keys with deck. similar how we do it in other places:

```
store/main.js
63:		navShown: localStorage.getItem('deck.navShown') === null || localStorage.getItem('deck.navShown') === 'true',
64:		compactMode: localStorage.getItem('deck.compactMode') === 'true',
223:			localStorage.setItem('deck.navShown', navState)
230:			localStorage.setItem('deck.compactMode', state.compactMode)
```

It would actually be nice to switch to https://github.com/nextcloud-libraries/nextcloud-browser-storage which allows scoping  local storage, but I'd be ok to skip that as we'd need to move all places then.

juliushaertl (changes requested):
Small comment regarding the local storage keys. Otherwise this looks good. Maybe you can have a look a the cypress test failure, that looks like it could be related

src/views/CreateNewCardCustomPicker.vue
juliushaertl: Example usages:
https://github.com/nextcloud/notifications/blob/7c48081354bd179bdc3e51bc3b5c5ce556efc022/src/services/BrowserStorage.js#L21C1-L23C77
https://github.com/nextcloud/notifications/blob/7c48081354bd179bdc3e51bc3b5c5ce556efc022/src/services/notificationsService.js#L49

luka-nextcloud:
@juliushaertl Please check again.

juliushaertl (approved):

Issue #5086, opened 2023-09-04
Card edit modal does not show up when access from direct url


luka-nextcloud:
#### Description
Card edit modal does not show up when access from direct url

#### Steps to reproduce: 
1. Open any card
2. Reload browser
#### Expected behaviour:
Card edit modal show up
#### Actual behaviour:
Card edit modal does not show up

gwnmikedus:
Hmm, my installation 27.0.2 does seem to work. Opened different cards on different boards and reload works just fine. I see the url change to the card and when reloading it re-opens the modal.

PR #5089, opened 2023-09-04
fix: open card with direct link

luka-nextcloud:
* Resolves: #5086
* Target version: main

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/src/components/card/CardSidebar.vue
+++ b/src/components/card/CardSidebar.vue
@@ -145,18 +145,17 @@ export default {
 			titleEditing: '',
 			hasActivity: capabilities && capabilities.activity,
 			locale: getLocale(),
+			currentCard: null,
 		}
 	},
 	computed: {
 		...mapState({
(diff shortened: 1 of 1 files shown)


luka-nextcloud:
@max-nextcloud Could you please review? Thanks.

juliushaertl:
@luka-nextcloud I fail to reproduce the issue this is supposed to fix. I would expect the previous code to already work as expected. Can you tell if you can still see this issue on the main branch and maybe provide some detailed steps you take to reproduce?

juliushaertl:
Maybe it is just something on stable27/stable26 ...

luka-nextcloud:
> Maybe it is just something on stable27/stable26 ...

I've tested on them, I cannot reproduce.

juliushaertl:
Let's close this then for now. We can reconsider the fix if we manage to reproduce or run into this again somehow.

Issue #4902, opened 2023-07-10
[1.9.2] Export doesn't handle lists with no cards


magikmw:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Describe the bug**
Export function doesn't output anything if the board being exported contains a list without any cards assigned.
Issue seems to be similar to the one solved here: #4566

**To Reproduce**
1. Go to Deck
2. Create a new deck containing at least two lists, and leave one of them empty
3. Try exporting via left panel 3 dot menu -> Export board
4. Nothing, except for an error in browser console:

```
Uncaught (in promise) TypeError: e.cards is undefined
    ht BoardApi.js:156
    ht BoardApi.js:156
    promise callback*value BoardApi.js:150
    actionExport AppNavigationBoard.vue:320
    VueJS 4
    onClick index.module.js:2
    VueJS 33
```

**Expected behavior**
Should export a board into CSV file.

**Screenshots**
n/a

**Client details:**
 - OS: Windows 10 22H2
 - Browser Firefox
 - Version 114.02
 - Device: desktop

I've left out server details, because I'm fairly sure it's JS related. Log in reproduction steps.

Using NC 26.0.3 and Deck 1.9.2

PR #5096, opened 2023-09-08
fix: export doesn't handle lists with no cards

magikmw:
* Resolves: #4902 
* Target version: main

### Summary

If statement prevents the script from crashing in case there's a stack without any cards.
I would love if someone took a look at it, I'm not a programmer by trade. 
I don't know if this requires tests, or even how to write them at this point.

One thing this still doesn't handle is exporting information about the empty stack even existing.
On import it cannot be recreated.
Probably something for a different PR though.

If needed, I'll update.

### TODO

- [x] ...

### Checklist

### DIFF
--- a/src/services/BoardApi.js
+++ b/src/services/BoardApi.js
@@ -163,7 +163,7 @@ export class BoardApi {
 					let CSV = row + '\r\n'
 
 					response.data.stacks.forEach(stack => {
-						stack.cards.forEach(card => {
+						stack?.cards?.forEach(card => {
 							row = ''
 							Object.keys(fields).forEach(field => {
 								if (field === 'createdAt' || field === 'lastModified') {


magikmw:
I've fixed the `if(` linter's issue, but I have no idea what's that other problem about.
Also rebase to current main.

src/services/BoardApi.js
```diff
@@ -163,30 +163,32 @@ export class BoardApi {
...
-									row += '"' + stack.title + '"' + '\t'
-								} else if (field === 'labels') {
-									row += '"'
-									card[field].forEach(label => {
-										row += label.title + ', '
-									})
-									if (card[field].length > 0) {
-										row = row.slice(0, -1)
+						if (stack.cards) {
+							stack.cards.forEach(card => {
```
juliushaertl, line 167: Thanks a lot for your contribution.

I would suggest to avoid the extra if branch. We can use optional chaining to not call forEach if cards is not set:

```js
stack?.cards?.forEach(card => {
```

That way the change is much smaller and less nesting means less complexity when reading the code.

juliushaertl:
Fine for me without tests for now as we don't have the CSV feature covered at all yet, I might push them separately

src/services/BoardApi.js
magikmw: Sure, I've made the change - I was not aware of this syntax.

juliushaertl (approved):

juliushaertl:
Thanks a lot for the fix :)

juliushaertl:
Ah only there on 26 so we can skip the stable26 backport

PR #5260, opened 2023-11-06
fix: Avoid too large index on postgres as indexing just the last_editor column is enough

juliushaertl:
Postgres as limitations on the index size and the length is actually not applied there. This drops the index and recreates one just with the first column. As we always set or reset the last_editor together with the description_prev column having the index only for the small column is enough to query the relevant entries in the mapper.

Otherwise updating might fail with the following error:

> 
Screenshot 2023-11-06 at 17 37 31

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -16,7 +16,7 @@
 - 🚀 Get your project organized
 
 </description>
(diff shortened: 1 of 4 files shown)


mejo- (approved):
Code changes look sensible to me :+1:

PR #5262, opened 2023-11-07
fix: Check both card reference url patterns

juliushaertl:
The browser URL patterns were unused

Fixes that and adds a phpunit test


mejo- (approved):
Nice, kudos for the test :heart:

Issue #1672, opened 2020-04-09
Wrap text properly around due date


juliushaertl:
- Wrap text around due date
- Make sure that editing is working multiline https://github.com/nextcloud/deck/issues/865
- Set a max-width for the due date

**Additional context**
![2020-06-23_15:31:12](https://user-images.githubusercontent.com/8270201/85409698-6ab8a980-b566-11ea-82f5-4cbe2fd3d579.png)
![2020-06-23_15:36:12](https://user-images.githubusercontent.com/8270201/85410219-1bbf4400-b567-11ea-8ec9-27d2fe1f257f.png)

armaccloud:
Hi,

Extending on my previous remarks in #2081 I would like to chip in with some suggestions.

As you can see from this screenshot, the cards are beautifully formatted when there is no due date set. Especially in Dutch, you can see the due date has a long name, making the overall view look very untidy and difficult to navigate when there are many items (due to the lack of aesthetics, one tends to lose overview). This will partially be solved if the text wraps around the due date. However, this will still make the first line(s) look cramped as you can see in the screenshot.

![image](https://user-images.githubusercontent.com/9272078/87522950-8eaf6c80-c686-11ea-8354-14ba72264c7e.png)

What I noticed, is that there is not only the subject + due date on one line, but also the menu button.

- Would it be an option to move the menu button to the right-hand bottom?
- Is it an idea to decrease the height of the due date to its text size, so the subject text wraps around it for the second line?
- Another (my preferred) option would be to give the due date and subject their own line and decrease the height of the due date to its text size so it doesn't eat up so much space.

It would be nice to see a mock-up of the new view with the deck cards from my screenshot as an example, how would they look good in the new view? If it works for this use case, it would probably work well for other peoples decks as well.

Hanskes:
Any progress??

PR #5264, opened 2023-11-10
Card layout polishing

juliushaertl:
- fix: Fix spacing and reserved scrollbar space
- fix: Align card details better to take less space and align all possible badges more equally
  - Do not take extra space for the menu button
  - Give the title as much available space in width as possible

@nextcloud/designers Would appreciate some feedback. One thing I'm unsure is that the labels and the due date badge have a similar design, but the rounded corners seem for fitting than the previous style.

Fixes https://github.com/nextcloud/deck/issues/1672
Fixes https://github.com/nextcloud/deck/issues/4299

| Before | After |
|---|---|
| Screenshot 2023-11-10 at 11 51 22 | Screenshot 2023-11-10 at 15 59 12 |


nimishavijay (commented):
Super nice! Looks so much better :)
Only 3 small details:
- increase the horizontal padding of the labels by a couple of pixels, similar to the due date indicator
- the number of sub tasks indicator (☑️ 1/3) seems to be slightly lower than it should be, it should be center aligned with the avatars
- we could also think about reducing the bottom padding of the card itself by a couple of pixels. The new screenshots definitely look better than the old one, but it could be reduced by 2-3 pixels. For example, the card with only badges seems to have some extra space at the bottom.



But those are small comments, overall it looks great! :)

juliushaertl:
Thanks for the feedback. All addressed and also slightly updated the due date indicator to be a bit more different from the labels.

luka-nextcloud (approved):

PR #5287, opened 2023-11-14
fix: Properly get done state for dav

juliushaertl:
Otherwise this won't work as expected and throw errors like:

```
Undefined property: OCA\\Deck\\Db\\Card::$7867 at /var/www/nextcloud/apps/deck/lib/Db/Card.php#153",
```


mejo- (approved):

mejo-:
> Otherwise ***this*** won't work

:sweat_smile:

PR #5290, opened 2023-11-14
Fix upcoming cards and label input

juliushaertl:
- fix: fix copy paste error to bring back label picker
- fix: Bring back upcoming cards view by adapting to proper API return value
- test: Add test selectors and label multiselect cypress test


blizzz (approved):

juliushaertl:
Ah mixed up the fixes, not needed to be backported

Issue #5134, opened 2023-09-19
Filter by Tag when clicking on Tag is disorienting


privatemaker:
**Is your feature request related to a problem?**

When a new Deck user first clicks on a Tag in a card the filter is applied to all cards on that list.

![deck-ux-click-filter](https://github.com/nextcloud/deck/assets/49612519/7c68ec93-2b7f-4d82-86b4-af36485bc655)

All other cards seemingly disappear very quickly and the only change to the UI is the `Unfiltered` icon changes to `Filter` 

![deck-ux-click-unfilter](https://github.com/nextcloud/deck/assets/49612519/b3de6c5b-6a6b-4000-b1ed-f6971036fd2d)

Due to the abruptness of this and the fact there is only one very small detail in the UI which changes, is easy to not notice or understand what happened.

This creates a feeling of discomfort and concern that data has been deleted or such.

**Describe the solution you'd like**

Show a stronger UI presence that filters have been applied to this view.

Perhaps a word next to the icon + a counter, like:

```
[Icon] No Filters
[Icon] 2 Filters
```

**Describe alternatives you've considered**

Visually showing Tag filters (limit 3 or so) with colored background in the topbar, perhaps on the left of the search field.

juliushaertl:
Thanks for the suggestion. We don't use counters for this anywhere in Nextcloud as far as I'm aware.

Currently we switch the icon depending on if filters are set or not:

- No filter active ![Screenshot 2023-10-10 at 09 17 30](https://github.com/nextcloud/deck/assets/3404133/fc8f122d-cc0a-4d2f-bc34-9bdbd19ab33f)
- Filter(s) active ![Screenshot 2023-10-10 at 09 17 35](https://github.com/nextcloud/deck/assets/3404133/2b4e623c-673d-4d49-a72d-4380bdd22948)

However I see that this is not enough and can easily be overseen.

Talk indicates active filters with a primary color button style, so maybe we should do the same:
<img width="236" alt="Screenshot 2023-10-10 at 09 18 27" src="https://github.com/nextcloud/deck/assets/3404133/cf324308-ec4e-4de7-acd3-f9689afd330f">

fitrahfm:
Hi @privatemaker @juliushaertl I would love to work on this

juliushaertl:
Feel free to tackle this. Pull requests are always welcome. If you have any question don't hesitate to reach out :)

privatemaker:
[...]

PR #5299, opened 2023-11-17
fix(controls): use primary when filter is active

fitrahfm:
* Resolves: #5134
* Target version: main

### Summary
Added a conditional button type and fixed filter button styles when the popover is open

### TODO

- [x] Adding a conditional button type to the filter button
- [x] Fixing filter button styles when the popover is open

### Checklist

### DIFF
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -90,7 +90,7 @@
 						<NcButton slot="trigger"
 							:title="t('deck', 'Apply filter')"
 							class="filter-button"
-							type="tertiary-no-background">
+							:type="isFilterActive ? 'primary' : 'tertiary'">
 							<template #icon>
 								<FilterIcon v-if="isFilterActive" :size="20" decorative />
 								<FilterOffIcon v-else :size="20" decorative />
@@ -509,8 +509,11 @@ export default {
 		width: 44px;
 		height: 44px;
 
-		&:hover, &:focus {
-			background-color: rgba(127,127,127,0.25) !important;
+		&[data-popper-shown] {
+			background-color: var(--color-background-hover);
+			&.button-vue--vue-primary {
+				background-color: var(--color-primary-element);
+			}
 		}
 	}
 </style>


src/components/Controls.vue
```diff
@@ -509,8 +509,14 @@ export default {
...
 
-		&:hover, &:focus {
-			background-color: rgba(127,127,127,0.25) !important;
+		&[data-popper-shown] {
+			background-color: rgb(245, 245, 245);
+			[data-themes=dark] & {
+				background-color: #212121;
+			}
+			&.button-vue--vue-primary {
+				background-color: var(--color-primary-element);
```
juliushaertl, line 518: This should already be set through the type itself

src/components/Controls.vue
```diff
@@ -509,8 +509,14 @@ export default {
 		width: 44px;
 		height: 44px;
 
-		&:hover, &:focus {
-			background-color: rgba(127,127,127,0.25) !important;
+		&[data-popper-shown] {
+			background-color: rgb(245, 245, 245);
+			[data-themes=dark] & {
+				background-color: #212121;
+			}
```
juliushaertl, line 516: Can you elaborate what that special styling would be needed for? I couldn't notice it would have any effect when testing out the PR. In general we should avoid any hard coded color values and use available css vars wherever appropriate these days.

juliushaertl:
Thanks a lot for your pull request. I left some comments inline to clarify.

src/components/Controls.vue
fitrahfm: The difference is that when the popover opens, the filter button doesn't look like it's been pressed.

With styles:
Screen Shot 2023-11-20 at 5 19 34 PM

Without styles:
Screen Shot 2023-11-20 at 5 20 06 PM

fitrahfm: I refer to Talk when the popover opens, the filter button will be like this

Screen Shot 2023-11-20 at 5 22 32 PM

src/components/Controls.vue
fitrahfm: I also need this style to override the background color of the filter button when the filter is checked

Screen Shot 2023-11-20 at 5 34 31 PM

src/components/Controls.vue
fitrahfm: After checking the existing color vars, the required style is just like this

Screen Shot 2023-11-20 at 5 56 00 PM

juliushaertl: Ah good point then, especially with those variables that seems good

fitrahfm: Thanks, I will update the code

juliushaertl (approved):

juliushaertl:
Thanks a lot for your contribution @fitrahfm

PR #5351, opened 2023-12-05
fix: Get proper rich object for card actions

juliushaertl:
When moving over the card actions to a different component the computed `cardRichObject` was missing. 

The actions are now covered with a test case registering a dummy action and asserting that its callback gets triggered with the expected result rich object when clicking the action.

We should think about enabling vue/no-undef-properties as eslint rule which could have catched this but unfortunately it throws a lot of unrelated errors with vuex mapGetters/mapState methods.

* Resolves: https://github.com/nextcloud/spreed/issues/10330
* Target version: main

### DIFF
--- a/cypress/e2e/cardFeatures.js
+++ b/cypress/e2e/cardFeatures.js
@@ -252,4 +252,52 @@ describe('Card', function() {
 
 	})
 
+	describe('Card actions', () => {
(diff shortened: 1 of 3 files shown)


nickvergessen (approved):
Works pretty well again, thanks a lot 🙇🏼

PR #5377, opened 2023-12-14
Next cloud deck table scrolling

adshrest:
Description:
Currently, when a table is created within a Deck card description, and the table width exceeds the display-box width, users have to use a touchpad to scroll horizontally. This can be inconvenient, as there are no visible horizontal scrollbars indicating additional columns.

Proposal:
I suggest implementing a horizontal scrollbar for tables within the Deck card description. This can be achieved with a simple CSS tweak, such as adding overflow-x: scroll; to the relevant style.

### DIFF
--- a/css/deck.css
+++ b/css/deck.css
@@ -7,3 +7,9 @@
 	background-image: url(../img/deck.svg);
 	filter: var(--background-invert-if-dark);
 }
+
+/* Add horizontal scrollbar for tables in card descriptions */
+.card-description table {
+	width: 100%; /* Ensure the table takes 100% width of the container */
+	overflow-x: auto; /* Enable horizontal scrolling */
+}
--- a/src/components/cards/CardItem.vue
+++ b/src/components/cards/CardItem.vue
@@ -434,6 +434,12 @@ export default {
 		}
 	}
 
+	/* Add horizontal scrollbar for tables in card descriptions */
+	.description-container table {
+    	width: 100%;
+    	overflow-x: auto;
+  	}
+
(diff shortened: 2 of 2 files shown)


css/deck.css
```diff
@@ -7,3 +7,9 @@
 	background-image: url(../img/deck.svg);
 	filter: var(--background-invert-if-dark);
 }
+
+/* Add horizontal scrollbar for tables in card descriptions */
+.card-description table {
+	width: 100%; /* Ensure the table takes 100% width of the container */
+	overflow-x: auto; /* Enable horizontal scrolling */
+}
```
juliushaertl, line 15: I don't think we need this duplicate rule over here, best would be to only adjust this in the component that renders the table which would be here: https://github.com/nextcloud/deck/blob/9e41cb162de805677f50dff34874ef0430db47de/src/components/card/Description.vue#L52

juliushaertl: Note the scoping might limit the css rule so you probably need something like
```css
.description__text:deep(table) {
```

juliushaertl (changes requested):
Thanks for your contribution. I left a comment inline about where to best apply this change. 

Please also make sure that your commits have a sign-off message and best squash the changes into one commit once done.

juliushaertl:
Hi @adshrest Would you still be up for addressing the comments? Let me know if there are any questions.

juliusknorr:
Thanks again for your contribution. I'll close this as there is no further feedback. I also checked the scroll behaviour right now and for me it seems ok on latest main branch.

Feel free to reopen this if still valid by addressing the review feedback.

Issue #5451, opened 2024-01-10
Icon for done cards in compact mode without due date


reox:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Is your feature request related to a problem? Please describe.**

The new "mark card as done" feature is great! However, the indication that a card is marked as done can only be seen in the expanded view and not in the compact view.

**Describe the solution you'd like**

A checkbox symbol could be added, such that it is possible to spot done cards in the compact view as well

**Describe alternatives you've considered**

The card could also be highlighted green. However, if something like #3592 is every implemented, that might interfere with it.

**Additional context**
Done card in normal mode:
![grafik](https://github.com/nextcloud/deck/assets/431045/947fc973-8306-4756-b013-e1207b5105b2)

Done card in compact mode:
![grafik](https://github.com/nextcloud/deck/assets/431045/c555c34d-c290-42fb-b4b4-0f4540ab8a13)

Mockup of done card in compact mode:
![grafik](https://github.com/nextcloud/deck/assets/431045/13979a80-5e65-4ea4-b8fb-63a0ebc0e864)

Other mockup using background color:
![grafik](https://github.com/nextcloud/deck/assets/431045/609e1bcb-0591-4831-a34d-610a7804d1c0)

reox:
I just noticed, that the icon is only missing if no due date was set before. If a due date was set, the card will look like this:
![grafik](https://github.com/nextcloud/deck/assets/431045/9040f761-f4e1-4567-986f-04d52cea4102)

I think there are two possible solutions:
1) Use the same icon if the card was marked as done but no due date was set
2) use a different icon in that case, to hightlight that this card was done even if no due date was set (not sure if that comes handy for anyone though)

juliushaertl:
We should just use the same icon as with a due date as in the non-compact view

PR #5455, opened 2024-01-11
Fixing #5451

mintsoft:
Resolves: https://github.com/nextcloud/deck/issues/5451

* Resolves: # 
* Target version: main

### Summary

No longer checks if it is compact before applying the due vue as it is handled internally anyway

### Checklist


juliushaertl (approved):
Thanks for the fix ❤️

PR #5458, opened 2024-01-12
fix: Add debounce to update due date

### LINKED ISSUE #5398: Typing a 0 in the date picker removes the due date
When typing a date (e.g. 01.01.2024), typing the 0 results in removing the die date; that also causes a lot of entries in the activity app.

Even if that was on purpose it doesn't seem very intuitive; mainly because the date format (01.01.) encourages to also type the 0.

juliushaertl:
Fix #5398

Not perfect but for now a quick improvement.

We could save only when leaving the input or closing the modal but that might have the risk to not save a due date unless we do larger changes to combine that with autosaving after some time and block closing the browser window on unsaved changes.

### DIFF
--- a/src/components/card/CardSidebarTabDetails.vue
+++ b/src/components/card/CardSidebarTabDetails.vue
@@ -37,7 +37,8 @@
 
 		<DueDateSelector :card="card"
 			:can-edit="canEdit"
-			@change="updateCardDue" />
+			@change="updateCardDue"
+			@input="debouncedUpdateCardDue" />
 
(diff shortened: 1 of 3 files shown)


nickvergessen (approved):

nickvergessen:
> We could save only when leaving the input or closing the modal but that might have the risk to not save a due date unless we do larger changes to combine that with autosaving after some time and block closing the browser window on unsaved changes.

Yeah I think it's an okay middle ground

Issue #534, opened 2018-07-13
Mark card as done


cloud2018:
**Situation**
Simple Board with stacks "to do", "doing" and "done" and due dates for cards. When a card is done in time the colour of the due date is grey and i move the card to the stack "done". After the due date the recalculated colour is red. There is no information, weather the card was done in time or over time, cause all my cards in stack "done" have a recalculated colour red of due date.

**Describe the solution you'd like**
I would like to see, that the card was done in time with no overdue. 

It would be helpful, if i could stop the recalculation of the colour of a specific card, Perhaps an additional button like the "remove due date" button. 

---
There is a **[$15 open bounty](https://www.bountysource.com/issues/60848444-possibility-to-fix-the-colour-of-due-date-and-stop-recalculation?utm_campaign=plugin&utm_content=tracker%2F52758431&utm_medium=issues&utm_source=github)** on this issue.

juliushaertl:
[GitMate.io](https://gitmate.io) thinks possibly related issues are https://github.com/nextcloud/deck/issues/329 (due date not working), https://github.com/nextcloud/deck/issues/106 (Attach due date to cards), https://github.com/nextcloud/deck/issues/205 (Current date should be highlighted in the due date selector), https://github.com/nextcloud/deck/issues/206 (Due date indicator shoud have title attribute), and https://github.com/nextcloud/deck/issues/324 (Description unable to saved if add Due Date).

juliushaertl:
That seems to be a pretty good idea. We should add an option to mark the card as finished in terms of the due date was made or not.

delightfuldude:
@juliushaertl thanks for this great nextcloud app!
Any progress so far? Is there an option to mark cards as done/finished?

delightfuldude:
Bump!
How to mark a card as done?

delightfuldude:
Seriously, how to mark a card as DONE??? Why is this request ignored since July 2018?

Relax-Roger:
Has this (extremely useful) improvement ever been realised in the meantime?

juliushaertl:
@Relax-Roger If it would have been implemented the issue would be closed. As you can see the issue is still labeled as "to develop" so no one has picked it up for development yet.

Dubidubiduu:
My request is not the same as @cloud2018, but tied to that topic. I don't want reminders of finished tasks, so the label "finished" should have a technical meaning of not pushing a notification if the due date is over/near.

alexanderdd:
[...]

PR #4137, opened 2022-10-18
Added ability to mark a card as done

TehThanos:
* Resolves: #534 
* Target version: master 

### Summary
This adds the ability to mark a card as done.
When the card is marked as done if it has a due date then it makes it un-editable, if it doesn't it simply displays "No due date"
Marking a card as done also hides it from the upcoming section.

### TODO

- [x] Add Rest API documentation
- [x] Design changes requested form @nimishavijay 
- [x] Remove [l10n changes](https://github.com/nextcloud/deck/pull/4137#discussion_r1000169153)
- [x] lint issues
- [x] style lint issues
- [x] document headline number wrong [here](https://github.com/nextcloud/deck/pull/4137/files#diff-43a3c216412e058983668ebe1bf88d8c7967b5d16fdc19884122d28bb9eb03e3R57)
- [ ] missing cypress tests
- [x] Move from boolean column to datetime to store that exact timestamp when a card was marked as done
  - [x] Change migration
  - [x] Properly handle the datetime in the backend
  - [x] Maybe indicate the done timestamp in the frontend on hover of the icon or in the details view next to the done label
  - [x] Expose it through the caldav backend https://github.com/nextcloud/deck/pull/4137#issuecomment-1521730474
  - [x] Double check visibility of icons in archived state
- [x] Find a proper layout for modal/sidebar for the done state and buttons

## Follow up issues
- [ ] Done by to store which user marked it (as suggested by @z0rgster)

### Checklist

- [x] Code is properly formatted
- [x] Sign-off message is added to all commits
- [ ] Tests (unit, integration, api and/or acceptance) are included
- [x] Documentation (manuals or wiki) has been updated or is not required

### Screenshots

<details>
<summary>Current state (12th September)</summary>

Screenshot 2023-09-12 at 15 42 26

Screenshot 2023-09-12 at 15 42 32

Screenshot 2023-09-12 at 15 42 35

</details>

PR #5491, opened 2024-01-18
fix(done): Mark card as undone when updating card

stefan-niedermann:
* Resolves: #534
* Target version: main

### Summary

As stated in https://github.com/nextcloud/deck/issues/534#issuecomment-1892061055 updating the done property of a card via the REST API (without calling the /done and /undone endpoints explicitly) does only work "one way".

This commit allows setting null as new value thus allowing to mark cards as undone without an additional HTTP request but within a usual update request.

Refs: #534 #4137 c3b4ed6e1fe706195176e011c6004d3a301a9458
CC: @desperateCoder @juliushaertl @TehThanos

### Checklist

### DIFF
--- a/lib/Service/CardService.php
+++ b/lib/Service/CardService.php
@@ -355,6 +355,8 @@ class CardService {
 		}
 		if ($done !== null) {
 			$card->setDone($done->getValue());
+		} else {
+			$card->setDone(null);
 		}
 
 


juliushaertl (approved):

juliushaertl:
Thanks a lot for the fix ❤️

PR #5625, opened 2024-02-27
feat: Implement a team resource provider

juliushaertl:
Implement a team resource provider for https://github.com/nextcloud/server/issues/42433


mejo- (approved):
Works well after the proposed change. CI found some formatting fixes that would need to be applied for making CI green.

mejo- (changes requested):
Ouch meant to request changes, not approve :see_no_evil:

mejo- (approved):

PR #5630, opened 2024-02-29
Interactive board widget

juliushaertl:
Implementing an interactive widget for deck boards

[Kapture 2024-02-29 at 17.59.35.webm](https://github.com/nextcloud/deck/assets/3404133/82cb4a84-b495-4169-bed9-5f494b4669f8)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,6 +13,7 @@
         "@babel/runtime": "^7.23.9",
         "@nextcloud/auth": "^2.2.1",
         "@nextcloud/axios": "^2.4.0",
+        "@nextcloud/capabilities": "^1.1.0",
         "@nextcloud/dialogs": "^4.2.6",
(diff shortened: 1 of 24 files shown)


juliushaertl:
@nextcloud/designers For feedback on the screen recording :)

jancborchardt (commented):
Wow that looks really nice!! :) All in there.

One thing I am not sure about is the height change:
- Currently it seems a bit janky when you modify or filter, the height just jumps around
- What could be better is a set height like 50vh which does not change but is scrollable. Drawback being that it also takes more when not needed.

What do you think @juliushaertl? I would also be fine with the current solution but generally it's nice if the UI feels "solid".

juliushaertl:
Yes, adjusted, much better that way. I think this is fine as the board certainly has a couple of cards most of the time.

luka-nextcloud (approved):

Issue #5235, opened 2023-10-25
Support migration of user data


privatemaker:
### Is your feature request related to a problem? Please describe.

As a user migrating from one NC to another, I should be able to bring my data with me. 

### Describe the solution you'd like

This should be done properly with the [user_migration](/nextcloud/user_migration/) app as outlined in [the docs](https://docs.nextcloud.com/server/stable/developer_manual/digging_deeper/user_migration.html)

### Describe alternatives you've considered

I figured out a hacky way to migrate data from `Instance 1: Deck / Tasks` -> `Instance 2: Tasks` and [explained here](https://help.nextcloud.com/t/data-migration-for-deck-tasks-app-data/172821). I recall about 9 months ago I was able to move data from `Tasks: Lists` -> `Deck: Boards` on the same instance within Tasks app. Am I imagining this or did some ACL / perms change somewhere?

raimund-schluessler:
Reading what you wrote in the forum, I see that you exported the Deck boards with the Tasks app. This exports from the Deck backend into a ICS file. However, importing this with the Calendar app, imports into the CalDAV backend from server, so it will not show up in the Deck app anymore.

Since this really is a Deck issue, I have to refer you to the Deck devs.

privatemaker:
> This exports from the Deck backend into a ICS file

@raimund-schluessler the ICS file I export is from `Tasks` app, but can also be from `Calendar` I suppose. Newer versions of `Deck` seem to export a CSV file, older version there is no way to export AFAICT.

> importing this with the Calendar app, imports into the CalDAV backend from server

I figured something like this is happening. So if I understand CalDAV interop between `Calendar` and `Deck` is not bi-directional as Deck Boards are immediately viewable in Calendar, but not the other way around. It just is a bit confusing from a user perspective but makes sense from a engineering perspective.

One cannot import Calendars events into Decks or tasks due to how these API's flow. 



> so it will not show up in the Deck app anymore.

Right I stated that in my forum post, hence why I opened issue here 😄

mvdw:
The problem remains: As a user migrating from one NC to another, I should be able to bring my data with me.
I succeeded to export a big JSON with all my data with the "occ deck:export" command, but how do I import this in my new nextcloud server?

NeroSec:
Its crazy, i am still not able to export/import a Deck... To import it want a json, when export in the app it gives a CSV, export with OCC gives a json, try to import this json fails because:
 `It's not a valid config file.
[...]

Issue #5213, opened 2023-10-17
deck:import misleading error message for config.json


ebertmicha:
**Describe the bug**
When using occ deck:import --config=/unreachable-location-for-www-user/config,json ...
the error message is "It's not a valid config file."

This is in my opinion misleading. 
It implies (at least to me) that there is something wrong with the parsing/formatting of the file.
But it checks the existence of the file.

```
BoardImportCommandService.php
if (!is_file($config)) {
  throw new NotFoundException('It\'s not a valid config file.');
}
```

**To Reproduce**
point occ deck:import to a not accessible file for the www user

**Expected behavior**
"File not found" or something similar.

**Nextcloud version:** (see Nextcloud admin page)
27.1.2

PR #5722, opened 2024-03-31
Change import command's config default argument

JimMadge:
* Resolves: #5235 
* Target version: main

### Summary

- Change the default argument of `deck:import --config ...` from `config.json` to `null`.
  This default value caused a problem (#5235) where if not config was supplied, `config.json` is not found and the message "It's not a valid config file" is shown (#5213)
- Change the `deck:import --data ...` option to `VALUE_REQUIRED`. From reading the [Symfony docs](https://symfony.com/doc/current/console/input.html#using-command-options) I think all options are optional. `VALUE_OPTIONAL` lets you include the option in a command line _without_ giving a value, _e.g._ `deck:import --data` would be valid.

### TODO

- [ ] ...

### Checklist

    - Is this n/a as there are no new features?
    - I think this should be consistent with the current docs, where not providing `--config` assumes the data is a deck json.

### DIFF
--- a/lib/Command/BoardImport.php
+++ b/lib/Command/BoardImport.php
@@ -43,7 +43,7 @@ class BoardImport extends Command {
 				null,
 				InputOption::VALUE_REQUIRED,
 				'Configuration json file.',
-				'config.json'
+				null
 			)
 			->addOption(
 				'data',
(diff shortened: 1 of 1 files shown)


juliushaertl (approved):
Thanks for your pull request. 👍

Issue #5823, opened 2024-05-01
Optional parameter $userId declared before required parameter $path is implicitly treated as a required parameter


modir:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Describe the bug**
I have Nextcloud 29 installed a week ago and my log keeps filled with this message:

`Optional parameter $userId declared before required parameter $path is implicitly treated as a required parameter at /path/to/nextcloud/apps/deck/lib/Service/ConfigService.php#234 `

It doesn't look like a big problem but it would be nice if it is addressed soon so that it doesn't fill the logs. I guess I get this message because I am already on PHP 8.3 and you are testing with earlier versions.

daffydock:
Just as an extra note, even just enabling the app will trigger this error, as well. However, using the mobile app and updating changes does not seem to trigger it. 

Worth noting that albeit OP is on PHP 8.3, the instance I am getting the same error is on PHP 8.1.

solracsf:
Fixed by https://github.com/nextcloud/deck/pull/5794 (yet to be released).

PR #5794, opened 2024-04-24
fix: Avoid optional before required parameter

juliushaertl:
Fix https://github.com/nextcloud/deck/issues/5793

PR #5885, opened 2024-05-12
Bugfix to Optional parameter $userId declared before required parameter

modir:
Optional parameter $userId declared before required parameter $path is implicitly treated as a required parameter

Since it is treated anyway as a required parameter we can define it as such as well.

* Resolves:  #5823
* Target version: main


juliushaertl (commented):
This has already been addressed by https://github.com/nextcloud/deck/pull/5794

The parameter should still be nullable

juliushaertl:
Thanks for your PR, I'll close this as it has already been addressed in the mentioned other PR. We'll spin a release with the fix soon.

modir:
As long as it is solved soon, I am fine.

PR #5957, opened 2024-06-01
choir(i18n): Change filter title

rakekniven:
You do not filter for completed when selecting "filter for open".

Status is a better name for all existing options


stefan-niedermann:
Not sure, `status` might also refer to the corresponding list. Recently `Completed` has been replaced with `Done`, so maybe `Filter for done state` might be more close? Granted, one would need to check whether the length causes visual issues...

juliushaertl (approved):

PR #6008, opened 2024-06-18
chore: Move comments event handler to use proper event dispatcher

juliushaertl:
Similar to https://github.com/nextcloud/server/pull/45951 this moves away from the not so lazy registration of comment event handlers.

- [x] Requires https://github.com/nextcloud/server/pull/45951/files#diff-862a3988c96ee1f18bc17076a3152274d876dec178266a6b2d237b9ae7f03cd0R1534

### Checklist

### DIFF
--- a/tests/unit/Listeners/CommentEventListenerTest.php
+++ b/tests/unit/Listeners/CommentEventListenerTest.php
@@ -21,8 +21,9 @@
  *
  */
 
-namespace OCA\Deck\Activity;
+namespace OCA\Deck\Listeners;
 
+use OCA\Deck\Activity\ActivityManager;
 use OCA\Deck\Db\Card;
 use OCA\Deck\Db\CardMapper;
 use OCA\Deck\Db\ChangeHelper;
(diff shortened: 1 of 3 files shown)


luka-nextcloud (dismissed):
The failed cypress tests don't seem to be related to the code change.
The failed integration test is related to the code change. Seems like comment event was not triggered.

tests/unit/Activity/CommentEventHandlerTest.php
```diff
@@ -49,7 +49,7 @@ public function setUp(): void {
 		$this->notificationHelper = $this->createMock(NotificationHelper::class);
 		$this->cardMapper = $this->createMock(CardMapper::class);
 		$this->changeHelper = $this->createMock(ChangeHelper::class);
-		$this->commentEventHandler = new CommentEventHandler(
```
blizzz, line 52: needs a use statement

blizzz (changes requested):

tests/unit/Listeners/CommentEventListenerTest.php
```diff
@@ -21,8 +21,9 @@
  *
  */
 
-namespace OCA\Deck\Activity;
+namespace OCA\Deck\Listeners;
```
blizzz, line 24: :grimacing: okay i guess you can do it that way

blizzz (approved):

tests/unit/Listeners/CommentEventListenerTest.php
juliushaertl: I also addressed the review comment from Luka with it as the namespace of the actual class was changed, it made sense to adapt the text namespace as well ;)

PR #6065, opened 2024-07-02
fix: Remove bindParam usage with simpler query

juliushaertl:
For https://github.com/nextcloud/server/pull/46192

bindParam is no longer available and we can actually use a simpler query with joins to avoid sub queries and make it a bit easier to understand.


mejo- (approved):
Looks sensible. I didn't test.

PR #6114, opened 2024-07-15
Optimize getSharedWith to fetch permissions with initial query instead of one per card

juliushaertl:
- **fix: Add method to map board to file permissions**
- **perf(sharing): Optimize getSharedWith to fetch permissions right away**
- **fix: Chunk query for getting labels for cards**

* Resolves: # 
* Target version: main

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/lib/Sharing/DeckShareProvider.php
+++ b/lib/Sharing/DeckShareProvider.php
@@ -267,7 +267,13 @@ class DeckShareProvider implements \OCP\Share\IShareProvider {
 		return $share;
 	}
 
-	private function applyBoardPermission($share, $permissions, $userId) {
+	private function applyBoardPermission($share, $permissions, $userId, ?array $shareToPermissionMap = null): void {
+		$cardId = (int)$share->getSharedWith();
+		if ($shareToPermissionMap !== null && isset($shareToPermissionMap[$cardId])) {
+			$share->setPermissions($this->permissionService->boardToFilePermission($shareToPermissionMap[$cardId]));
+			return;
+		}
+
 		try {
 			$this->permissionService->checkPermission($this->cardMapper, $share->getSharedWith(), Acl::PERMISSION_EDIT, $userId);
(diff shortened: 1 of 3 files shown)


lib/Sharing/DeckShareProvider.php
```diff
@@ -267,7 +267,13 @@ private function createShareObject(array $data): IShare {
...
-	private function applyBoardPermission($share, $permissions, $userId) {
+	private function applyBoardPermission($share, $permissions, $userId, ?array $shareToPermissionMap = null): void {
+		$cardId = (int)$share->getSharedWith();
+		if ($shareToPermissionMap !== null && isset($shareToPermissionMap[$cardId])) {
+			$share->setPermissions($this->permissionService->boardToFilePermission($shareToPermissionMap[$cardId]));
+			return;
+		}
+
 		try {
 			$this->permissionService->checkPermission($this->cardMapper, $share->getSharedWith(), Acl::PERMISSION_EDIT, $userId);
```
juliushaertl, line 278: This is the previously expensive part as it needs to:
- find the board id for the card which we already have in the result of the query that fetches all shares
- fetches the permissions for each board (N+1) which can be optimized by only fetching once we have all board ids
- Then a map from card id to the related board permissions is built and passed along to the logic above to actually build the file permissions

luka-nextcloud:
@juliushaertl I think cypress/integration tests are failing. :)

juliusknorr:
Tests should pass now.

juliusknorr:
Query number increase is coming from the change of using BoardMapper:findAllForUser instead of BoardMapper::findBoardIds which does one more query then the other method.
Screenshot 2024-12-19 at 18 24 21

Might be worth to double check, however the noticable query saving does not happen in the integration tests as we mostly share one card with an attachment to another user.

Needs checking if we can still optimize the above part

luka-nextcloud:
@juliusknorr Can we get this merged soon?

Issue #6118, opened 2024-07-19
Comment json dump


grnd-alt:
the occ deck:export does currently not include comments
related to: #3828

PR #6119, opened 2024-07-19
export json data of commments

grnd-alt:
* Resolves: #6118
* Target version: main

### Summary
add comments to json export

### Checklist

### DIFF
--- a/lib/Service/Importer/Systems/DeckJsonService.php
+++ b/lib/Service/Importer/Systems/DeckJsonService.php
@@ -6,6 +6,7 @@
 
 namespace OCA\Deck\Service\Importer\Systems;
 
+use OC\Comments\Comment;
 use OCA\Deck\BadRequestException;
 use OCA\Deck\Db\Acl;
 use OCA\Deck\Db\Assignment;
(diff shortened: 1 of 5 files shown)


juliushaertl:
Code looks good, tests need some adjustment:

There was 1 error:

1) OCA\Deck\Command\UserExportTest::testExecute
TypeError: OCA\Deck\Command\UserExport::__construct(): Argument #7 ($commentService) must be of type OCA\Deck\Service\CommentService, Mock_IUserManager_f15701a4 given, called in /home/runner/actions-runner/_work/deck/deck/apps/deck/tests/unit/Command/UserExportTest.php on line 62

lib/Service/Importer/Systems/DeckJsonService.php
```diff
@@ -103,8 +104,17 @@ public function getCardAssignments(): array {
 	}
 
 	public function getComments(): array {
-		// Comments are not implemented in export
-		return [];
+		$comments = [];
+		foreach ($this->tmpCards as $sourceCard) {
+			$commentsOriginal = $sourceCard->comments;
```
juliushaertl, line 109: Do we need to handle cases where comments is not defined or null here? I could imagine this breaking otherwise if an old json without comments is used?

juliushaertl (changes requested):
One remark/question

lib/Service/Importer/Systems/DeckJsonService.php
grnd-alt: I don't think we have to. The next loop in line 113, for my understanding,  just gets skipped if $commentsOriginal is not iterable. I tested it without comments in the json and it worked without any errors.

juliushaertl (approved):

PR #6136, opened 2024-07-23
fix: fix invalid optional parameter

icewind1991:
Fixes: `Optional parameter $userId declared before required parameter $path is implicitly treated as a required parameter at /var/www/html/nextcloud/web/apps/deck/lib/Service/ConfigService.php#234` warning


juliushaertl:
Where did you get this error? The parameter is already required just nullable as type, the actual warning should already have been addressed by https://github.com/nextcloud/deck/pull/5794

icewind1991:
The log I was reading was probably from before that fix and I misread the current code.

Issue #6132, opened 2024-07-22
(Un)Archive cards programmatically via REST API


xyjonas:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Is your feature request related to a problem? Please describe.**
I am planning to create a small script that will fetch some external data and "reset" the state of given cards (e.g. un-archiving it and maybe unchecking all previous checked checkboxes).
A lot is already possible with the documented REST API, however, I could not find a way to programmatically archive/unarchive cards.

**Describe the solution you'd like**
I imagine two new urls like:
* PUT /boards/{boardId}/stacks/{stackId}/cards/{cardId}/archive
* PUT /boards/{boardId}/stacks/{stackId}/cards/{cardId}/unarchive

to perform the desired action.

PR #6138, opened 2024-07-25
Feat(REST API): add url to (un)archive cards

xyjonas:
* Resolves: #6132 
* Target version: main

### Summary
I implemented two new routes `/api/v{apiVersion}/boards/{boardId}/stacks/{stackId}/cards/{cardId}/archive` & `/api/v{apiVersion}/boards/{boardId}/stacks/{stackId}/cards/{cardId}/unarchive`, that will archive/unarchive cards via `cardService`. Documentation and unit tests are also included.
Please review my work and let me know what you think.

### Checklist


juliushaertl (changes requested):
Thanks for your pull request and sorry for taking so long to review. Generally this seems good to me. However since this is a new endpoint we need to bump the API version so that clients can detect if the new API endpoint is actually available in https://github.com/nextcloud/deck/blob/fix/nextcloud-30-design-issues/lib/Capabilities.php#L39

luka-nextcloud (approved):

luka-nextcloud:
@juliushaertl The PR looks good to me, can we merge it?

Issue #5663, opened 2024-03-13
Losing focus while editing Title field on new or existing card


estux:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Describe the bug**
If I start editing the title on new or existing card and I move the mouse cursor out of the card border, the title field looses focus. While I continue typing I can accidentally activate other keyboard commands or edit the title of another card where the mouse cursor is over.

**To Reproduce**
1. Go to a Deck board
2. Click on + in a List to add a new card.
3. Move mouse cursor over another card
4. Cannot type in title field anymore

or

1. and 2. as before
3. Start typing the title in the field
4. Move mouse cursor (by mistake, as you have physical mouse that get in the way near the keyboard)
5. You moved over another card title field and you're editing its title

or

1. and 2. as before
3. Start typing the title in the field
4. Move mouse cursor (by mistake) in an empty area of the board (you stopped editing the title field)
5. As you're still typing you could activate random keyboard commands associated to archive card, assign, create a new card and so on, creating a mess!

or

1. and 2. as before
3. Click on a title of an already existing card in the board, to edit it.
4. Start typing
5. Move your mouse cursor (by mistake) out of the card border
6. As you're still typing you risk again creating a mess activating random keyboard commands

**Expected behavior**
The title field should not loose focus if I'm still editing it. The title field should not loose focus until I save/confirm what I typed.

**Screenshots**
Sadly screenshots don't help to explain the problem this time.

**Client details:**
 - OS:  Windows 10
 - Browser:  Firefox
 - Version 122.0.1
 - Device: laptop

<details>
<summary>Server details</summary>

**Operating system**:  Linux 5.15.136-1-pve

**Web server:** nginx/1.18.0 (fpm-fcgi)

**Database:**  mysql 10.5.23

**PHP version:** 8.2.14

**Nextcloud version:** 28.0.3 - 28.0.3.2

**Where did you install Nextcloud from:** unknown (as I'm paying a provider for the service)

**Signing status:**
No errors have been found.

**List of activated apps:**

Enabled:
 - activity: 2.20.0
 - bookmarks: 13.1.3
 - bruteforcesettings: 2.8.0
 - calendar: 4.6.6
[...]

PR #6140, opened 2024-07-26
fix: losing focus while editing title field

luka-nextcloud:
* Resolves: #5663 
* Target version: main

### TODO

- [ ] ...

### Checklist


src/components/cards/CardItem.vue
```diff
@@ -11,8 +11,7 @@
 			:tabindex="0"
 			class="card"
 			@click="openCard"
-			@keyup.self="handleCardKeyboardShortcut"
-			@mouseenter="focus(card.id)">
+			@keyup.self="handleCardKeyboardShortcut">
```
juliushaertl, line 14: As discussed i think we should still keep the focus logic but instead ignore the mouseenter if the shortcutlock is set https://github.com/nextcloud/deck/blob/main/src/components/cards/CardItem.vue#L211

luka-nextcloud:
@juliushaertl cypress issue was resolved.

elzody (approved):
Works great on my side

juliushaertl:
/backport 2e552f474a3ecd4cf2af62f9615513b3df577297 to stable28

PR #6201, opened 2024-08-13
use deleted_users for users that do not exist

grnd-alt:
### Summary
When a user is deleted the activity tab of deck causes errors for all activities done by that user. To prevent that we should use a deleted_users value just as we do for deck comments already

### Checklist


mejo- (approved):
I didn't test, but code changes look sensible.

Issue #6164, opened 2024-08-02
Nextcloud 30 design issues


juliushaertl:
- [ ] List/column headings bottom of some letters are cut off
  <img width="360" alt="Screenshot 2024-08-02 at 13 04 00" src="https://github.com/user-attachments/assets/0ed20419-23d6-4b9c-b103-e8c5c254a077">
- [ ] Spacing between a list with scroll bar/no bar is different at the top
  <img width="360" alt="Screenshot 2024-08-02 at 13 04 00" src="https://github.com/user-attachments/assets/0ed20419-23d6-4b9c-b103-e8c5c254a077">
- [ ] Should the share icon be white like 3 dot menu -> Migrate to a vue-material-design-icon that picks up the text color
    <img width="121" alt="Screenshot 2024-08-02 at 13 03 52" src="https://github.com/user-attachments/assets/462ff59e-15fb-46e5-96a0-09e286228e41">
- [ ] Still have that issue with card's content being pushed to left.
  <img width="491" alt="Screenshot 2024-08-02 at 13 06 29" src="https://github.com/user-attachments/assets/d956af35-addd-4a0f-92eb-d83421ad2c06">
- [x] Board / upcoming card heading has wrong indent
<img width="339" alt="Screenshot 2024-08-02 at 13 06 52" src="https://github.com/user-attachments/assets/74220279-19a4-4d0a-beaa-522cab89fcb7">
<img width="442" alt="Screenshot 2024-08-02 at 13 06 47" src="https://github.com/user-attachments/assets/e1dfc72f-0083-4411-bbdf-17057ce7242b">

viktorix:
A few additional styling issues:

1. Some cards with content have 3 scrollbars:
![image](https://github.com/user-attachments/assets/46a65dd5-80ea-4190-bcb7-e0c7984b56bc)

2. Left corners of the card are not round:
![image](https://github.com/user-attachments/assets/48872be0-b360-46d6-a962-f3fd22b458a3)

3. Move card screen issues:
![image](https://github.com/user-attachments/assets/94ed2fa2-ed72-4660-82e3-a1d7deb63402)

4. Not sure if it's an issue, card title/input seems to be missing top margin:
![image](https://github.com/user-attachments/assets/b7eae9f9-dd4a-4a2a-9534-74a209283df6)

alexanderdd:
If you create many cards (~10) in a column, the spacing is wrong:
[...]

PR #6232, opened 2024-08-21
fix: 30 design issues

luka-nextcloud:
* Resolves: #6164
* Target version: main

### Things that were addressed

1. Spacing between a list with scroll bar/no bar is different at the top
Before
image
After
image

2. Should the share icon be white like 3 dot menu -> Migrate to a vue-material-design-icon that picks up the text color
Before
image
After
image

3.Board / upcoming card heading has wrong indent
Before
image
After
image

4. Left corners of the card are not round
Before
image
After
image

5. Move card screen issues
Before
image
After
image

6. Card has many scrollbars
Before
image
After
image

7. Card title/input seems to be missing top margin
Before
image
After
image

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -192,7 +192,7 @@ export default {
 		min-width: 100%;
 		height: 100vh;
 		max-height: calc(100vh - 120px);
-		overflow: auto;
+		overflow: hidden;
 	}
 
 </style>
--- a/src/App.vue
+++ b/src/App.vue
@@ -190,8 +190,7 @@ export default {
 	.modal__card {
 		width: 100%;
 		min-width: 100%;
-		height: 100vh;
-		max-height: calc(100vh - 120px);
+		height: calc(100% - 20px);
 		overflow: hidden;
 	}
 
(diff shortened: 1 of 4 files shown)


src/components/board/Board.vue
```diff
@@ -284,6 +284,7 @@ export default {
 					padding-top: 15px;
 					margin-top: -10px;
 					scrollbar-gutter: stable;
+					max-height: 94%;
```
juliushaertl, line 287: This seems like an odd magic number, can we caluclate that rather somehow based on what elements reduce the height from 100%?

juliushaertl (changes requested):
One comment, can you please also list the items from the original issue that were addressed in the pr description?

src/components/board/Board.vue
luka-nextcloud: I have checked again, we can fix this issue by removing `min-height` of stack header. So, we don't need that `max-height` anymore.

juliushaertl: All right, makes sense 👍

juliushaertl (approved):

juliushaertl:
Pushed a small fix for the duplicate scrollbar issues ;)

alexanderdd:
I have this indent issue on NC 29.0.5 Deck 1.13.2:
grafik

@juliusknorr  can you do a backport to 29?

PR #6257, opened 2024-08-26
copy labels when moving cards between boards

grnd-alt:
* Target version: main

### Summary
labels assigned to cards that move between boards now get cloned
old json deck files that have cross referencing labels now skip those labels instead of throwing an error

### Checklist

### DIFF
--- a/lib/Service/CardService.php
+++ b/lib/Service/CardService.php
@@ -40,6 +40,7 @@ class CardService {
 	private StackMapper $stackMapper;
 	private BoardMapper $boardMapper;
 	private LabelMapper $labelMapper;
+	private LabelService $labelService;
 	private PermissionService $permissionService;
 	private BoardService $boardService;
 	private NotificationHelper $notificationHelper;
@@ -61,6 +62,7 @@ class CardService {
 		StackMapper $stackMapper,
 		BoardMapper $boardMapper,
 		LabelMapper $labelMapper,
+		LabelService $labelService,
 		PermissionService $permissionService,
 		BoardService $boardService,
 		NotificationHelper $notificationHelper,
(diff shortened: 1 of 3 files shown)


stefan-niedermann:
I did a quick code review and I am wondering whether or not there should be permission check (I think the "manage board" permission is required when a user wants to create new labels)?

lib/Service/CardService.php
```diff
@@ -351,6 +354,29 @@ public function update($id, $title, $stackId, $type, $owner, $description = '',
 
 
 		$card = $this->cardMapper->update($card);
+
+		$stack = $this->stackMapper->find($card->getStackId());
+		$board = $this->boardService->find($stack->getBoardId());
+		$boardLabels = $board->getLabels();
+		if ($boardLabels == null) {
+			$boardLabels = [];
+		}
```
juliushaertl, line 363:
```suggestion
		$boardLabels = $board->getLabels() ?? [];
```

lib/Service/CardService.php
```diff
@@ -351,6 +354,29 @@ public function update($id, $title, $stackId, $type, $owner, $description = '',
 
 
 		$card = $this->cardMapper->update($card);
+
+		$stack = $this->stackMapper->find($card->getStackId());
+		$board = $this->boardService->find($stack->getBoardId());
```
juliushaertl, line 359:
```suggestion
		$board = $this->boardService->find($this->cardMapper->findBoardId($card->getId());
```

We can get the board id directly which is likely cached

lib/Service/CardService.php
```diff
@@ -351,6 +354,29 @@ public function update($id, $title, $stackId, $type, $owner, $description = '',
...
+		$boardLabels = $board->getLabels();
+		if ($boardLabels == null) {
+			$boardLabels = [];
+		}
+		foreach($card->getLabels() as $cardLabel) {
```
juliushaertl, line 364: I think we should only do all this if the card was actually moved to a different board, can we check that in any way?

juliushaertl (approved):

PR #6286, opened 2024-09-01
feat: Implement reference resolving for cards that have a link in the title

juliusknorr:
I started this as a small fun experiment to see if we can make use of the reference API in deck to actually have additional details if you just drop a link to a card

[Kapture 2024-09-01 at 11.59.42.webm](https://github.com/user-attachments/assets/a473cde1-a7f7-4c0b-a7c5-24165e59ee20)

- [x] Improve rendering of link preview in the modal/sidebar
- [x] Fix editing the link


juliusknorr:
Ready for review. I dropped a small performance optimization but that was causing more issues and is rather a micro-optimization (see last commit).

src/components/cards/CardItem.vue
```diff
@@ -38,7 +38,10 @@
 				</h4>
```
grnd-alt, line 38: the card.title is somehow not shown when triggerEditTitle is called. The user still edits the displayTitle.

grnd-alt (changes requested):
the menu button for inline title edit causes issues. When editing the title, the display name is in the input field. It's very easy to edit that displayname title thinking it will keep the url reference. I think we should either have a url field (so users can change the title), or change the editing title to the url in the UI when clicking edit Title

src/components/cards/CardItem.vue
juliusknorr: Should be fixed, there was some odd behaviour around contenteditable that I could only fix by using a key property on the two elements. Otherwise it seems some parts of the dom were not updated properly.

Added a cypress test as well for the renaming.

grnd-alt (approved):
🎉

PR #6327, opened 2024-09-18
fix/card title

juliushaertl:
- **fix: Fix title selection and editing**
- **fix: Inline links in card title**

* Resolves: # 
* Target version: main

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/src/components/cards/CardItem.vue
+++ b/src/components/cards/CardItem.vue
@@ -20,7 +20,7 @@
 			<CardCover v-if="showCardCover" :card-id="card.id" />
 			<div class="card-upper">
 				<h4 v-if="!editingTitle" dir="auto">
-					<span class="dragDisabled">{{ card.title }}</span>
+					<span v-auto-link class="dragDisabled">{{ card.title }}</span>
 				</h4>
 				<h4 v-else
 					dir="auto"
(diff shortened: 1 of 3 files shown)


src/components/cards/CardItem.vue
luka-nextcloud, line 1: The `inlineEditingBlocked` computed doesn't seem to be used anymore.

luka-nextcloud:
It works well. BTW, I think we should autofocus the card title when user select menu "Edit title". Currently, after selecting "Edit title", user has to click on card title to edit it.

juliushaertl:
Yes, that is the main issue I'm still struggling with :/

juliusknorr:
All resolved and now based on #6286

src/components/cards/CardItem.vue
```diff
@@ -213,8 +241,18 @@ export default {
 
 			this.$root.$emit('open-card', this.card.id)
 		},
+		triggerEditTitle() {
+			this.editingTitle = TITLE_EDITING_STATE.PENDING
+			this.$nextTick(() => {
+				this.$refs.titleContentEditable.focus()
```
grnd-alt, line 247: The title field was not focused when testing the pr.
this.$nextTick callbacks are executed after the vDom updates, but .focus() requires the dom to be updated already, setting the timeout with 0 allows the browser to update the DOM and the .focus() function to have an effect.

```suggestion
		setTimeout(() => {
				this.$refs.titleContentEditable.focus()
		), 0})
```

grnd-alt (changes requested):
Title was not focused for me when testing. Everything else is flawless

src/components/cards/CardItem.vue
juliusknorr: Thanks, fixed that and also a small issue where the shortcut lock was not triggered to early so moving the mouse could also have lead to the focus not working.

grnd-alt (approved):

Issue #2905, opened 2021-03-22
Adding a markdown (link, etc.) feature to the comments block


neerro:
Please add a possibility to use a full markdown features in comments block as link, etc.

![pic](https://user-images.githubusercontent.com/75830569/111969122-f6b62c80-8b0a-11eb-9e53-ac3bfe219934.jpg)

stefan-niedermann:
I am not entirely sure, but i think the comments section is reused from the nextcloud core packages, so fixing this would require to fix this in the nextcloud core. There is already [an open issue](https://github.com/nextcloud/server/issues/17375) which tracks this feature request.

But let's wait for @jakobroehrl or @juliushaertl to approve that this must be fixed over there before closing.

juliushaertl:
Yep, I would only do this globally so whenever this becomes available in server we should make sure this then applies to all apps that provide comments so closing as a duplicate of https://github.com/nextcloud/deck/issues/2143#issuecomment-676662256

PR #6416, opened 2024-10-16
feat: Add markdown support for comments

juliushaertl:
As this was very easy to implement. This contributes to https://github.com/nextcloud/server/issues/17375 for deck and has been requested in the past here directly https://github.com/nextcloud/deck/issues/2905

Screenshot 2024-10-16 at 15 43 03

### DIFF
--- a/src/components/card/CommentForm.vue
+++ b/src/components/card/CommentForm.vue
@@ -167,7 +167,10 @@ export default {
 
 	.comment-form__contenteditable {
 		word-break: break-word;
-		border-radius: var(--border-radius-large)
+		border-radius: var(--border-radius-large);
+		padding: var(--default-grid-baseline);
+		max-height: 200px;
+		overflow: auto;
 	}
 
 	.atwho-wrap {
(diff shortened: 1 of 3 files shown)


luka-nextcloud (approved):

ronau:
@juliusknorr 
Do I understand the related project boards correctly that this is going to be released in conjunction with Nextcloud 31?
Can this be expected for the 31.0 or will it be included in a later 31 minor update?
Will it be backwards-compatible with older versions of Nextcloud?

JitchMackson:
@ronau  I'm reading that the same as you; the Release milestone does appear to be for v31, so we can expect this then - I suspect it *will* be backwards compatible with previous versions of NC as "untested", as is usually the case, but YMMV. 

Personally this change alone is enough that I will be upgrading my fleet to 31.X as soon as we can get it through QA

juliusknorr:
Yes, we will release this as part of 31 https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule

PR #6417, opened 2024-10-16
fix: update nextcloud/ocp dependency

luka-nextcloud:
* Resolves: #6408


juliushaertl (changes requested):
I think we should only update the composer lock file. Could you double check with the steps Joas described in the mail on the engineering list?

PR #6430, opened 2024-10-25
Fixed some formatting that wasn't displaying correctly in readthedocs

thibaultmol:
I fixed some formatting that caused a section not to appear correctly on readthedocs:

image

And while at it, I made some general improvements in that area of the docs.

btw two questions:

- Is the plan to have the `Trello JSON` duplicate in the docs like this?
- why is the import/export doc not in readthedocs?

### DIFF
--- a/docs/User_documentation_en.md
+++ b/docs/User_documentation_en.md
@@ -105,17 +105,21 @@ It is possible to import from the following sources:
 
 #### Trello JSON
 
-Steps:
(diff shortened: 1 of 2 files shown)


juliusknorr (approved):
v

juliusknorr:
Thanks a lot. Sorry this slipped trough and the review took that look.

Issue #1813, opened 2020-05-06
Clone/copy a card


Nyco:
**Is your feature request related to a problem? Please describe.**

When a user wants to:

* split a card
* copy the same card in another stack or board (for optional later editing)

the only option is to manually:

* go to the card
* edit the description
* copy all
* close editing
* create a new card
* edit the description
* paste

It is very long as this needs lots of tiny steps. => A lot of time could be saved here.

**Describe the solution you'd like**
The meatballs menu ("...") of the card may add an action/icon "Clone card".
The result may be a second card immediately created right below the original.
Optionally, the newly created card could be in editing mode.

**Describe alternatives you've considered**

*Button location*
The "Clone card" action/icon could be located in the card detail on the rightmost area/column of the browser.

*Where to*
When cloning a card, options could be shown to the user:
Clone card to:
* same board:
  * same stack:
    * below original card
    * above original card
    * top of stack
    * bottom of stack
  * other stack:
    * top of stack
    * bottom of stack
* other board

But that may be be too much, smart defaults are cool, and principle of least surprise is better.

*Content*
It could be possible to let the user choose what to clone:
* tags
* assignees
* date
* attachments
* comments
* timeline

**Additional context**

Nils160988:
An even easier solution would be to enhance the "move card" dialog with a checkbox to copy.

So, change the "move card" option to "move/copy card" and then have an option to copy instead of move when choosing the target location.

jakobroehrl:
I can do this after https://github.com/nextcloud/deck/pull/1777 is merged

the-cloud:
@jakobroehrl I'm in a great need of this feature, and lack of knoweledge of how to do that. If you would like to review your interest in implementing this feature I would greatly appreciate that.

Well, if anyone is interested in coding this, that would be awesome. I could send some euro or crypto as thank you.

osm-frasch:
I also think that it would make work a lot easier and faster if you could simply copy cards.

TalusUnheil:
Hey there! I am also very interested in this function. I have created a card as a template, which provides a checklist, as well as a Markdown formatted skeleton for the history per card.
If I could just clone instead of copy'n'paste, that would be very helpful.

pete-at-future-campus-ruhr:
Hey guys, first off, thanks for all the great work! Is there a way to contribute to the feature? I am not a programmer, but I could offer to create syntax in form of a wireframe.

[...]

PR #6452, opened 2024-10-29
feat(cards): add card cloning ability

grnd-alt:
* Resolves: #1813 
* Target version: main

### Checklist


lib/Service/CardService.php
```diff
@@ -390,6 +391,29 @@ public function update($id, $title, $stackId, $type, $owner, $description = '',
 		return $card;
 	}
 
+	public function cloneCard(int $id, ?int $targetStackId = null):Card {
+		$this->permissionService->checkPermission($this->cardMapper, $id, Acl::PERMISSION_READ);
+		$originCard = $this->cardMapper->find($id);
+		if ($targetStackId == null) {
```
juliusknorr, line 397:
```suggestion
		if ($targetStackId === null) {
```

Always do strict comparison to be safe

lib/Service/CardService.php
```diff
@@ -77,6 +77,7 @@ public function __construct(
 		LoggerInterface $logger,
 		IRequest $request,
 		CardServiceValidator $cardServiceValidator,
+		private AssignmentService $assignmentService,
```
juliusknorr, line 80: If we touch the constructor we can move all other properties to property promition as well

lib/Service/CardService.php
```diff
@@ -390,6 +391,29 @@ public function update($id, $title, $stackId, $type, $owner, $description = '',
...
+		$originCard = $this->cardMapper->find($id);
+		if ($targetStackId == null) {
+			$targetStackId = $originCard->getStackId();
+		}
+		$this->permissionService->checkPermission($this->stackMapper, $targetStackId, Acl::PERMISSION_EDIT);
+		$newCard = $this->create($originCard->getTitle(), $targetStackId, $originCard->getType(), $originCard->getOrder(), $originCard->getOwner());
+		$boardId = $this->stackMapper->findBoardId($targetStackId);
+		foreach ($this->labelMapper->findAssignedLabelsForCard($id) as $label) {
+			if ($boardId != $this->stackMapper->findBoardId($originCard->getStackId())) {
+				$label = $this->labelService->cloneLabelIfNotExists($label->getId(), $boardId);
```
juliusknorr, line 405: A user is only able to add/edit labels on a board with manage permission iirc so we should also take that into account when cloning the labels

lib/Service/CardService.php
```diff
@@ -390,6 +391,29 @@ public function update($id, $title, $stackId, $type, $owner, $description = '',
...
+			}
+			$this->assignLabel($newCard->getId(), $label->getId());
+		}
+		foreach ($this->assignedUsersMapper->findAll($id) as $assignement) {
+			$this->assignmentService->assignUser($newCard->getId(), $assignement->getParticipant());
```
juliusknorr, line 410: I'm unsure if we should only assign users that have access to the target board. Might be confusing otherwise if you see people assigned but they don't have access to the card.

lib/Service/CardService.php
juliusknorr: Ah it is checked in https://github.com/nextcloud/deck/pull/6452/files#diff-28de0536daa5474c78d675a3bd3c4ab7a9d26f4de82031e5b16bf89dbdc30f3cR96, then let's catch the exception here and skip the label cloning I'd say

luka-nextcloud:
@grnd-alt Missing unit test for function `CardService->cloneCard()`
Missing unit test for function `LabelService->cloneLabelIfNotExists()`
I think cypress tests should be added as well.

luka-nextcloud (changes requested):
Please review tests. Otherwise it looks good!

hectorandac:
This is awesome! I'm eagerly waiting for this feature 🎉

juliusknorr:
@grnd-alt I pushed two commits to get CI passing again, the crashing electron process is a mystery still to me but seems giving it a bit of time before typing (likely to animate the modal showing) solves that.

Generally I've tested and works fine, will have another look at the code.

src/CardCloneDialog.vue
```diff
@@ -0,0 +1,120 @@
+<!--
+  - SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+<template>
+	<NcModal v-if="modalShow" :title="t('deck', 'Clone card')" @close="modalShow = false">
+		<div class="modal__content">
+			<h3>{{ t('deck', 'Clone card to another board') }}</h3>
```
juliusknorr, line 8: Seems confusing as it can also be cloning to the same board.

src/CardCloneDialog.vue
```diff
@@ -0,0 +1,120 @@
+<!--
+  - SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+<template>
+	<NcModal v-if="modalShow" :title="t('deck', 'Clone card')" @close="modalShow = false">
```
juliusknorr, line 6: Would it actually make sense to to the following?

- Call the action in the card menu "Copy/clone card"
- Merge both modals into one and just have two buttons "Copy" and "Move"

appinfo/routes.php
```diff
@@ -43,6 +43,7 @@
 		['name' => 'card#read', 'url' => '/cards/{cardId}', 'verb' => 'GET'],
 		['name' => 'card#create', 'url' => '/cards', 'verb' => 'POST'],
 		['name' => 'card#update', 'url' => '/cards/{cardId}', 'verb' => 'PUT'],
+		['name' => 'card#clone', 'url' => '/cards/{cardId}/clone', 'verb' => 'POST'],
```
juliusknorr, line 46: Deck currently has this API duplication where we tend to have two endpoints, one for web another one for external API consumers like mobile apps.

Ideally we would introduce new endpoints as OCS API endpoints, that way they can be used by both.

Not a blocker for this merge, but we should follow up that part before the final release.

src/CardCloneDialog.vue
```diff
@@ -0,0 +1,120 @@
...
+				:placeholder="stacksFromBoard.length === 0 ? t('deck', 'No lists available') : t('deck', 'Select a list')"
+				:input-label="t('deck', 'Select a list')"
+				:options="stacksFromBoard"
+				:max-height="100"
+				label="title" />
+
+			<button :disabled="!isBoardAndStackChoosen" class="primary" @click="cloneCard">
+				{{ t('deck', 'Clone card') }}
+			</button>
+			<button @click="modalShow = false">
```
juliusknorr, line 27: Should use `NcButton`

juliusknorr (commented):
Some comments inline, generally looks good to me

src/CardCloneDialog.vue
juliusknorr: Also throws a console warning:

```
[Vue warn]: [NcModal] You need either set the name or set a `labelId` for accessibility.

found in

---> <NcModal>
       <CardCloneDialog> at src/CardCloneDialog.vue
         <NcContent>
           <App> at src/App.vue
             <Root>
```

juliusknorr (approved):

juliusknorr:
Did one small commit as i thought we could have a quick refactoring to rather use NcDialog instead of NcModal

juliusknorr:
Pending documentation as we should add the new endpoint to `docs/API.md`

PR #6490, opened 2024-11-10
[main] Update nextcloud/ocp dependency

nextcloud-command:
Auto-generated update of [nextcloud/ocp](https://github.com/nextcloud-deps/ocp/) dependency

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -491,12 +491,12 @@
             "source": {
                 "type": "git",
                 "url": "https://github.com/nextcloud-deps/ocp.git",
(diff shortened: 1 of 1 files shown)


luka-nextcloud:
The error comes from a dependency `justinrainbow/json-schema`. It seems like a syntax update on PHP8.4. I guess this dependency will not be updated until PHP8.4 official release. So, I think we can ignore this issue for now.

grnd-alt:
@luka-nextcloud the library has adapted to 8.4 syntax already. Update pr for 3rdparty is here: https://github.com/nextcloud/3rdparty/pull/1986

Issue #3201, opened 2021-07-17
Add "count" for each Kanban list


ldmpub:
Hi,

I'm used to work with Jira and I misses a little info in Deck on lists: Can we display the number of cards contained in a list?

**Suggestion**
This info could be added in the list Title
See below screenshot => "how Jira does it" 😉
![image](https://user-images.githubusercontent.com/1452146/126028077-4ed72e6c-dfe5-4697-a189-4dd17a90ca16.png)

andre5tz:
I think the idea is very good.
It would be great if the idea could be implemented.

ant0nwax:
+1

PR #6511, opened 2024-11-16
feat(cards): add card counters

miqmago:
* Resolves: https://github.com/nextcloud/deck/issues/3201
* Target version: main

### Summary
Add card counters to each stack

### TODO

- [ ] Add card counters to sidebar: would require to change server response to make counting on server side
- [ ] Add card counters to all boards view (same as before

### Checklist

### DIFF
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -22,7 +22,8 @@
 		<div v-else-if="board" class="board-title">
 			<div :style="{backgroundColor: '#' + board.color}" class="board-bullet" />
 			<h2 dir="auto">
-				{{ board.title }}
+				<span class="board__title-text">{{ board.title }}</span>
+				<span class="board__title-counter">{{ cardsCount }}</span>
 			</h2>
 			<p v-if="showArchived">
 				({{ t('deck', 'Archived cards') }})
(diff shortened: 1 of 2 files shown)


src/components/Controls.vue
```diff
@@ -358,6 +365,13 @@ export default {
 				this.setPageTitle(current.title)
 			}
 		},
+		stacksByBoard: {
```
luka-nextcloud, line 368: I think this watch is not necessary.

src/components/Controls.vue
```diff
@@ -22,7 +22,8 @@
 		<div v-else-if="board" class="board-title">
 			<div :style="{backgroundColor: '#' + board.color}" class="board-bullet" />
 			<h2 dir="auto">
-				{{ board.title }}
+				<span class="board__title-text">{{ board.title }}</span>
+				<span class="board__title-counter">{{ cardsCount }}</span>
```
luka-nextcloud, line 26: I think we should add aria-label/title for this number.

luka-nextcloud (changes requested):

src/components/Controls.vue
luka-nextcloud: Could you please explain why this watcher is necessary even though the `cardsCount` is a computed variable?

luka-nextcloud: @miqmago

luka-nextcloud:
@miqmago Seems like PR was rebased in a wrong way...

ant0nwax:
+1

PR #6566, opened 2024-12-01
[main] Update nextcloud/ocp dependency

nextcloud-command:
Auto-generated update of [nextcloud/ocp](https://github.com/nextcloud-deps/ocp/) dependency


juliusknorr:
Replaced by https://github.com/nextcloud/deck/pull/6610

PR #6611, opened 2024-12-17
fix: Use new query builder

juliusknorr:
Avoid reusing a query builder instance for another query which causes some log spam and should not be done anyways


elzody (approved):
Nice 👍🏻

Issue #6263, opened 2024-08-28
delete card in modal view: modal should close


alexanderdd:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**To Reproduce**
1. open a card (modal view)
2. in top right ...-menu, click delete
3. modal is still open but just blank

**Expected behavior**
Modal should close

NC 30.0.0 RC2
c.nc.com
(can't check app version)

alexanderdd:
@nextcloud/office can you please triage this? I just reproduced it on c.nc.com

max-nextcloud:
Here's a brief video of the reproduction.
https://github.com/user-attachments/assets/0df6a62f-b610-4222-bc4c-11dbdef223a4


Seems like an easy thing to add. So marking as good first issue.

ludij:
This was a great first issue for me :)  Please see my PR above or here: https://github.com/nextcloud/deck/pull/6617

juliusknorr:
Thanks a lot for your contribution @ludij 

Looking forward to any further. In case you are interested, we have a community chat to stay in contact with active contributors and have a more direct channel for questions and discussions. I can invite you, in case you're interested just drop me a mail to the address in my github profile.

alexanderdd:
@juliusknorr is this fix already in production on cloud.nextcloud.com? (30.0.5 RC1 / cant check deck version)

cause there it is not fixed yet

max-nextcloud:
https://apps.nextcloud.com/apps/deck/releases?platform=30#30 says the last release happened in November. So the fix has not been released yet.

juliusknorr:
@luka-nextcloud @grnd-alt Can you align on building one? Maybe we should have a fixed schedule for the releases (we can discuss this in the team call next week)

alexanderdd:
can confirm it is fixed now on cloud.nextcloud.com - Thank you!!

PR #6617, opened 2024-12-19
fix: close modal after deleting card

ludij:
* Resolves: # 6263 -> https://github.com/nextcloud/deck/issues/6263
* Target version: main

### Summary
When a card is opened in the modal or sidebar and then deleted, the router now goes to the board, so that the modal closes. The sidebar was already closing, but incorrectly kept the URL for the card.

### Checklist


grnd-alt (approved):
🎉 Congratulations on your first contribution

Issue #6457, opened 2024-10-29
New cards are created at the bottom of the screen


sparunakian:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Describe the bug**
Since Nextcloud 30 and Deck, the create card button in a board creates a text input at the bottom of the screen.

**To Reproduce**
1. Open the Deck app
2. Go to a board
3. Click the create card button ![Image](https://github.com/user-attachments/assets/bbd3c21d-c9b7-49e7-b6d0-380e2e1a6d14)
4. The text input for the new card title is displayed at the bottom of the screen

**Expected behavior**
Previously, the text input was displayed next to the last card in the column.

**Screenshots**
![Image](https://github.com/user-attachments/assets/83c7490f-9deb-405d-947a-b4049e8d5c0e)

**Client details:**
 - OS: ArchLinux & Debian
 - Browser: Firefox
 - Version: 132 & 128.3.1 ESR
 - Device: Desktop

<details>
<summary>Server details</summary>

**Operating system**: Debian & ArchLinux

**Web server:** Apache

**Database:** MySQL

**PHP version:** 8.2

**Nextcloud version:** 30.0.1

**Where did you install Nextcloud from:** Zip archive (Debian) & distribution package (ArchLinux)

**Signing status:**

```
Login as admin user into your Nextcloud and access
http://example.com/index.php/settings/integrity/failed
paste the results here.
```

**List of activated apps:**

```
If you have access to your command line run e.g.:
sudo -u www-data php occ app:list
from within your Nextcloud installation folder
```

**Nextcloud configuration:**

```
If you have access to your command line run e.g.:
sudo -u www-data php occ config:list system
from within your Nextcloud installation folder

or

Insert your config.php content here
Make sure to remove all sensitive content such as passwords. (e.g. database password, passwordsalt, secret, smtp password, …)
```

**Are you using an external user-backend, if yes which one:** LDAP/ActiveDirectory/Webdav/...

</details>

<details>
<summary>Logs</summary>

#### Nextcloud log (data/nextcloud.log)
```
Insert your Nextcloud log here
```

#### Browser log
```
Insert your browser log here, this could for example include:

a) The javascript console log
b) The network log
[...]

Issue #3331, opened 2021-09-21
Dynamic column width


timkrief:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Is your feature request related to a problem? Please describe.**
I keep having to change the window's size or zoom scale, or I keep having to scroll horizontally, to be able to fit or view all stacks on the screen, sometimes I even forget about one stack that is overflowing. Plus, board with a small number of stacks feel like they could use more of the available space.

**Describe the solution you'd like**
The idea would be to let the columns stretch to fill the screen.
We could define a minimum and maximum width for a stack, close to the width value we use right now, and then let the stacks stretch to fill the page as long as they are not bigger or smaller than those limits.

This has some unexpected design bonuses.
- If there are not enough stacks to fill the screen, at least they will have the max value, reducing the amount of unused space.
- If there are not enough space to fit the stacks, at least they will have the min value, increasing the amount of information displayed.
- Sometimes, when the left or right sidebar appears, most, or even all, stacks stay visible, they just shrink a bit instead of being covered. (see the gif)

**Describe alternatives you've considered**
Being able to change column width for each board manually, but this seems a bit clunky and unreliable.

**Additional context**
I made that work using some custom CSS and it really feels nicer to use and more professional. The values of min and max could be discussed, but using values close to what we've been using for the default width ensures that this will not massively change the look and feel of the app.

![Peek 2021-09-21 05-04](https://user-images.githubusercontent.com/14089949/134105551-f1431b6b-3034-4105-8ff6-cce1a2d39e0d.gif)

This is the CSS I used. It could be cleaner if this was done natively instead of as a custom style on top of the existing one.
```CSS
.smooth-dnd-container.horizontal>*,
.dashboard-column{
    flex: 1 !important;
    width: 0 !important;
    min-width: 210px !important;
    max-width: 300px !important;
}
```

juliushaertl:
Seems like a nice improvement to me, would you be up to send a pull request for that?

dodedodo:
Off-Topic: @timkrief, what did you do to get such a pleasant-looking deck? Would you be willing to share your css? It looks way better than the default (no offence, I know deck is limited by nextclouds theming conventions).

[...]

PR #6640, opened 2025-01-01
Feature/3331 dynamic column width

ludij:
* Resolves: #3331 + #6457
* Target version: main

### Summary

Following the descriptions in the linked ticket, I have updated the styles for the board, overview, stack and search results cards, so that they behave more responsive according to the space available. As a side-effect it also fixed issue #6457.

On top of that, this PR contains the following changes, which aren't necessarily part of the issue, but I thought they'd fit in:
- aligned design for board and overview
- aligned spacing inside stack between header, cards and new card form
- refactored how border radius is achieved on card image
- added v-for loop in overview to get rid of code duplication
- fixed border around add list
- used `existingIndex` to define `existingCard` to avoid extra lookup
- fixed binding of title attibute on stack column title for boards
- improved accessibility attributes on column for overview
- changed title phrasing of search results depending on view

<table>
  <tr>
    <th>new</th>
    <th>old</th>
   </tr> 
   <tr>
      <td>1a overview new</td>
      <td>1b overview old</td>
  </tr>
   <tr>
      <td>2a board new</td>
      <td>2b board old</td>
  </tr>
</table>

### TODO
N/A

### Checklist

### DIFF
--- a/css/deck.css
+++ b/css/deck.css
@@ -14,6 +14,7 @@
 
 input[type=submit].icon-confirm {
 	border-color: var(--color-border-maxcontrast) !important;
+	border-style: solid;
 	border-left: none;
 }
 
--- a/src/components/board/Board.vue
+++ b/src/components/board/Board.vue
@@ -310,11 +310,12 @@ export default {
 				position: relative;
 
 				.smooth-dnd-container.vertical {
+					$margin-x: calc($stack-gap * -1);
 					display: flex;
 					flex-direction: column;
 					gap: $stack-gap;
-					padding: 5px calc(#{$stack-gap / 2}) $stack-gap;
-					margin: 0 calc(#{$stack-gap / -2});
+					padding: $stack-gap;
+					margin: 0 $margin-x;
 					overflow-y: auto;
 					scrollbar-gutter: stable;
 				}
--- a/src/components/board/Board.vue
+++ b/src/components/board/Board.vue
@@ -282,7 +282,7 @@ export default {
 
 	.board {
 		position: relative;
-		overflow: auto;
+		overflow-x: auto;
 		flex-grow: 1;
 		scrollbar-gutter: stable;
 	}
@@ -296,6 +296,7 @@ export default {
 		align-items: stretch;
 		gap: $board-gap;
 		padding: 0 $board-gap;
+		height: 100%;
 
 		&:deep(.stack-draggable-wrapper.smooth-dnd-draggable-wrapper) {
 			display: flex;
@@ -312,7 +313,9 @@ export default {
 					display: flex;
 					flex-direction: column;
 					gap: $stack-gap;
-					padding: 5px 0 $stack-gap;
+					padding: 5px calc(#{$stack-gap / 2}) $stack-gap;
+					margin: 0 calc(#{$stack-gap / -2});
+					overflow-y: auto;
 					scrollbar-gutter: stable;
 				}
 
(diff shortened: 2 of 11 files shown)


juliusknorr:
Thanks a lot for your contribution. This looks quite good. I'd love to collect some feedback from designers like @marcoambrosini on that. 

The code changes look sane to me, I'll give this a test run later. Can you maybe address the stylelint and eslint errors that CI reported?

juliusknorr:
Is the change of the scroll container intentional? I'm personally not a fan of scrolling the full board area and would rather prefer to stick with the per column scrolling we used to have before.

juliusknorr:
Just noticed the merge commit, could you do a rebase to the main branch instead so we have a clean git history of your changes?

ludij:
Thanks for your review! I've just added a commit to fix the linting issue. Regarding the scrolling, my change is intentional, but of course I can add an new commit so that scrolling is inside the column again. I should have time during the weekend to do that.

juliusknorr:
> Regarding the scrolling, my change is intentional, but of course I can add an new commit so that scrolling is inside the column again. I should have time during the weekend to do that. I'll do a rebase too!

That would be great, thanks a lot already.  :)

marcoambrosini:
Hi @ludij, I just had a quick look at our competitors and it seems that varying the with of the columns is not a very common practice. These columns are usually fixed in width and horizontal scrolling is used for overflowing lists. @juliusknorr what do you think about that?
That said, in principle this seems like a sane change to me design-wise, I would just make sure to the min and max values not too far apart from each other to not completely change appearance because that could be disorienting.
I need some time to setup my development environment again to properly test it. I'll come back for a review tomorrow.
Thanks a lot for for picking this up :)

ludij:
> Hi @ludij, I just had a quick look at our competitors and it seems that varying the with of the columns is not a very common practice. These columns are usually fixed in width and horizontal scrolling is used for overflowing lists. @juliusknorr what do you think about that? 

It's pretty easy to update this code so that the column width is fixed. Imho my refactor probably makes sense either way, because it allows the widths and spacings for the different parts to be maintained from a single source (`variables.scss` (at least that's what I aimed for :) )), whereas before it was needed to adjust multiple files and lines.

> That said, in principle this seems like a sane change to me design-wise, I would just make sure to the min and max values not too far apart from each other to not completely change appearance because that could be disorienting. I need some time to setup my development environment again to properly test it. I'll come back for a review tomorrow. 

Sounds good. I'm not so certain whether my changes to the spacing inside the card look good (with/without labels, with/without image, etc.). I also just noticed that the spacing inside the "Add new card" card doesn't align perfectly with the header and other cards in the column. I could easily fix that for the "Add new card", but it'd be great to get some feedback on the updated spacing inside the cards in general before, so I can do an update in one go. 

> Thanks a lot for for picking this up :)

Very happy to contribute and thanks for the feedback!

grnd-alt (commented):
thanks for that contribution, I like the way the width is limited right now and would be happy getting this merged. Let's wait for a final approve from @marcoambrosini and get this in. 🎉

marcoambrosini (approved):
Sorry for the late reply, I tested with the current parameters and everything looks good even if cards have lots of elements. Thanks a lot again for the contribution!

grnd-alt (approved):

ludij:
@juliusknorr + @grnd-alt Thanks for the feedback and sorry for the delay. I'm not very familiar with the rebase workflow, so it took me a few pushes, but I think everything should be fine now 🚀

luka-nextcloud (approved):

marcoambrosini:
@luka-nextcloud what's preventing this from being merged?

PR #6642, opened 2025-01-02
fix: Avoid failing on public dav URLs

juliusknorr:
It seems that the dav plugins are now also called on public share links, so this changes ensure that we do not throw errors as deck is not expecting its classes to be initialized without a user session.


lib/Service/AssignmentService.php
```diff
@@ -88,8 +88,6 @@ public function __construct(
 		$this->changeHelper = $changeHelper;
 		$this->activityManager = $activityManager;
 		$this->eventDispatcher = $eventDispatcher;
-
-		$this->assignmentServiceValidator->check(compact('userId'));
```
juliusknorr, line 92: This is fine as it is checked in the controller methods

luka-nextcloud (approved):

PR #6671, opened 2025-01-20
fix: Properly show attachment extension

juliusknorr:
Properly show attachment extensions as a separate dom element to avoid mixing up filenames and extensions.


luka-nextcloud:
@juliusknorr Looks like failed cypress test is related to your code change.

elzody (approved):
Good change, also gave it a quick look locally

Issue #6446, opened 2024-10-26
Deleted cards break "occ deck:export"


fracklaus:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Describe the bug**

The occ deck:export functionality has a bug and does not work when a card has been deleted within deck. The following error appears:
 [OCA\Deck\NoPermissionException]
 Card is deleted

It seems that deck:export is checking for read access in `/var/www/nextcloud/apps/deck/lib/Service/CommentService.php:46` by calling `$this->permissionService->checkPermission(...)`. When executed on a deleted card, checkPermission throws an error "Card is deleted" that is not being catched.
The right reaction in deck:export should be to ignore the deleted card and continue with the next, instead of exiting due to the thrown exception.

**To Reproduce**
1. Create a board in nextcloud deck as user USER
2. Enter some cards
3. Delete at least one card
4. Execute `occ deck:export USER`
5. Export will fail with "Card is deleted" exception

**Expected behavior**

Export continues, deleted card is not part of export.

**Screenshots**

**Client details:**

Error happens in command line on the server, client does not seem to be relevant.

 - Device: [e.g. iPhone6, desktop]

<details>
<summary>Server details</summary>

**Operating system**: Debian 12

**Web server:** apache

**Database:** postgresql

**PHP version:** 8.2.24

**Nextcloud version:** (see Nextcloud admin page) 30.0.1.2

**Where did you install Nextcloud from:** 

**Signing status:**

```
Login as admin user into your Nextcloud and access
http://example.com/index.php/settings/integrity/failed
paste the results here.
```
No errors have been found.

**List of activated apps:**

```
If you have access to your command line run e.g.:
sudo -u www-data php occ app:list
from within your Nextcloud installation folder
```

**Nextcloud configuration:**

```
If you have access to your command line run e.g.:
sudo -u www-data php occ config:list system
from within your Nextcloud installation folder

or

Insert your config.php content here
Make sure to remove all sensitive content such as passwords. (e.g. database password, passwordsalt, secret, smtp password, …)
```

**Are you using an external user-backend, if yes which one:** LDAP/ActiveDirectory/Webdav/...

QctiveDirectory

</details>

<details>
[...]

PR #6692, opened 2025-01-24
fix: skip exporting a deleted card

elzody:
* Resolves: https://github.com/nextcloud/deck/issues/6446
* Target version: main

### Summary
Checks if the card is deleted before continuing -- if it is, just move on

### Checklist

### DIFF
--- a/lib/Command/UserExport.php
+++ b/lib/Command/UserExport.php
@@ -62,6 +62,7 @@ class UserExport extends Command {
 			if ($board->getDeletedAt() > 0) {
 				continue;
 			}
+
 			$fullBoard = $this->boardMapper->find($board->getId(), true, true);
 			$data[$board->getId()] = $fullBoard->jsonSerialize();
 			$stacks = $this->stackMapper->findAll($board->getId());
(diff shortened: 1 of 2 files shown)


elzody:
@luka-nextcloud @grnd-alt I never touched any of the integration tests, so I'm not sure why the query count would have increased. The only tests I added were unit tests, which I would assume would also increase the query count since the command makes database queries. Are unit tests included in the query diff? The message is a bit confusing.

elzody:
@grnd-alt  I cannot figure out why it says the number of database queries has increased. If anything, the changes should make it query the database *less*. And the unit tests, from what I can see, should not be affecting it. Am I missing something?

Maybe @luka-nextcloud can have a look?

luka-nextcloud (approved):

luka-nextcloud:
> @grnd-alt I cannot figure out why it says the number of database queries has increased. If anything, the changes should make it query the database _less_. And the unit tests, from what I can see, should not be affecting it. Am I missing something?
> 
> Maybe @luka-nextcloud can have a look?

@elzody It is the query count total, not slow query. I guess it might be related to the update of server repository. It's just a few queries of increasing. We just need to update the `tests/integration/base-query-count.txt` and nothing to worry about.

PR #6705, opened 2025-01-29
fix: set cypress ci server version to stable31

grnd-alt:


.github/workflows/cypress-e2e.yml
```diff
@@ -47,7 +47,7 @@ jobs:
 
       - name: Register text Git reference
         run: |
-          text_app_ref="$(if [ "${{ matrix.server-versions }}" = "master" ]; then echo -n "main"; else echo -n "${{ matrix.server-versions }}"; fi)"
+          text_app_ref="$(if [ "${{ matrix.server-versions }}" = "stable31" ]; then echo -n "main"; else echo -n "${{ matrix.server-versions }}"; fi)"
```
juliusknorr, line 50: We should keep this part, stable31 is used then already (see the else branch)

juliusknorr (approved):

PR #6740, opened 2025-02-11
feat: update default content

luka-nextcloud:
* Target version: main

### Summary
Add new default content for deck.

### Screenshots
image
image
image
image
image
image
image

### Checklist


grnd-alt:
could you add some screenshots, to check it out?

lib/Service/DefaultBoardService.php
```diff
@@ -71,19 +79,62 @@ public function checkFirstRun($userId): bool {
...
+			}
+		}
+
+		$cardsDataJson = file_get_contents(__DIR__ . '/fixtures/default-cards.json');
+		$cardsData = json_decode($cardsDataJson, true);
+
+		foreach ($cardsData as $cardData) {
+			$stackIndex = $cardData['stackIndex'];
+			$defaultCards[] = $this->cardService->create(
+				$this->l10n->t($cardData['title']),
```
juliusknorr, line 115: This won't work. The translations are extracted and then put for translation based on regex matches for the t() method calls, so the strings in the json file would not be found by that.

We should rather have the card title strings defined in PHP directly using the t method calls.

juliusknorr:
Can you also check the cypress tests? Maybe we can get them back to green as well :)

grnd-alt (approved):
default board looks great 🔥
cypress seems unrelated to me

PR #6778, opened 2025-02-20
fix: cypress issue

luka-nextcloud:
* Resolves: # 
* Target version: main

### Summary
Fixes cypress issue in the workflow.

### DIFF
--- a/cypress/e2e/cardFeatures.js
+++ b/cypress/e2e/cardFeatures.js
@@ -87,8 +87,8 @@ describe('Card', function () {
 		cy.get('.modal-mask.card-selector button.button-vue--vue-primary').should('be.visible').click()
 		cy.wait('@save', { timeout: 7000 })
 
-		cy.visit(`/apps/deck/#/board/${boardId}`)
 		cy.reload()
+		cy.visit(`/apps/deck/#/board/${boardId}`)
 		cy.get('.board .stack').eq(0).within(() => {
 			cy.get(`.card:contains("${newCardTitle}")`).should('be.visible')
 		})


cypress/e2e/cardFeatures.js
```diff
@@ -88,7 +88,6 @@ describe('Card', function () {
 		cy.wait('@save', { timeout: 7000 })
 
 		cy.visit(`/apps/deck/#/board/${boardId}`)
-		cy.reload()
```
blizzz, line 91: Was this check not intended to double check that the card indeed exists even after a reload? Otherwise it could (theoretically) be that there is only a presentation, but the backend actually did not store anything.

luka-nextcloud: I think you are right... let me try to find another solution.

blizzz (approved):
It would be good if the action from line 80 would also change the URL so that a reload indeed works. Not sure whether this might be a cypress limitation? That would be favorable, also from UX wise. Otherwise OK.

luka-nextcloud:
> It would be good if the action from line 80 would also change the URL so that a reload indeed works. Not sure whether this might be a cypress limitation? That would be favorable, also from UX wise. Otherwise OK.

I don't understand what you mean. Line 78 to 88 is actions on modal. Why should they change the URL?

blizzz:
> > It would be good if the action from line 80 would also change the URL so that a reload indeed works. Not sure whether this might be a cypress limitation? That would be favorable, also from UX wise. Otherwise OK.
> 
> I don't understand what you mean. Line 78 to 88 is actions on modal. Why should they change the URL?

Screenshot_20250220_183356

is what I mean, i hope it is more clear?

luka-nextcloud:
@blizzz So, this is a suggestion for a behavior update of the create card modal. If we implement this, after a card is created, the user will stay on the selected board page. I'm not sure if this would be a better UX. I think some users might prefer to stay on the overview page to create cards for multiple boards and check the overview more easily.

PR #6854, opened 2025-03-25
 fix: make bin icon red on button hover 

SethFalco:
* Resolves: N/A
* Target version: main

### Summary

The bin icon would only highlight in red when specifically hovered over the icon, when it should've probably been when hovered over the button.

This just fixes that!

#### Befor

https://github.com/user-attachments/assets/994bfabd-e7ec-4867-ab2e-1208da8e4cd6

#### After

https://github.com/user-attachments/assets/53c0f779-956e-4633-8224-ec70b28f082d

### Checklist

- [x] Tests (unit, integration, api and/or acceptance) are included — N/A
- [x] Documentation (manuals or wiki) has been updated or is not required - N/A

### Chores

Also does a chore!

The image `ghcr.io/juliushaertl/nextcloud-dev-php81:latest` is a 4xx error now. 

A maintainer of the project changed their GitHub username, which broke the command in the README for running the nextcloud-dev Docker container. This needs to have their up-to-date username in the image path.

While updating that, I also changed a few other references as well.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,7 +10,7 @@ updates:
     timezone: Europe/Paris
   open-pull-requests-limit: 10
   reviewers:
-  - juliushaertl
+  - juliusknorr
   - luka-nextcloud
 
 - package-ecosystem: npm
@@ -90,7 +90,7 @@ updates:
     timezone: Europe/Paris
   open-pull-requests-limit: 10
   reviewers:
-  - juliushaertl
+  - juliusknorr
   - luka-nextcloud
 
 - package-ecosystem: composer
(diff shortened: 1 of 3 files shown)


elzody:
Thanks a lot for handling the updates to the readme and such :)
As for the icon maybe @nextcloud/designers could have a look and see which makes the most sense? I think for the most part our icons in the action menus like this are just white/black.

jancborchardt:
In Talk we always color-code the action "Delete conversation" red (which is also always last in the list, as all delete actions should be).
This would be nice to do elsewhere too since it not only makes it quick to find and easier to understand, but also very immediately reduces the list you have to scan by one if what you are looking for is not deletion.
image

I would say we should even put this in the components and/or design guidelines @marcoambrosini @nimishavijay?

PR #6872, opened 2025-04-01
feat: add board import and export

luka-nextcloud:
* Resolves: # 
* Target version: main

### Summary
It makes it possible to export and import a deck board as a JSON/CSV file. CSV is only for export while JSON works for both import and export.

### Video recording
[demo.webm](https://github.com/user-attachments/assets/d51bcbb2-50b2-4fce-a36d-da8991254898)

### Checklist


elzody:
It works quite well, the only thing I noticed was that when I upload a clearly incorrect `.json` file to see what it does, it throws an error (as expected) but with the actual log message that you would see in the `nextcloud.log` file:

```
"OCA\\Deck\\Service\\Importer\\BoardImportService::setData(): Argument #1 ($data) must be of type stdClass, string given, called in /var/www/master/apps-extra/deck/lib/Service/Importer/BoardImportService.php on line 123"
```

Maybe instead it should display something more readable, and also something that doesn't expose the internal path of some of the files (e.g. the `/var/www/...`)

elzody (approved):
Seems good to me now.

Issue #6901, opened 2025-04-11
Clear Selected Stack When Changing Boards in "Move/Copy Card" Dialog


luka-nextcloud:
**Describe the bug**
In the "Move/Copy Card" dialog, the selected stack should be cleared upon changing the selected board. This enhancement prevents confusion when multiple boards have stacks with identical names, which could lead to incorrect stack selection and card was not moved correctly

**Screenshots**

![Image](https://github.com/user-attachments/assets/7f4821ff-59ab-41d4-b8b4-44dabeca0584)

PR #6902, opened 2025-04-11
Clear selected stack when selected board changed

luka-nextcloud:
* Resolves: #6901
* Target version: main

### Summary
In the "Move/Copy Card" dialog, the selected stack should be cleared upon changing the selected board. This enhancement prevents confusion when multiple boards have stacks with identical names, which could lead to incorrect stack selection and card was not moved correctly.

### Checklist


elzody (commented):
Seems sensible, but the Cypress test failure does seem related. The "Copy card" button is disabled now so the `click()` doesn't go through

elzody (approved):

PR #7030, opened 2025-05-30
fix: unstable cypress test

luka-nextcloud:
### Summary
Fix unstable cypress issue.


cypress/e2e/cardFeatures.js
```diff
@@ -301,12 +302,13 @@ describe('Card', function () {
 				.first().click()
 			cy.get(`.card:contains("${newCardTitle}")`).should('be.visible').click()
 
-			cy.get('#app-sidebar-vue [data-test="tag-selector"]').should('be.visible').click()
+			cy.get('#app-sidebar-vue [data-test="tag-selector"]').should('be.visible')
+				.wait(1000).click()
```
blizzz, line 306: `wait()` does not make it more reliable, it only varies the flakyness. What are the issues here when it fails? In tables we have good results with `should('be.enabled')` iirc.

elzody: Alternatively you can change the timeout for this specific operation only:

```js
cy.get().should('be.visible').click({ timeout: 1234 })
```

`cy.wait()` is an anti pattern IMO

blizzz (approved):

PR #7104, opened 2025-07-10
fix: unstable cypress

luka-nextcloud:
* Resolves: # 
* Target version: main

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/cypress/e2e/cardFeatures.js
+++ b/cypress/e2e/cardFeatures.js
@@ -302,6 +302,9 @@ describe('Card', function () {
 				.first().click()
 			cy.get(`.card:contains("${newCardTitle}")`).should('be.visible').click()
 
+			// Add delay to ensure the events are bound
+			cy.wait(1000)
+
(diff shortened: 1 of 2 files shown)


elzody:
Usually not a fan of using `cy.wait()`, what is the reason for it in this case?

luka-nextcloud:
> Usually not a fan of using `cy.wait()`, what is the reason for it in this case?

@elzody I understand but I suspect the click event hadn't been bound yet at some point. That's why I try adding a short delay before performing the click action.

elzody:
> ... I suspect the click event hadn't been bound yet at some point.

Do you mean that the element was not visible yet and that the click event could not go through? Just a bit confused on what you mean by "hadn't been bound yet." I guess it's fine either way, just want to see if there is another way since this is considered an anti-pattern: https://docs.cypress.io/api/commands/wait#Time

@luka-nextcloud

luka-nextcloud:
@elzody Based on the Cypress report, we can see that the element is visible. However, when a click is performed, nothing happens. So, I suspect the click event listener on that element might not be ready yet.

elzody (approved):

Issue #6739, opened 2025-02-10
labelId must be provided and must be numeric


jgarte:
Hi,

When creating a new card from the dashboard I get the following error:

`labelId must be provided and must be numeric`.

Has anyone else encountered this?

![Image](https://github.com/user-attachments/assets/3764a8bc-ce8d-4664-963b-df6aa881532d)

grnd-alt:
Thanks for your report. I was able to reproduce somewhat.
Could you please share the exact order of things you inserted the data into the form? When trying to reproduce I noticed that when you insert the title and set a tag after, everything works. But when you set the tag and then insert the title I get "cardId must be provided and must be numeric" I was not able to reproduce to get "labelId must be provided and must be numeric".
I think they are of similar origin but it would be helpful if you could help me reproduce the exact issue, by describing your steps.

rsi-art:
Hi, I'm facng de same issue (Nextcloud 31 Nextcloud hub 10) AIO  deployement on docker

Steps to reproduce : 

From Deck, next card, click to add a new card

- Filling juste the name of the card => OK
- Filling name + descirption => OK
- Filling name + user => KO error message
- 

![Image](https://github.com/user-attachments/assets/177aba34-8949-4f5f-a140-2ecae31d1875)

- Filling name + user + due date => KO error message
- 

![Image](https://github.com/user-attachments/assets/32c8f2d9-9b29-46c4-bc25-37e518254c23)

- Filling name + user + due date + description => KO error message

slu-010101:
We can reproduce this if we add the card from "Upcoming cards" + **and** assign a user to this card.
Nextcloud 31.0.0 with Deck 1.15.0

MarcSchuh:
I am having the same issue. 
For a 
`php occ maintenance:repair`
I get a
`     - ERROR: An exception occurred while executing a query: SQLSTATE[22007]: Invalid datetime format: 1366 Incorrect string value: '\xF0\x9D\x97\x9F\xF0\x9D...' for column `nextcloud`.`oc_deck_cards`.`description` at row 760`

also at a nextcloud 31.0.4 and Deck 1.15.1

MarcSchuh:
Firstly, thanks for the awesome work :) 
Having the same issue with Nextcloud 31.0.4 and Deck 1.15.1

Aveyron-RetD:
yes duplicate of #6818

luka-nextcloud:
[...]

PR #7139, opened 2025-07-28
fix:  Use getId() method for card ID retrieval

enjeck:
Resolves: #6739 

### Checklist


grnd-alt (approved):

luka-nextcloud:
@enjeck Code change is ok.
My feedback: I think it's worth adding a Cypress test for creating a card from the dashboard.

grnd-alt:
I tested a bit, and for me creating a label in the dialog still results in the "labelId must be present..." error as reported in the issue. Just the same error for cardId is gone. I think we should look into that as well.

Issue #7147, opened 2025-07-30
Optimizing events


stpronk:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Is your feature request related to a problem? Please describe.**
I am making a app based on the deck app, and I trying to use the events as they are available within the code. The issue is that they all use the same event which isn't a problem by itself, but they don't give through all the required data. As an example, the biggest that I want to use are the CardUpdatedEvents but when I reorder a card, only one card is passed through to the event which is not even the updated card in the first place.

**Describe the solution you'd like**
The events should be altered to represented the correct actions by using an action parameter and/or giving every action it's own event class. Besides that, the full and correct data should be passed through to the event.

**Describe alternatives you've considered**
I tried building on top of the events with if statements, but the before data is missing by a lot of events or pass through incorrect data which halted every solution.

stpronk:
@grnd-alt As mentioned in https://github.com/nextcloud/deck/pull/7124#pullrequestreview-3070702549 , you would like some more context about the app i'm speaking of.

The app is still private as we just started working on it, but I could describe the the functionality of the app. We want to hook into the events to setup more email notifications with a specific format. The infrastructure where we are running NextCloud is completely closed from the internet and the only way out is through email at this moment in time (might be an secure api in the future). When a card changes stack we want need to inform our client that something has happened with their card which leaves us no choice then to created an email while copy pasting a lot of information from NextCloud. 

These actions are all done by hand, which requires a lot of time & energy and is prone to errors. We are trying to create a solutions for this to automate these action.

_(As we are trying to create app that is a real edge-case app within NextCloud, we don't expect the solution within the deck-app but would be nice to have some little improvements to make it more durable & expandable.)_

vdiezel:
Having a very related request!

We want to use the Events-API (because notify-push-client and Activity-API were not suitable here for several reasons)  to run some code after something has happened (e.g. call another API) for a deeper integration into our platform. 

I've just setup a simple NextCloud App that will relay those events (https://git.wechange.de/gl/code/wechangenotifier) into our platform and I run into the same limitation of the `CardUpdatedEvent`, the missing `cardBefore`  on most of them. 

[...]

PR #7149, opened 2025-07-31
[Enhancement] Added events for Stack actions which can be hooked into

stpronk:
* Resolves part of: #7147
* Target version: main

### Summary

Added multiple events for Stacks with the app. We can easily hook into the events when the stacks are updated.

Left the event dispatch for the board within the actions to keep it backwards compatible. Might be a argument to remove them now or later.

### TODO

- [x] Add events for the different type of actions in "StackService"
- [x] Dispatch the correct events within the correct actions in "StackService"

### Checklist

### DIFF
--- a/lib/Event/AStackEvent.php
+++ b/lib/Event/AStackEvent.php
@@ -1,0 +1,27 @@
+<?php
+
+/**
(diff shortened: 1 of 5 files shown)


stpronk:
Didn't saw any tests for the events being fired, shouldn't they exist?

Issue #6960, opened 2025-04-29
Rest API: Reorder card not working / missing type casting for Ids


vdiezel:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Describe the bug**

I built a custom frontend by using the provided REST API (thank you very much for providing one!) and encountered an issue in the endpoint for reordering cards via: 

`PUT /boards/{boardId}/stacks/{stackId}/cards/{cardId}/reorder`

This endpoint does not seem to work. The relative card order does not change.  
The native VueJS-client does call the same CardService.reorder function, but via the CardController and it works, while the call via the CardApiController that is used by the REST API fails. 

I think I have found the issue: When calling CardService.reorder via the CardApiController, the cardId and the stackId are passed as strings (order seems unaffected and is always a number). This probably breaks the comparison and lookups by id somewhere. I didn't dig deep enough to check why other endpoint seem unaffected by this.

I did setup a quick dev environment and could confirm that casting them to ints in the CardApiController.reorder-function's call to CardService.reorder (just like the CardController does) fixed the REST endpoint.

I am willing to contribute a PR to fix it, if I get an ok :). While at it, it might be worth to check other places of the REST API for type mismatches that might cause trouble. Even if they don't, it might be worth to fix them to prevent future issues.

I also think this issue is related: 
https://github.com/nextcloud/deck/issues/4369

**To Reproduce**
1. Create a stack with a bunch of cards
2. Try to change their order via the REST API endpoint (*not* /deck/cards/$cardId/reorder which is used by the native client)
3. The relative order of the cards will not change

**Expected behavior**
The relative order of cards changes correctly

**Nextcloud version:** 

Confirmed on 26, 30 and 31

**Signing status:**

```
Login as admin user into your Nextcloud and access
http://example.com/index.php/settings/integrity/failed
paste the results here.
```

Issue #6830, opened 2025-03-12
Reorder API has Changed?


gnilebein:
I have a few scripts that automatically move Nextcloud Deck cards to another stack when overdue. These no longer work with the current version.

I use this endpoint as described in the API:

/boards/{boardId}/stacks/{stackId}/cards/{cardId}/reorder

In my opinion, this was also used for the web interface. 

However, the web interface now uses this endpoint:

/cards/{cardId}/reorder as the endpoint.

Has the API changed here and the documentation has not been adapted?

luka-nextcloud:
@gnilebein As far as I know, the Rest API endpoints did not change. Please share system info and error messages.

user8446:
Hi you mentioned it changed with the "current version". Did you update to NC 31 which upgrades deck to 1.15.0? 

I use that API as well and want to be on the lookout for it.

nikhub:
I can confirm that the behavior of the API has changed after updating Deck.
Currently I am running Nextcloud 31.0.6 with Deck version 1.15.1.

I am also using the REST API to move cards. Before updating Deck (or NC respectively) moving Cards worked fine.
Now I still get a "200 Success" response, but the card stays at its position.

@gnilebein 
Can you please explain, what you changed in your code? I think I roughly understand what you did, but am not that familiar with PHP.
Mainly you are now passing indivual objects of each card to the CurlRequest instead of the common data object and you changed the order from 0 to 999. Am I missing something?

nikhub:
I just updated to Deck version 1.15.2. The issue is still active.
Does anybody have an idea for a workaround? Currently I am unable to use the API properly as this is a major feature of Deck.

user8446:
I am on deck v 1.15.2 as well and reordering is still working with no changes on our end.

However, we are just creating a new card in the stack and then reordering it to the top in the same stack (order=0). We are not moving it to a different stack.

Maybe that is where the issue is?

nikhub:
> I am on deck v 1.15.2 as well and reordering is still working with no changes on our end.
> 
> However, we are just creating a new card in the stack and then reordering it to the top in the same stack (order=0). We are not moving it to a different stack.
> 
> Maybe that is where the issue is?

Sorry, I was a little imprecise.
Yes, I want to move cards from one stack to another. I did not try to move cards within a stack, yet.

vdiezel:
[...]

PR #7154, opened 2025-08-04
fix: parse arguments to CardService.reorder correctly to int

vdiezel:
* Resolves: #6960 #6830 
* Target version: main

### Summary

This correctly casts cardId and stackId (and order, just for consistency) into `integers` in CardService.reorder, no matter where CardService.reorder is called from.  This fixes the call from CardApiController.reorder, because it passes cardId and stackId as strings.  I've seen other places where there is no type cast applied, but it seems it is not an issue most of the time when no identity operator `===` or `!==` is used. Reorder does use it, so that is why it broke. 

The type casting currently is a little bit inconsistent:  There is everything from relying on php's implicit type casts from type defs in function parameters, to explicit type casting or none at all, but I was too afraid to touch it, as it might break things. So I kept my change minimal. 

Tested this from the native client by modifying the URL in CardApi.reorder: 

```js
	reorderCard(card) {
		return axios.put(this.url(`/api/v1.0/boards/$YOUR_BOARD_ID_HARDCODED/stacks/${card.stackId}/cards/${card.id}/reorder`), card)
			.then(
				(response) => {
					return Promise.resolve(response.data)
				},
				(err) => {
					return Promise.reject(err)
				},
			)
			.catch((err) => {
				return Promise.reject(err)
			})
	}
```

### Checklist

### DIFF
--- a/lib/Controller/CardApiController.php
+++ b/lib/Controller/CardApiController.php
@@ -185,7 +185,7 @@ class CardApiController extends ApiController {
 	 * Reorder cards
 	 */
 	public function reorder($stackId, $order) {
-		$card = $this->cardService->reorder($this->request->getParam('cardId'), $stackId, $order);
+		$card = $this->cardService->reorder((int)$this->request->getParam('cardId'), (int)$stackId, (int)$order);
 		return new DataResponse($card, HTTP::STATUS_OK);
 	}
 }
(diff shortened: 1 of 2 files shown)


vdiezel:
> thanks a lot for your first contribution 🔥 I think the type definitions in cardService should definitely be there. The type coercion though is only done in non strict mode (which is not set, but still) so I think we should keep the explicit type conversion we have in CardController right now, as it also doesn't have any downside afaik

Thank you!
I see. Then we also should not rely on it in the CardApiController.reorder. I added explicit conversion there, too.

grnd-alt (approved):

PR #7156, opened 2025-08-05
fix!: Consistent card/cardBefore on CardUpdatedEvent

vdiezel:
* Resolves: related to #7147 
* Target version: main

### Summary

As mentioned in #7147, the CardUpdatedEvent has two problems: 

1. It often doesn't have the cardBefore even if it could
2. The card it contains is often not the updated version, but still the old one

This makes it difficult for consumers of the event to act upon it.

The CardUpdatedEvent now always has post- and pre-update version of the card. I had to introduce an additional database call in some methods, but they are not ones that are called very frequently, so I think it is fine. 

I did check the internal consumers of the CardUpdatedEvent and they should be unaffected. FullTextSearchEventListener relies only on the card ID, which doesn't change, and LiveUpdateListener is only concerned when the boardId of a card changes, which can only happen in the CardService.update which I didn't touch. 

I still flagged this as a breaking change because anything outside of the deck app that relies on the inconsistent event might break.

#### Regarding tests:

I'm not very familiar with phpunit. I was able to add a basic assertion that the CardUpdatedEvent is dispatched and that cardBefore is not `null`. However I was not able to assert that the cards are actually the cards before and after the update. This involves some interventions at the setters of the Card class that are beyond my skills. If there is a way, you can nudge me towards it and I will happily add the assertions.

### Checklist


luka-nextcloud (approved):

vdiezel:
@grnd-alt 

Is there anything I can do to get this PR merged? Happy to fix any issues. 
We are maintaining a fork of the deck app because we rely on these changes and it would be a big relief to "return home".

grnd-alt:
@vdiezel tbh I think we can merge this after a rebase to get main back in!

Issue #6528, opened 2024-11-22
Loosing focus when editing the description in a card


arnowelzel:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Describe the bug**
When I open a card to add a description, the card will be saved automatically as soon as I stop typing. However in this moment the focus of the text editor gets lost and instead the title is focused. This is very confusing since you do not expect that the cursor suddenly jumps from the text editor to the title.

**To Reproduce**
1. Create a new card.
2. Open the card.
3. Click the "Description" field and type something.
4. When the card is saved, it can happen, that the focus is lost and jumps to the title.

**Expected behavior**
The focus should always stay where it is and must not be changed without user interaction to change it intentionally.

**Screenshots**
![Image](https://github.com/user-attachments/assets/153a51a9-64ab-4c5f-9584-97a50361fbb8)

**Client details:**
 - OS: Windows 11
 - Browser: Vivaldi, Chrome, Firefox, Edge
 - Version: Latestt versions as of 2024-11-22
 - Device: Desktop

<details>
<summary>Server details</summary>

**Operating system**:

**Web server:**

**Database:**

**PHP version:**

**Nextcloud version:** (see Nextcloud admin page)

**Where did you install Nextcloud from:**

**Signing status:**

```
No errors have been found.
```

**List of activated apps:**

```
Enabled:
  - activity: 3.0.0
  - admin_audit: 1.20.0
  - app_api: 4.0.0
  - bruteforcesettings: 3.0.0
  - calendar: 5.0.5
  - circles: 30.0.0
  - cloud_federation_api: 1.13.0
  - collectives: 2.15.1
  - comments: 1.20.1
  - contacts: 6.1.1
  - contactsinteraction: 1.11.0
  - dav: 1.31.1
  - deck: 1.14.2
  - dicomviewer: 2.2.1
  - drawio: 3.0.3
  - external: 5.5.2
  - federatedfilesharing: 1.20.0
  - federation: 1.20.0
  - files: 2.2.0
  - files_downloadlimit: 3.0.0
  - files_pdfviewer: 3.0.0
  - files_reminders: 1.3.0
  - files_sharing: 1.22.0
[...]

PR #7162, opened 2025-08-13
fix: do not change focus when card id stays the same

grnd-alt:
* Resolves: #6528 
* Target version: main

### Summary
the notify_push sends updates when anything on the board changes triggering the watch hook as well.
We should not focus the header when the card id has not actually changed.

### Checklist


luka-nextcloud (approved):

luka-nextcloud:
@grnd-alt backport needed?

PR #7164, opened 2025-08-13
perf(stack): Optimize fetching all cards from a stack

CarlSchwan:
- Don't fetch stack, if we already know it
- fetch all comments at the same time, instead of for each card

Requires https://github.com/nextcloud/server/pull/54409

### Checklist


Olen:
Not sure if this is the right PR to comment on, but the it would be nice if the "enricher" also added the card `labels` to each card.
Right now, I need a separate query for each card just to get the labels when getting the cards for a stack..

CarlSchwan:
https://github.com/nextcloud/server/pull/54409 is now merged, I'll rebase this soon

PR #7165, opened 2025-08-13
perf(cards): fetch all cards at once

CarlSchwan:
### Summary

Fetch all cards at once instead of one by one

### Checklist

### DIFF
--- a/lib/Db/StackMapper.php
+++ b/lib/Db/StackMapper.php
@@ -77,12 +77,10 @@ class StackMapper extends DeckMapper implements IPermissionMapper {
 
 	/**
 	 * @param numeric $boardId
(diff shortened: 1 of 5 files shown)


CarlSchwan:
friendly ping :)

lib/Db/StackMapper.php
```diff
@@ -77,12 +77,10 @@ public function findStackFromCardId($cardId): ?Stack {
 
 	/**
 	 * @param numeric $boardId
-	 * @param int|null $limit
-	 * @param int|null $offset
```
luka-nextcloud, line 81: Could you please explain why those lines should be removed?

PR #7254, opened 2025-09-22
Fix autosave focus jump in card sidebar

hweihwang:


elzody:
Is there a relevant issue for this, or could you quickly give some reproduction steps for this? The code looks alright but I think some more context would help me fully understand what is happening

luka-nextcloud:
@hweihwang Could you add description to the PR?

PR #7259, opened 2025-09-23
fix(l10n): Use numbered placeholders

rakekniven:
Reported in Transifex.


rakekniven:
If my change is correct, please give me a hint as I found similar lines which need to get changed.

rakekniven:
Can someone please review?

Issue #7263, opened 2025-09-24
Draging User Avatar over card indicates possible drop


grnd-alt:
# Steps to reproduce
create 2 deck cards
assign a user to one of them
drag the user icon over the other card

the dragged over card shows that a file drop is possible but there is no actual Item in the dragOver event

# expected behavior
dragOver events without data items should not be considered for the view change of the deck card.

PR #7264, opened 2025-09-24
fix: do not show dragged over state when no data in drag event

grnd-alt:
* Resolves: #7263 
* Target version: main


luka-nextcloud:
@grnd-alt The code change looks ok. But when I test it, I still sometimes able to drag the avatar icon into a card and make it being uploaded.

luka-nextcloud:
https://github.com/user-attachments/assets/817f451b-0915-42a6-a793-a7240847fc06

luka-nextcloud:
@grnd-alt Could you check again?

PR #7283, opened 2025-09-29
ci(cypress): Use cypress-split for running tests in parallel

luka-nextcloud:
* Resolves: cypress memory issue in the workflow
* Target version: main

### Summary
- added `experimentalMemoryManagement: true` into cypress.config.js
- It looks like the memory issue is within cypress itself. So updated the workflow file to make it run parallel on multiple runners (same as the Text app)

image

### DIFF
--- a/tests/integration/composer.json
+++ b/tests/integration/composer.json
@@ -1,7 +1,7 @@
 {
     "require-dev": {
         "phpunit/phpunit": "~9",
-        "behat/behat": "~3.23.0",
+        "behat/behat": "~3.26.0",
         "guzzlehttp/guzzle": "7.9.2",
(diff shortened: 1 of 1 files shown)


grnd-alt (approved):
greatly needed to improve CI 🔥 
we will have to adapt branch protection rules for the changed check name

juliusknorr:
I think we run into the same issue as in text here where we have not found any feasible solution other then sticking to a rather old cypress version

https://github.com/cypress-io/cypress/issues/27415

Since the number of cypress tests is rather small we could think about moving to playwright instead. This happened on some other repos and so far my impression is that it is more stable to use.

luka-nextcloud:
Ok, I will look into playwright integration soon.

PR #7287, opened 2025-10-04
Update README.md

holger-dev:
Replace the outdated (non-existing) iOS app reference with the newly released and fully working app Next Deck. The new app looks great and works reliably.


holger-dev:
Hey! If you want to you can add the Repo now - maybe it helps for getting the last review
[Next Deck Repo](https://github.com/holger-dev/nextdeck)

juliusknorr:
Thanks, that looks very interesting :) Need to give it a try as well.

ornanovitch:
This app is not free to use... our users now have to update to a paid app whereas the previous one was free.

holger-dev:
But the Android-Version is also not for free...

ornanovitch:
It is on F-Droid :)

It's your choice, I don't want to deny it, but users used to have the choice to pay or not

PR #7341, opened 2025-11-04
fix: Check owner can create boards when importing

luka-nextcloud:
* Resolves: # 
* Target version: main

### Summary
Adds permission checks to board import functionality to ensure that only users with the appropriate rights can create boards. The main changes involve enforcing these checks in both the UI controller and API controller, and extending unit tests to cover permission scenarios.

**Permission enforcement improvements:**

* Added a check in `BoardController::import` to throw `NoPermissionException` if the user is not allowed to create boards.
* Added a similar permission check in `BoardImportApiController::import`, preventing board creation via API for unauthorized users.

**Dependency and test updates:**

* Injected `PermissionService` into `BoardImportApiController` and updated its constructor accordingly.
* Updated the unit test `BoardImportApiControllerTest` to mock `PermissionService`, test successful import when permitted, and add a new test for denied permission, ensuring `NoPermissionException` is thrown.

### Checklist


juliusknorr:
Just to double check, do we already hide the import functionality in the ui then?

luka-nextcloud:
> Just to double check, do we already hide the import functionality in the ui then?

@juliusknorr Yes, the `Import board` menu is already hidden if the user doesn't have create permission.

Issue #7340, opened 2025-11-02
Can't add or modify comment when mentioning user


inthemill:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Bug Description

**Describe the bug**
When i want to add a comment to a card where i am mentioning an other nextcloud user, pressing the "arrow" button will do nothing

**To Reproduce**
1. Go to the 'Details' of a Card
2. Go to the tab 'comments' i guess. In German it's "Kommentare"
3. Type somthing into the field to add a new comment
4. Type @ to select an other user to mention
5. Select an other user from the drodown
6. Press the "arrow" button to add the comment
7. What i see: while pressing the button a small white field apears above the mousepointer. the comment does not gets added. Javascript error in console: "TypeError: Cannot read properties of undefined (reading 'data-mention-id')"
9. What i would have expected: The comment should have been added

<img width="1491" height="498" alt="Image" src="https://github.com/user-attachments/assets/c3346ee3-4e1b-4fcf-aa10-74460c7456c5" />

**Client details:**
 - OS: MacOS ARM
 - Browser: Firefox (144.0.2 (aarch64)) & Chrome (41.0.7390.108)
 - Device: Laptop

<details>
<summary>Server details</summary>

**Apps**
relevant app "Deck" Version 1.15

| Name | Version | Level | Aktionen
 | -- | -- | -- | --
  | Activity | 4.0.0 | Vorgestellt | Deaktivieren
  | Brute-force settings | 4.0.0 | Vorgestellt | Deaktivieren
  | Calendar | 5.5.7 | Vorgestellt | Deaktivieren
  | Client Push | 1.2.0 |   | Deaktivieren
  | Collaborative tags | 1.21.1 | Vorgestellt | Deaktivieren
  | Comments | 1.21.0 | Vorgestellt | Deaktivieren
  | Contacts | 7.3.5 | Vorgestellt | Deaktivieren
  | Contacts Interaction | 1.12.0 | Vorgestellt | Deaktivieren
  | Dashboard | 7.11.0 | Vorgestellt | Deaktivieren
  | Deck | 1.15.3 | Vorgestellt | Deaktivieren
[...]

PR #7354, opened 2025-11-09
fix: Can't add or modify comment when mentioning user because parentNode returns null

ArtemSkit:
* Resolves: #7340 
* Target version: main

### Summary
When requesting parentNode of "div.comment-form__contenteditable" we get null. Request only the first parent.

### Checklist


ArtemSkit:
I installed the v1.16.1 update, and I can mention users in the comments. The issue is fixed, no need for the merge.

PR #7364, opened 2025-11-11
ci: do not force phpunit version

grnd-alt:


juliusknorr (approved):

juliusknorr:
Seems good but does not pass yet ;)

Issue #7375, opened 2025-11-19
Multiple embed board widgets only show same board


luka-nextcloud:
**Describe the bug**
Embed multiple boards with smartpicker can only show 1 board.

**To Reproduce**
1. Open markdown file or whiteboard file
2. Add a board with smartpicker
3. Add another board with smartpicker

**Expected behavior**
Different board widget shows different board.

**Screenshots**
All widgets shows the same board.

PR #7383, opened 2025-11-20
refactor: convert store to factory functions

luka-nextcloud:
* Resolves: #7375
* Target version: main

### Summary
Refactors the Vuex store setup to use factory functions for creating new store instances. This change allows Board reference widgets to maintain separate store instances, resolving the issue where loading multiple board widgets would always load the same board.

**Store architecture refactor:**

* Changed `store/main.js` to export a `storeFactory` function that returns a new Vuex store instance, instead of exporting a singleton store. All imports and usages of the store are updated to use `storeFactory()`.

* Refactored `store/stack.js` and `store/card.js` to export factory functions (`stackModuleFactory`, `cardModuleFactory`) that return new module objects, replacing singleton module exports.

**Widget and component store usage:**

* Updated components and widgets (e.g., `BoardReferenceWidget.vue`, widget registration in `init-reference.js`) to use new store instances by calling `storeFactory()`

### Checklist

### DIFF
--- a/src/sessions.js
+++ b/src/sessions.js
@@ -5,7 +5,6 @@
 
 import { listen } from '@nextcloud/notify_push'
 import { sessionApi } from './services/SessionApi.js'
-import storeFactory from './store/main.js'
 import axios from '@nextcloud/axios'
 
 const SESSION_INTERVAL = 90 // in seconds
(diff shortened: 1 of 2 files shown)


src/sessions.js
```diff
@@ -14,6 +14,8 @@ let hasPush = false
 
 let syncRunning = false
 
+const store = storeFactory()
```
juliusknorr: Could it be problematic here that we are not using the same instance as in the main app?

Issue #7369, opened 2025-11-14
[Bug] comment input ignored key input after user mention


luka-nextcloud:
**Describe the bug**
Cannot type any string after adding user mention. Issue only occurs on Firefox.

**To Reproduce**
1. Open a card then navigate to comment tab
2. Click on comment input to type something and mention a user
3. After selecting a user mention, click on the free space of input box after user mention
4. Try to input something

**Expected behavior**
continue typing without any problem

PR #7397, opened 2025-11-25
fix(comment): improve comment form behavior in firefox

luka-nextcloud:
* Resolves: #7369
* Target version: main

### Summary
Fix for Firefox where clicking on/after a mention in card comment input

### Checklist

### DIFF
--- a/src/components/card/CommentForm.vue
+++ b/src/components/card/CommentForm.vue
@@ -5,40 +5,11 @@
 
 <template>
 	<div class="comment-form">
-		<form @submit.prevent="submit">
-			<At ref="at"
-				v-model="commentText"
-				:members="members"
-				name-key="displayname"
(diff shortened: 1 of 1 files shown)


luka-nextcloud:
@juliusknorr Sounds reasonable :) I will update the PR.

css/deck.css
```diff
@@ -22,3 +22,9 @@ input[type=text]:focus+input[type=submit].icon-confirm,
 input[type=text]:hover+input[type=submit].icon-confirm {
 	border-color: var(--color-main-text) !important;
 }
+
+[class^="_tribute-container-autocomplete_"],
```
juliusknorr: Could this go to the CommentForm.vue file? We can add style block that is not scoped that then applies globally for this.

Issue #7226, opened 2025-09-12
[Bug]: [Deck] Unable to remove shared user or leave board (Nextcloud 30.0.12~, 31.x)


007Poe:
### ⚠️ This issue respects the following points: ⚠️

- [x] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [x] This issue is **not** already reported on [Github](https://github.com/nextcloud/server/issues?q=is%3Aopen+is%3Aissue+label%3Abug) OR [Nextcloud Community Forum](https://help.nextcloud.com/) _(I've searched it)_.
- [x] Nextcloud Server **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Bug description

A shared user cannot be removed from a Deck boad by the owner.
Additionally,the shared user cannot leave the board by themselves.

### Steps to reproduce

1.Create a new board 
2.Share the board with anothe Nextcloud user.
3.Try to remove that user from the board.
4.Log in as the shared user and try to leave the board/

### Expected behavior

1.Owner should be able to remove sharee users.
2.Shared user should be able to leave the board.

### Nextcloud Server version

30

### Operating system

Debian/Ubuntu

### PHP engine version

PHP 8.3

### Web server

Apache (supported)

### Database engine version

MySQL

### Is this bug present after an update or on a fresh install?

None

### Are you using the Nextcloud Server Encryption module?

Encryption is Disabled

### What user-backends are you using?

- [x] Default user-backend _(database)_
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Configuration report

```json

```

### List of activated Apps

```shell

```

### Nextcloud Signing status

```shell

```

### Nextcloud Logs

```json

```

### Additional info

Tested on multiple browsers(Chrome , Vivaldi, Firefox), same behavior.
Issue occurs in both NC 30.0.12~ and NC 31.
Deck Version:　1.15.2

Gobytego:
i did notice you can remove a group so i have been assigning users their own groups which is nuts but it works.

osm-frasch:
**Same issue here.** 

Testet with Browsers Firefox, Chrome (latest) . Deck: latest, NC 31.0.8 Enterprise. 
[...]

PR #7412, opened 2025-11-28
feat(sharing): ability to leave a shared board

luka-nextcloud:
* Resolves: #7226
* Target version: main

### Summary
The shared users can leave the board by themselves.

### Checklist

### DIFF
--- a/lib/Service/BoardService.php
+++ b/lib/Service/BoardService.php
@@ -432,6 +432,37 @@ class BoardService {
 		return $deletedAcl;
 	}
 
+	public function leave(int $boardId): ?Acl {
+		if ($this->permissionService->userIsBoardOwner($boardId)) {
+			throw new BadRequestException('Board owner cannot leave board');
+		}
+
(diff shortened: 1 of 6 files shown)


lib/Service/BoardService.php
```diff
@@ -432,6 +432,38 @@ public function deleteAcl(int $id): ?Acl {
...
+			throw new BadRequestException('Board owner cannot leave board');
+		}
+
+		$acl = $this->aclMapper->findParticipantFromBoard($boardId, Acl::PERMISSION_TYPE_USER, $this->userId);
+
+		if (!$acl) {
+			throw new BadRequestException('Not a participant of this board');
+		}
+
+		$this->boardMapper->mapAcl($acl);
```
grnd-alt: I don't know if we need this function call, don't we already have everything we need even without this call?

grnd-alt (changes requested):
overall looks and works nice, some small thoughts though.

Issue #6185, opened 2024-08-07
Windows compatibiliy: unspecified error when uploading a forbidden file


tobiasKaminsky:
- open deck card
- upload a file from device, e.g. CON
- see 
![image](https://github.com/user-attachments/assets/c30a4cc3-3f21-4028-af6b-88b40fb40c06)

XueSheng-GIT:
Just got the same error on NC 31.0.7 with Deck 1.15.1.

Error is:
```
[deck] Fehler: "/user/files/Deck" is locked, existing lock on file: none
	POST /apps/deck/cards/198/attachment
```

<details>

<summary>Show full log entry</summary>

```
[...]

PR #7429, opened 2025-12-04
Fix/6185: Windows compatibiliy: unspecified error when uploading a forbidden file

samin-z:
* Resolves: # [[6185](https://github.com/nextcloud/deck/issues/6185)]
* Target version: main

### Summary
Fix for showing proper user-friendly error message when uploading a forbidden file name.

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/lib/Service/FilesAppService.php
+++ b/lib/Service/FilesAppService.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 /**
  * SPDX-FileCopyrightText: 2018 Nextcloud GmbH and Nextcloud contributors
  * SPDX-License-Identifier: AGPL-3.0-or-later
@@ -7,6 +9,7 @@
 
 namespace OCA\Deck\Service;
 
+use OCA\Deck\BadRequestException;
 use OCA\Deck\Db\Acl;
 use OCA\Deck\Db\Attachment;
 use OCA\Deck\Db\CardMapper;
@@ -16,7 +19,9 @@ use OCA\Deck\StatusException;
 use OCP\AppFramework\Http\StreamResponse;
 use OCP\Constants;
 use OCP\Files\Folder;
+use OCP\Files\IFilenameValidator;
 use OCP\Files\IMimeTypeDetector;
+use OCP\Files\InvalidPathException;
 use OCP\Files\IRootFolder;
 use OCP\Files\NotFoundException;
 use OCP\IDBConnection;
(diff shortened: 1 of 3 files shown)


lib/Service/FilesAppService.php
```diff
@@ -29,6 +33,12 @@
 use Psr\Log\LoggerInterface;
 
 class FilesAppService implements IAttachmentService, ICustomAttachmentService {
+	/**
+	 * With intensive testing, the only case of sanitization is "/" to ":" before the code runs, it's mapped now for better UX.
```
juliusknorr: Can you tell how you triggered this? I would be surprised if we really need such a mapping, but maybe I miss some case here

lib/Service/FilesAppService.php
```diff
@@ -307,4 +321,18 @@ private function getShareForAttachment(Attachment $attachment): IShare {
...
+
+	private function validateFilename(string $fileName): void {
+		try {
+			$this->filenameValidator->validateFilename($fileName);
+		} catch (\Exception $e) {
+			$errorMessage = $e->getMessage();
+			if (preg_match('/\\\\?["\']([^"\']+)\\\\?["\'] is not allowed/', $errorMessage, $matches)) {
+				$invalidCharacter = stripslashes($matches[1]);
+				$invalidCharacter = self::SANITIZED_CHAR_MAPPING[$invalidCharacter] ?? $invalidCharacter;
+				throw new BadRequestException($this->l10n->t('Invalid filename. The character "%s" is not allowed in filenames.', [$invalidCharacter]));
```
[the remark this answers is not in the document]
samin-z: i added the regex for the edge case of ":". it was over engineering, i will updated it with simple str_replace.

lib/Service/FilesAppService.php
samin-z: i uploaded a file "test/file.txt" and in the error it returned "test:file.txt", so i found it as a edge case.

juliusknorr: Ah did you test on windows then? I was just thinking about macOS/Linux but there the : makes sense as a replacement then indeed.

juliusknorr: Let's keep as you proposed, can you add a comment to the code with a hint that this is specific to mac os with above reasoning? That way it is easier to understand if checking the code later on. Approving already as the other changes look good.

juliusknorr:
The integration test failure does not seem related, let me rebase to see

samin-z:
@juliusknorr, thank you. i ran unittest and integration test locally and both passed with no issue.

Issue #7427, opened 2025-12-04
Unassigned option malfunction


ahangarha:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Describe the bug**
When filtering cards based on assignment, the "Unassigned" option doesn't work properly. It shows that it is checked but it doesn't filter anything. On the second click still it shows it is checked and now it filters but one cannot keep it checked while checking other options.

**To Reproduce**
Simply do what I described above.

**Expected behavior**
When clicked and activated, it should show unassigned cards. but 

**Screenshots**
~If applicable, add screenshots to help explain your problem.~

**Client details:**
 - OS: Ubuntu
 - Browser Firefox
 - Version 145.0.2 (64-bit)
 - Device: desktop

<details>
<summary>Server details</summary>

**Operating system**: Ubuntu 24.04

**Web server:** Apache

**Database:** MariaDb

**PHP version:** 8

**Nextcloud version:** 31.0.9

**Where did you install Nextcloud from:** web installer (the same issue is seen in other installations too)

**Signing status:** ?

```
Login as admin user into your Nextcloud and access
http://example.com/index.php/settings/integrity/failed
paste the results here.
```

**List of activated apps:**

```
If you have access to your command line run e.g.:
sudo -u www-data php occ app:list
from within your Nextcloud installation folder

Enabled:
  - activity: 4.0.0
  - admin_audit: 1.21.0
  - app_api: 5.0.2
  - bruteforcesettings: 4.0.0
  - calendar: 5.5.9
  - circles: 31.0.0
  - cloud_federation_api: 1.14.0
  - collectives: 3.3.0
  - comments: 1.21.0
  - contactsinteraction: 1.12.0
  - dashboard: 7.11.0
  - dav: 1.33.0
  - deck: 1.15.5
  - external: 6.0.2
  - extract: 1.3.6
  - federatedfilesharing: 1.21.0
  - federation: 1.21.0
  - files: 2.3.1
  - files_archive: 1.2.8
  - files_downloadlimit: 4.0.0
  - files_pdfviewer: 4.0.0
  - files_reminders: 1.4.0
  - files_sharing: 1.23.1
  - files_trashbin: 1.21.0
[...]

PR #7448, opened 2025-12-11
Fix 'Unassigned' board filter'

theoholl:
* Resolves: #7427 
* Target version: main

### Summary

The issue was caused by the unassigned filter property not being initialized in the Vuex store's initial state.

https://github.com/user-attachments/assets/da705dee-865e-40ee-abf1-36145599366e

### Checklist

### DIFF
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -430,7 +430,7 @@ export default {
 			}
 		},
 		clearFilter() {
-			const filterReset = { tags: [], users: [], due: '', completed: 'both' }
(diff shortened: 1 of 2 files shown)


luka-nextcloud (approved):

juliusknorr:
@theoholl Great to see all those good contributions, that is highly appreciated. We can also invite you to the organization as a more frequent contributor so you can push your branches to this repo which makes collaboration a bit easier (e.g. github actions run automatically). We also have a public talk chat room for community contributors in case there is something to talk about regarding development. If you want to get an invite there, just let me know through the mail in my profile.

PR #7450, opened 2025-12-12
chore:bump @nextcloud/files to 4.0.0-beta.4

grnd-alt:
https://github.com/nextcloud/office/issues/32

seems like we only use formatFileSize from @nextcloud/files and I think this didn't have any changes


juliusknorr:
(Edited the description to remove the resolves work as that would autoclose the overview issue)

Issue #7449, opened 2025-12-12
Issue with Modifying Board and Tag Colors in Nextcloud Deck (v32.0.2)


frozenbluetw:
Hello,

A user recently reported an issue after NextCloud AIO was upgraded to version 32.0.2, where the board colors and tag colors in Deck can no longer be modified.

I attempted to reproduce the issue on my side and was able to confirm the same behavior. I have also tried clearing the browser cache, but the problem persists.

Please advise if there are any known fixes or if further investigation is required.

https://github.com/user-attachments/assets/7626b47d-2896-4883-af1c-a60e3ff472de

Deck version:1.16.2

mercury1337:
I can also confirm the issue - using a hosted Nextcloud on Hetzner 
Nextcloud Hub 31.0.10 / Deck 1.15.5

patrik-31415:
Found the same behaviour. Colours for tags and boards can't be edited. 
I'm on Nextcloud Hub 25 Autumn (32.0.1) and deck on 1.16.2

luka-nextcloud:
This is a duplicate of https://github.com/nextcloud/deck/issues/6651

PR #7462, opened 2025-12-14
Fix: update NcColorPicker event binding to use @update:value

theoholl:
* Resolves: #7449
* Target version: main

### Summary
Fix color picker event handlers for changing board and tag colors.

https://github.com/user-attachments/assets/5957e41f-15fe-4518-b852-8bb8be2cac27

TODO: The logic for changing the color of a new tag seems to be more complex.


luka-nextcloud:
Already fixed by https://github.com/nextcloud/deck/pull/7447

Issue #7485, opened 2025-12-19
Renaming a board should immediately update the title on the board


theoholl:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Current situation**
When renaming a board, the board name is not updated on the main view. To see the updated name there, one has to navigate somewhere else and back or reload the page.

https://github.com/user-attachments/assets/cbbe346f-74d0-4d71-84e6-fe83a2dddea8

**Describe the solution you'd like**
I would like the new board name to appear immediately after the new board name has been submitted. I think with a reactive framework like Vue and Vuex we can hopefully do that rather easily.

PR #7493, opened 2025-12-23
[main] Improvement: Show updated board title on the open board

theoholl:
* Resolves: #7485
* Target version: main

### Summary
The editing the title of the current board, the new title is displayed immediately when the user is on the current board.

https://github.com/user-attachments/assets/3f89fd4d-2244-4dd7-b32b-8350324fb888


luka-nextcloud:
@theoholl I think we should add cypress test for this case, could you please take care of it?

theoholl:
@luka-nextcloud I have added two Cypress tests. They look sane to me but I couldn't run them yet because the Cypress pipeline seems to have some pre-existing issues.

theoholl:
@grnd-alt, I think this is a nice UX improvement. Can we merge this?

grnd-alt (approved):

PR #7497, opened 2025-12-26
Input field for new cards moved to the top of the column

xstable:
Resolves: #6457

Target version: main

### Summary
Input field for new cards moved to the top of the column

### DIFF
--- a/css/deck.css
+++ b/css/deck.css
@@ -22,3 +22,11 @@ input[type=text]:focus+input[type=submit].icon-confirm,
 input[type=text]:hover+input[type=submit].icon-confirm {
 	border-color: var(--color-main-text) !important;
 }
+.app-deck {
+  .stack__card-add form {
+    order: 1;
+  }
+  .smooth-dnd-container.vertical {
+    order: 2;
+  }
+}


luka-nextcloud:
@xstable Thanks for your contribution. I don't think we need this PR since the issue https://github.com/nextcloud/deck/issues/6457 was already fixed on main branch. The related change: https://github.com/nextcloud/deck/pull/6640/files#diff-81373dc73b1e925b5c12bf0860a06f0def544c9b35b18dabec6c0548fad1d536R313

image

xstable:
@luka-nextcloud ok, thx for the info. 
Just a quick question (no criticism!): Why are there so much not approved PR's?
I've seen, that some are pretty old (2024). Are they all pending, or is there a way to clean them up (approve/remove them) so  that it don't look like the Project would be abandoned?

ziggycross:
Should #6457 be closed then, or is there still more in the works?

theoholl:
@luka-nextcloud @ziggycross 

I don't see this issue being fixed on `main` yet or perhaps if https://github.com/nextcloud/deck/pull/6640 had solved it, something might have changed since to break it again.

image

Issue #6680, opened 2025-01-21
Adding a new tag to a card has unexpected behaviour


reox:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Describe the bug**
When I add a new tag, which was previously not created, directly to a tag, it leads to the following unexpected behavior:
* The tag is correctly created and assigned to the card
* However, the colors in the list of available tags and the assigned tag are different. The assigned tag usually is gray.
* There are now two different behaviors:
  * a) closing the card
    * no tag shown in the board view
    * But the tag is still attached to the card
    * When opening the card again and selecting the same tag name from the list, the tag appears twice and the one with the wrong color can be deleted
    * If you close the card now _again_ and open it, you cannot delete the duplicated tag with wrong color! But you can delete the correct tag - which will also delete the wrong tag.
    * If before opening the card again, the board is reloaded, then both tags will have the same (correct) color. You can delete them independently, but if you close the card both will be gone, even if a single one was deleted.
  * b) staying in the card
    * If now the tag is selected from the list of available tags, nothing happens and the tag with the wrong color stays in the list
    * But the tag cannot be removed either. It can only be deleted when the card is closed and opened again
  

**To Reproduce**
1. Create a new tag from within the card:
![Image](https://github.com/user-attachments/assets/097dbad5-16ba-42a6-879b-987e049fad30)
2. Tags have now different color:
![Image](https://github.com/user-attachments/assets/396931b8-f41c-4bc1-a026-4acdd7fee66e)
3. No tag is shown for card "fazbaz":
![Image](https://github.com/user-attachments/assets/c0c78f76-460c-4417-81e0-0f485d573de6)
4. But tag is still there:
![Image](https://github.com/user-attachments/assets/10691768-d865-4f0d-804a-9aaaef7bae40)
5. Adding the same tag again duplicates the tag:
![Image](https://github.com/user-attachments/assets/7829c183-a879-433f-b9b5-bfc61b3eb6d3)

**Expected behavior**
If a new tag is created, it should behave like a normal tag

**Screenshots**
see above

**Client details:**
 - OS: W11
 - Browser: Firefox
 - Version: 134
 - Device: Desktop

<details>
[...]

PR #7503, opened 2026-01-02
[main] Fix: Adding new tag creates a 'ghost' tag

theoholl:
* Resolves: #6680
* Target version: main

### Summary
 - Preventing ghost/duplicate tag pills by letting the tag selector emit only labels with ids and by using NcSelect’s tracked id to avoid placeholder options. However, the user has to wait for the server to create a new tag and send it back to the client before it is getting rendered in the browser. This will result in a small delay before the new tag appears.
- Removed Vuex strict-mode warnings.

### Screencasts

**Before:**

https://github.com/user-attachments/assets/c75f1150-8b35-4c74-af7d-7b43c3d36092

**After:**

https://github.com/user-attachments/assets/96256af5-19e5-44e7-8c00-4242d1bea393

### DIFF
--- a/src/components/card/CardSidebarTabDetails.vue
+++ b/src/components/card/CardSidebarTabDetails.vue
@@ -152,17 +152,23 @@ export default {
 		}, 500),
 
 		addLabelToCard(newLabel) {
-			this.copiedCard.labels.push(newLabel)
-			const data = {
-				card: this.copiedCard,
-				labelId: newLabel.id,
+			if (!newLabel?.id || !this.card?.id) {
+				return
 			}
-			this.$store.dispatch('addLabel', data)
+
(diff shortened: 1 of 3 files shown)


grnd-alt:
Thanks a lot for your contribution 🫶 
I like the approach in general, but I think making the store logic more complex might not be the right solution as the existing functions are already in place and work properly.
The function calls to them are not correct however, I think we should keep the store logic as is and fix the calls for now.
I've created a PR with a simpler fix to not push to your fork: #7508 

As mentioned there we should think about adding _some_ more logic to the card Store, or at least be more declarative about what the addLabels function requires as params.
I am open to discuss on how to improve that as a follow up.

Issue #3534, opened 2022-01-13
Undelete for boards


HonkXL:
**Is your feature request related to a problem? Please describe.**
Sometimes user accidently deleate a board. This removes everything related with the board and a restore is possible, but not very easy because many database entries and (sometimes) files are affceted.
If a board is shared with outhers and they have the right to manage the board, it can be deleted by them.
Sometimes this happes unwanted, because the user just want to get a better overview and since there is no grouping or hiding of boards is possible, the user creates this overview by deleting the board for everyone. Sure it's a fail of the user, but it's very bad to fix this, after it has been deleted.

**Describe the solution you'd like**
A board should just marked for deletion and the delete job should not run 5 minutes after marked for deletion. Better would be to make this time span configurable, so that a manager or admin can undelete the board without restoring the DB.

**Describe alternatives you've considered**
Alternative would be, that a deleted board would be exported to a textfile (or json or whatever) on server and could be reimported by admin.

dodedodo:
There's already a brief period where deletion can be undone by the user. If you know the board ID it's even accessible while in a 'deleted' state. Maybe that time could be increased (It's ~7 seconds right now). Or maybe the undo function could be redesigned as a recycle bin?

HonkXL:
Yes, but 7 seconds is a bit to short when you detect the mistake on the next day.
I think a kind of recycle bin would be really helpful, because there is no really good recovery strategy for undoing deleted boards.

jivanpal:
FWIW, boards don't seem to actually get deleted (see #2198). If we decide to implement this, it is probably best to tackle this and #2198 together.

PR #7505, opened 2026-01-04
[main] Fix: Board cannot be undeleted

theoholl:
* Resolves: #7506 
* Target version: main

Related to: [#3534](https://github.com/nextcloud/deck/issues/3534)

### Summary
On `main` (and maybe other branches/versions), the undelete action is broken. The web endpoint returns a `403 Permission denied` error. 

This is because the `PermissionService::getBoard()` method is called without the allowDeleted parameter (it defaults to false), and, thus, throws a `DoesNotExistException` error. This error is then caught in `PermissionService::getPermissions()`, setting both `$owner = false` and `$acls = []`. This results in all permissions being false.

This pull request adds an optional `allowDeleted` parameter to relevant functions.

### Screencast

https://github.com/user-attachments/assets/ff2fee69-ad0e-415c-95bd-6ac02b5c4809

### DIFF
--- a/lib/Service/BoardService.php
+++ b/lib/Service/BoardService.php
@@ -246,7 +246,7 @@ class BoardService {
 	public function deleteUndo(int $id): Board {
 		$this->boardServiceValidator->check(compact('id'));
 
-		$this->permissionService->checkPermission($this->boardMapper, $id, Acl::PERMISSION_MANAGE);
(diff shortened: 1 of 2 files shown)


grnd-alt (approved):

grnd-alt:
thanks for your contribution, the code changes look fine, and work as expected 🔥 
The test runs seem to be unrelated!

luka-nextcloud (approved):

PR #7508, opened 2026-01-06
fix: add new tags to card and board

grnd-alt:
* Resolves: #6680
* Target version: main

### Summary
Thanks to @theoholl for coming up with a contribution.
This is a simpler fix, to not over complicate the card store logic.

We should think about some refinements to the way labels are added in the frontend in general though.
(we have no description at all on what kind of props 'addLabel' requires e.g.)

### Checklist

### DIFF
--- a/src/components/card/TagSelector.vue
+++ b/src/components/card/TagSelector.vue
@@ -75,7 +75,10 @@ export default {
 	},
 	methods: {
 		onSelect(options) {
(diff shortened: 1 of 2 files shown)


theoholl:
Brilliant, this looks like a clean and simple solution! Thank you! :octocat:

juliusknorr (approved):

PR #7510, opened 2026-01-07
Fix/display labels with invalid colors

grnd-alt:
Not validating the hex values in php resulted in the possibility to set any string as a color value for labels.
not parseable color strings broke the ui of labels alltogether.

This PR validates if the color is valid hex in the backend.

In the frontend the text color of the tag is set depending of the color of the tag to black or white. In this code path the color is now also validated to repair boards where there is already a tag with an invalid color.


juliusknorr:
Needs backports, right?

PR #7515, opened 2026-01-09
fix: Correct typing

artonge:
- phpunit issues are fixed in https://github.com/nextcloud/deck/pull/7516
- integration test fixed in: https://github.com/nextcloud/deck/pull/7517


grnd-alt:
rebased to get ci fixes from main

PR #7545, opened 2026-01-22
feat(deps): Add Nextcloud 34 support on main

nickvergessen:
Signed-off-by: Joas Schilling <coding@schilljs.com>


juliusknorr:
@grnd-alt Let's observe a bit if cypress always fails on the same test, then we could just skip that one case and have green CI again

PR #7546, opened 2026-01-22
fix: Use strict comparaison

CarlSchwan:
### Checklist


grnd-alt:
rebased, so cypress can run on nc34

PR #7548, opened 2026-01-22
ci(actions): Update all actions

nickvergessen:
- Best reviewed commit by commit
- :warning: Please note "ci(action)!: Remove action that builds each push" and judge if okay, otherwise that file needs fixing, but I felt it is over the top


nickvergessen:
I can fix psalm and the current cypress step

But not a lot ideas about Node CI

PR #7602, opened 2026-02-05
fix: improve wording for global search results

grnd-alt:


juliusknorr:
Only relevant for 32?

PR #7626, opened 2026-02-08
Fix missing due date remember icons

### LINKED ISSUE #7511: missing icons on shared boards dropdown
### Steps to reproduce
* create a Board and share it with at least one user
* click on the three dots icon (which is missing)
* observe the missing/white icon and wrong bold text in the menu (see screenshot)

Image

### Expected behavior
* all icons are visible
* Due date reminders is not bold

SideNote: without looking at the code path in detail the menu should not be that different between shared boards and unshared boards, maybe we can unify it.

theoholl:
* Resolves: #7511
* Target version: main

### Summary

- Update the three-dot menu to contain a regular 'Due date reminder' item with an icon
- Load due date reminder settings lazily

https://github.com/user-attachments/assets/de829c8d-0c65-40e9-bd09-f3e739d3c808


grnd-alt (changes requested):
one improvement that would be nice for ux, otherwise works well 🔥

Issue #7588, opened 2026-01-26
Archived Cards not shown when searching in specific deck


bentuna:
**Is your feature request related to a problem? Please describe.**
When I search on the home page of Deck, I get results from all decks and no matter if a card is archived or not. When I search for the same term while being in a specific deck, I also get the "results from all boards" but no archived cards. So in that case it appears to the user that there is no such card.

**Describe the solution you'd like**
Showing archived and not archived cards when searching, no matter where you are inside the app.

samin-z:
Hello @bentuna 
the functionality of search in 'Boards' is the correct behavior, 
1. we have the global search which shows all the archived/unarchived cards from all boards (and also all boards).
2. the results shown for search inside the active board is also noted as 'Search for <keyword> in other boards' which does not include search results form the current one.
3. we also have the option from the current board to show archived/unarchived cards.

<img width="256" height="171" alt="Image" src="https://github.com/user-attachments/assets/816f619a-41d8-44dd-b2c5-97d4b3376a7a" />

so the functionality is working as it should.

samin-z:
I'll close this ticket as there has not been any response and the functionality of the feature is as it's supposed to be. in case of request for a change a new Ticket can be opened.

PR #7641, opened 2026-02-13
#7588: Archived Cards not shown when searching in specific deck

samin-z:
* Resolves: [#7588](https://github.com/nextcloud/deck/issues/7588)
* Target version: main

### Summary
If user inside Board A search for keyword 'Card', if Board A has two cards of Card A and Card B (which is archived), will not be shown in search result.

### TODO
Updated the login that inside Board A when search for keyword 'Card', the archived result (Card B) show to the user.
### Checklist

### DIFF
--- a/src/components/search/GlobalSearchResults.vue
+++ b/src/components/search/GlobalSearchResults.vue
@@ -87,17 +87,18 @@ export default {
 			searchQuery: state => state.searchQuery,
 		}),
 		filteredResults() {
-            const sortFn = (a, b) => a.archived - b.archived || b.lastModified - a.lastModified
-            if (this.$route.params.id) {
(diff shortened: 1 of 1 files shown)


samin-z:
> So maybe we can adapt the filter in some way to have an option to show archived and unarchived and automatically set that when using the search bar.

Thank you for your feedback, in that case i would say lets close this PR as the functionality is as it's supposed to be (ill drop a message on the main issue ticket and explain) because for this it's as you said, it searched all other board and if they want to see full results (un/archived) then user can use global search.
One point could be (for later as its not high priority) to maybe with a talk with design team-if needed-change the functionality to make it more simple for users to understand.

Issue #7542, opened 2026-01-20
Show Creator Information in Deck Cards


pwstechnik:
Description:
Currently, when a user creates a problem or a to-do in Nextcloud Deck, there is no visible information about who created the card. This makes it difficult to identify the responsible person or the point of contact for follow-up questions or clarifications.

Feature Request:
Please add a feature that displays the creator's name (and optionally the creation date) directly on each card in Nextcloud Deck. This would improve transparency and accountability within teams.

Use Case:
Teams can quickly identify who created a task or issue.
Users can reach out to the creator for further details if needed.

Additional Context:
This feature is common in many project management tools (e.g., Trello, Jira) and would align Nextcloud Deck with industry standards.

That would help us a lot :)

putt1ck:
Does it not show in the activity section?

samin-z:
> Does it not show in the activity section?

yes, but this is for detail section overview, easier for user to see if the activity is not active.

PR #7650, opened 2026-02-16
#7542:show creator and created date in respective card

samin-z:
* Resolves: [#7542](https://github.com/nextcloud/deck/issues/7542)
* Target version: main

### Summary
Show more info regarding card like creator and date the card was created

### Checklist

Screenshot 2026-04-14 at 17 04 53

### DIFF
--- a/src/components/card/CardSidebar.vue
+++ b/src/components/card/CardSidebar.vue
@@ -8,7 +8,6 @@
 		ref="cardSidebar"
 		:active="tabId"
 		:name="displayTitle"
-		:subname="subtitle"
 		:subtitle="subtitleTooltip"
 		:name-editable.sync="isEditingTitle"
 		@update:name="(value) => titleEditing = value"
@@ -17,6 +16,13 @@
 		@submit-name="handleSubmitTitle"
 		@opened="focusHeader"
 		@close="closeSidebar">
+		<template #subname>
+			<span>{{ subtitle }}</span>
+			<template v-if="cardOwner">
+				<span> ⸱ </span>
+				<NcUserBubble :user="cardOwner.uid" :display-name="cardOwner.displayName" />
+			</template>
+		</template>
 		<template #secondary-actions>
 			<NcActionButton v-if="cardDetailsInModal && isFullApp" icon="icon-menu-sidebar" @click.stop="closeModal()">
 				{{ t('deck', 'Open in sidebar view') }}
@@ -75,7 +81,7 @@
 </template>
 
 <script>
-import { NcActionButton, NcAppSidebar, NcAppSidebarTab } from '@nextcloud/vue'
+import { NcActionButton, NcAppSidebar, NcAppSidebarTab, NcUserBubble } from '@nextcloud/vue'
 import { NcReferenceList } from '@nextcloud/vue/dist/Components/NcRichText.js'
 import { getCapabilities } from '@nextcloud/capabilities'
 import { mapState, mapGetters } from 'vuex'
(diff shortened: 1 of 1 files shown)


samin-z:
@grnd-alt @elzody 
hey, i added these info that user requested, but not sure if we should go over a design first or can just add them. any suggestions are greatly welcomed.

pipeman1981:
Hej All, hi grnd-alt,
I'm really looking forward to this feature and would be completely happy if it could be realised like you suggested.
Regards,
Carsten

samin-z:
@grnd-alt, ill also apply ownership point we talked about here too in this task.

src/components/card/CardSidebar.vue
```diff
@@ -153,11 +153,14 @@ export default {
...
+		cardOwnerDisplayName() {
+			return this.currentCard.owner?.displayname ?? this.currentCard.owner?.uid ?? this.currentCard.owner ?? null
+		},
 		subtitle() {
-			return t('deck', 'Modified') + ': ' + this.relativeDate(this.currentCard.lastModified * 1000) + ' ⸱ ' + t('deck', 'Created') + ': ' + this.relativeDate(this.currentCard.createdAt * 1000)
+			return this.buildSubtitle()
 		},
 		subtitleTooltip() {
-			return t('deck', 'Modified') + ': ' + this.formatDate(this.currentCard.lastModified) + '\n' + t('deck', 'Created') + ': ' + this.formatDate(this.currentCard.createdAt)
+			return this.buildSubtitleTooltip()
```
grnd-alt: why have this in an extra method and not just in the computed?

src/components/card/CardSidebar.vue
```diff
@@ -190,6 +193,34 @@ export default {
 		},
 	},
 	methods: {
+		buildSubtitle() {
+			const modifiedDate = this.relativeDate(this.currentCard.lastModified * 1000)
```
grnd-alt: I think we should remove this entire building code and just use the subname as a slot and not as a prop, that way we can use NcUserBubble to display the user with actual profile picture etc.

with <template #subname> <CONTENT> </template> you can render components inside the subname

src/components/card/CardSidebar.vue
```diff
@@ -190,6 +193,34 @@ export default {
...
+					undefined, { escape: false }),
+			}
+			return owner ? cardMeta.withOwner : cardMeta.generic
+		},
+		buildSubtitleTooltip() {
+			const modifiedDate = this.formatDate(this.currentCard.lastModified)
+			const createdDate = this.formatDate(this.currentCard.createdAt)
+			const owner = this.cardOwnerDisplayName
+			const cardMeta = {
+				generic: t('deck', '{modified}: {modifiedDate}\n{created}: {createdDate}',
```
grnd-alt: these translation strings don't need to be that modular, only actual variables should be in the {}.
like
`t('deck', 'Modified: {modifiedDate}', {modifiedDate: modifiedDate})`
same applies for the subtitle itself.

src/components/card/CardSidebar.vue
samin-z: because of the previous comment about using proper translation formatting, i extracted the logic into separate methods for better readability, with the new changes the logic simplified enough to be moved back into the computed properties.

samin-z:
/backport ffb1fd7d9f2719c84bce113c2583029c7bf88cd9 to stable33

samin-z:
/backport ffb1fd7d9f2719c84bce113c2583029c7bf88cd9 to stable32

samin-z:
/backport ffb1fd7d9f2719c84bce113c2583029c7bf88cd9 to stable31

samin-z:
/backport ffb1fd7d9f2719c84bce113c2583029c7bf88cd9 to stable30

samin-z:
/backport ffb1fd7d9f2719c84bce113c2583029c7bf88cd9 to stable29

Issue #2399, opened 2020-10-08
Write access over CalDAV


juliushaertl:
Split off from https://github.com/nextcloud/deck/issues/15 so we have individual issues to track:

Current plan from the discussion below:

| Deck | CalDAV | Tasks |
|-|-|-|
| Board | `VCALENDAR` | List |
| Stack | `VTODO` without `RELATED-TO` property | Root-Task |
| Card | `VTODO` with `RELATED-TO` (UUID of the Root-Task / Stack) | Subtask |
| Tags | [`CATEGORIES`](https://tools.ietf.org/html/rfc5545#section-3.8.1.2) or custom implementation (Nextcloud tagging system) | Currently uses `CATEGORIES`, integration with Nextcloud tags would be better, see https://github.com/nextcloud/tasks/issues/89 |
| Assigned users | [`ATTENDEE`](https://tools.ietf.org/html/rfc5545#section-3.8.4.1) or custom implementation | [`ATTENDEE`](https://tools.ietf.org/html/rfc5545#section-3.8.4.1) or custom implementation, see https://github.com/nextcloud/tasks/issues/72 |
| Recurring stacks & cards | [`RRULE`](https://tools.ietf.org/html/rfc5545#section-3.8.5.3) | Recurring tasks & subtasks, see https://github.com/nextcloud/tasks/issues/34 |
| Attachments | [`ATTACH`](https://tools.ietf.org/html/rfc5545#section-3.8.1.1) or custom implementation | [`ATTACH`](https://tools.ietf.org/html/rfc5545#section-3.8.1.1) or custom implementation, see https://github.com/nextcloud/tasks/issues/91 |

> Tags, assigning users, attachments and recurrence is not supported by Tasks yet either, but it would make much sense to implement these features. And if we see that it is not really possible to implement this with pure CalDAV I would be open to implement it without syncing it via CalDAV and have it only available in the webinterface. This would still be better than not having it at all in Tasks and would not limit Deck.

1. :heavy_check_mark: Expose deck boards as VCALENDAR, try to map as much data from deck to the VTODO (read only access) 
2. :white_check_mark: **Allow updating VCALENDAR entries, try to map as much as possible from vtodo to the deck database and store the rest as plain vtodo data, so we can still push it out over CalDAV (write access)**
3. :white_check_mark: Deck data could be moved to the DAV backend.

tensau:
Is it realistic that this issue could be solved with the 1.3.0 release?

[...]

PR #7655, opened 2026-02-18
Write access over CalDAV

Jaggob:
## Summary

This PR implements a first small part of #2399: updating existing Deck cards via CalDAV.

The scope is intentionally limited to `PUT` on existing card VTODOs. Create, delete, move, tags/categories, attendees, recurrence, attachments, and lossless raw VTODO round-tripping are left for follow-up PRs.

## Implemented

- Existing Deck cards can be updated via CalDAV `PUT`.
- Write ACLs are exposed only for users with board edit permissions.
- Stack VTODOs remain read-only.
- `CREATE` and `DELETE` remain forbidden.
- Supported VTODO mapping:
  - `SUMMARY` -> card title
  - `DESCRIPTION` -> card description
  - `DUE` -> due date
  - `STATUS`, `COMPLETED`, `PERCENT-COMPLETE` -> done state
- Unsupported VTODO fields are ignored and not stored.

## Not Included

Intentionally out of scope:

- Creating new cards (needs a clear CalDAV object identity / href strategy)
- Deleting cards
- Moving cards between stacks
- Mapping `RELATED-TO`
- Syncing `CATEGORIES` / Deck tags
- Handling `ATTENDEE`, `RRULE`, attachments, or raw VTODO data

Unsupported properties may therefore be accepted on `PUT`, but they will not be returned by the next `GET`.

## Notes

`DELETE` is deferred because some clients implement moves as `DELETE` + `CREATE`; allowing delete while create is forbidden could make failed moves destructive.

## Testing

Added unit tests for update mapping, done-state handling, read-only stack objects, forbidden create/delete, ACL behavior, ETag refresh, invalid payloads, stream PUT payloads, exception mapping, and permission-cache behavior.

Manually tested with Thunderbird 150.0.1 and macOS Reminders 26.4.1


lib/Db/CardMapper.php
```diff
@@ -134,7 +134,8 @@ public function find($id, bool $enhance = true): Card {
 	 * @return Card[]
 	 * @throws \OCP\DB\Exception
 	 */
-	public function findAll($stackId, ?int $limit = null, int $offset = 0, int $since = -1) {
+	public function findAll($stackId, ?int $limit = null, ?int $offset = 0, int $since = -1) {
+		$offset ??= 0;
```
grnd-alt: ???

lib/DAV/DeckCalendarBackend.php
```diff
@@ -30,22 +38,29 @@ class DeckCalendarBackend {
...
 		$this->boardMapper = $boardMapper;
+		$this->labelService = $labelService;
+		$this->configService = $configService;
 	}
 
 	public function getBoards(): array {
 		return $this->boardService->findAll(-1, false, false);
 	}
 
+	/** @psalm-suppress InvalidThrow */
```
grnd-alt: I don't see an InvalidThrow

lib/DAV/DeckCalendarBackend.php
```diff
@@ -70,4 +85,806 @@ public function getChildren(int $id): array {
...
+
+		return null;
+	}
+
+	/** @return Card[] */
+	public function getChildrenForStack(int $stackId): array {
+		return $this->stackService->find($stackId)->getCards() ?? [];
+	}
+
+	public function getCalDavListMode(): string {
```
grnd-alt: unused?

grnd-alt (changes requested):
Hey, thanks for your contribution, would be nice to get this feature in. I have some smaller code-style remarks, that are partly up for discussion.

The most important remark is that one though: https://github.com/nextcloud/deck/pull/7655/changes#r2840797696

I think this should be oriented to work with tasks as much as possible before merging.

lib/DAV/DeckCalendarBackend.php
```diff
@@ -64,10 +84,907 @@ public function updateBoard(Board $board): bool {
...
+		}
+		if ($left === null || $right === null) {
+			return false;
+		}
+
+		return $left->getTimestamp() === $right->getTimestamp();
+	}
+
+	private function isSabreVCalendar($value): bool {
+		/** @psalm-suppress UndefinedClass */
```
grnd-alt, line 960: wrapper function only to suppress psalm? same for VTodo

lib/DAV/DeckCalendarBackend.php
```diff
@@ -64,10 +84,907 @@ public function updateBoard(Board $board): bool {
...
+		if (count($todos) === 0 || !$this->isSabreVTodo($todos[0])) {
+			throw new \InvalidArgumentException('Calendar payload contains no VTODO');
+		}
+		return $todos[0];
+	}
+
+	private function extractStackIdFromRelatedTo($todo): ?int {
+		$parentCandidates = [];
+		$otherCandidates = [];
+		foreach ($todo->children() as $child) {
```
grnd-alt, line 516: this for loop can be something like this using the sabre types to not do the property handling etc yourself, then you could also remove some of the helper functions.
```
		$relatedToArray = $todo->select("RELATED-TO");
		foreach($relatedToArray as $relatedTo) {
			$reltype = $relatedTo["RELTYPE"] ?? null;
			if ($reltype instanceof \Sabre\VObject\Parameter) {
				$reltypeValue = $reltype->getValue();
				if ($reltypeValue === 'PARENT') {
					$parentCandidates[] = (string)$relatedTo;
				} else {
					$otherCandidates[] = (string)$relatedTo;
				}
			}
		}
```

lib/DAV/DeckCalendarBackend.php
```diff
@@ -64,10 +84,907 @@ public function updateBoard(Board $board): bool {
...
+		}
+		if ($relatedStackId === null) {
+			$relatedStackId = $this->inferStackIdFromTodoHints($boardId, $todo, $mode);
+		}
+		if ($relatedStackId !== null) {
+			$stackId = $this->resolveStackIdForBoard($boardId, $relatedStackId);
+		} elseif ($targetBoardId !== null && $currentBoardId !== $targetBoardId) {
+			$stackId = $this->getDefaultStackIdForBoard($targetBoardId);
+		} else {
+			$stackId = $card->getStackId();
```
grnd-alt, line 399: this entire stackId handling is not very easily readable, and seems overly complex, it does not make clear *why* the stackId is selected the way it is.

lib/DAV/DeckCalendarBackend.php
```diff
@@ -64,10 +84,907 @@ public function updateBoard(Board $board): bool {
...
+
+	private function findCardByIdIncludingDeleted(int $cardId): ?Card {
+		try {
+			return $this->cardService->find($cardId, true, false);
+		} catch (\Throwable $e) {
+			return null;
+		}
+	}
+
+	private function normalizeDavUriForStorage(?string $name): ?string {
```
grnd-alt, line 901: weird naming, does not seem to normalize anything but rather validate.

lib/Service/CardService.php
```diff
@@ -156,6 +162,18 @@ public function find(int $cardId): Card {
 		return $card;
 	}
 
+	public function findByDavUriLite(string $davUri, ?int $boardId = null, ?int $stackId = null, bool $includeDeleted = true): Card {
```
grnd-alt, line 165: why is this named lite?

lib/Service/CardService.php
```diff
@@ -176,13 +194,16 @@ public function findCalendarEntries(int $boardId): array {
...
+	public function create(string $title, int $stackId, string $type, int $order, string $owner, string $description = '', $duedate = null, ?string $davUri = null): Card {
 		$this->cardServiceValidator->check(compact('title', 'stackId', 'type', 'order', 'owner'));
 
 		$this->permissionService->checkPermission($this->stackMapper, $stackId, Acl::PERMISSION_EDIT);
 		if ($this->boardService->isArchived($this->stackMapper, $stackId)) {
 			throw new StatusException('Operation not allowed. This board is archived.');
 		}
+		$stack = $this->stackMapper->find($stackId);
+		$boardId = $stack->getBoardId();
+		$board = $this->boardMapper->find($boardId);
```
grnd-alt, line 206: I don't think those have any benefit as the activitymanager gets board and stack data

lib/Service/CardService.php
```diff
@@ -233,13 +261,17 @@ public function delete(int $id): Card {
...
 	 */
-	public function update(int $id, string $title, int $stackId, string $type, string $owner, string $description = '', int $order = 0, ?string $duedate = null, ?int $deletedAt = null, ?bool $archived = null, ?OptionalNullableValue $done = null): Card {
+	public function update(int $id, string $title, int $stackId, string $type, string $owner, string $description = '', int $order = 0, ?string $duedate = null, ?int $deletedAt = null, ?bool $archived = null, ?OptionalNullableValue $done = null, ?int $knownBoardId = null): Card {
 		$this->cardServiceValidator->check(compact('id', 'title', 'stackId', 'type', 'owner', 'order'));
 
-		$this->permissionService->checkPermission($this->cardMapper, $id, Acl::PERMISSION_EDIT, allowDeletedCard: true);
+		if ($knownBoardId !== null) {
+			$this->permissionService->checkPermission($this->boardMapper, $knownBoardId, Acl::PERMISSION_EDIT);
+		} else {
+			$this->permissionService->checkPermission($this->cardMapper, $id, Acl::PERMISSION_EDIT, allowDeletedCard: true);
```
grnd-alt, line 270: only doing this check if knownBoardId is not passed feels like a security issue. If KnownBoardId does not match the board the card's on it's possible to move a card from a board where a user has only read permission. It looks like this is verified for in `getBackendChildren()` but that is multiple functions deep and very implicit.

grnd-alt (changes requested):
I did not manage to grasp everything in here, but this feels quite spaghetti and hard to read or follow tbh. I do think this can be implemented in a more readable manner, but might as well be a skill issue on code reading on my side. However it would be nice if this could be split up to e.g. not have the list_modes in there as well and get this to be easier to review

src/components/DeckAppSettings.vue
```diff
@@ -22,6 +22,12 @@
 				<NcFormBoxSwitch v-model="configCalendar"
 					:label="t('deck', 'Show boards in calendar/tasks')" />
 			</NcFormBox>
+			<NcSelect v-model="caldavListModeSelection"
+				:options="caldavListModeOptions"
+				:clearable="false"
+				label="label"
+				track-by="id"
+				:input-label="t('deck', 'CalDAV list mapping mode')" />
```
blizzz: as a technical user, I do not know what this means. Seeing the possible value the meaning gets clearer, the main label should still be easier to understand.

Also, isn't this out of scope for the writable caldav feature and should be split into a separate PR?

lib/Db/Card.php
```diff
@@ -66,6 +68,7 @@ class Card extends RelationalEntity {
 	protected string $title = '';
 	protected $description;
 	protected $descriptionPrev;
+	protected $davUri = null;
```
blizzz: type is missing

lib/Db/Card.php
```diff
@@ -123,34 +127,54 @@ public function setDatabaseType($type) {
...
+		$json = parent::jsonSerialize();
+		unset($json['davUri']);
+		return $json;
+	}
+
 	public function getCalendarObject(): VCalendar {
 		$calendar = new VCalendar();
 		$event = $calendar->createComponent('VTODO');
 		$event->UID = 'deck-card-' . $this->getId();
+		$event->{'X-NC-DECK-CARD-ID'} = (string)$this->getId();
```
blizzz: this is already part of the UID?

lib/Db/Card.php
```diff
@@ -123,34 +127,54 @@ public function setDatabaseType($type) {
...
 		if ($this->getDone() || $this->getArchived()) {
-			$date = new DateTime();
-			$date->setTimestamp($this->getLastModified());
 			$event->STATUS = 'COMPLETED';
-			$event->COMPLETED = $this->getDone() ? $this->getDone() : $this->getArchived();
+			$event->{'PERCENT-COMPLETE'} = 100;
+			if ($this->getDone()) {
+				$event->COMPLETED = $this->getDone();
+			} else {
+				$event->COMPLETED = $lastModified;
```
blizzz: looks strange?

lib/Db/Card.php
```diff
@@ -123,34 +127,54 @@ public function setDatabaseType($type) {
...
+			}
 		} else {
 			$event->STATUS = 'NEEDS-ACTION';
+			$event->{'PERCENT-COMPLETE'} = 0;
```
blizzz: why?

lib/Db/CardMapper.php
```diff
@@ -130,6 +130,36 @@ public function find($id, bool $enhance = true): Card {
...
+				->where($qb->expr()->eq('dav_uri', $qb->createNamedParameter($davUri, IQueryBuilder::PARAM_STR)));
+		}
+
+		if (!$includeDeleted) {
+			$deletedColumn = $boardId !== null && $stackId === null ? 'c.deleted_at' : 'deleted_at';
+			$qb->andWhere($qb->expr()->eq($deletedColumn, $qb->createNamedParameter(0, IQueryBuilder::PARAM_INT)));
+		}
+
+		$qb->orderBy('order')
+			->addOrderBy('id');
```
blizzz: The return code later suggest only one specific entry is returned, then why sorting?

lib/DAV/DeckCalendarBackend.php
blizzz: Strange to have the parameter nullable, but OK. Personally, I would expect an exception thrown in the negative case.

Is this user controlled, though? Is the check sufficient?

lib/Activity/ActivityManager.php
```diff
@@ -521,16 +526,23 @@ private function findDetailsForStack($stackId) {
 		];
 	}
 
-	private function findDetailsForCard(int $cardId, ?string $subject = null): array {
-		$card = $this->cardMapper->find($cardId);
-		$stack = $this->stackMapper->find($card->getStackId());
-		$board = $this->boardMapper->find($stack->getBoardId());
+	private function findDetailsForCard($cardOrId, ?string $subject = null, ?Stack $knownStack = null, ?array $knownBoard = null): array {
```
blizzz: why is this extended? It does not seem that this is related to the feature?

blizzz (commented):
I am confused about wide parts of this PR. Some aspects maybe makes sense, but do not really belong here, about others I am really puzzled. Maybe we can extract the necessary parts and have a very specific and lean as possible change set? Aspects that are not directly related to the write access can still be implemented in a separate PR.

Jaggob:
Thanks for all the feedback, and sorry for the chaotic first version of this PR. I tried to cover too much of the CalDAV write feature and future improvements at once and ended up neglecting code quality and readability.

I have now reduced the PR to a much smaller first step: updating existing Deck cards via CalDAV. Also updated the PR description with the current scope and a list of things that are intentionally left for follow-up work.

For reference (and because I still think it might be a good way to deal with the create mapping flow), the previous broader implementation with the different create stack mapping approaches is still available here:
https://github.com/Jaggob/deck/tree/backup/feature-cal-dav-write-full-a2dce36

madaxel:
Hi @grnd-alt , hi @blizzz ,
are the latest changes more structured and readable? As far as I can see these look promising, but I'm just a user in nextcloud context ;-)

Thanks for the effort

PR #7659, opened 2026-02-19
fix: Handle share attributes in the share provider

grnd-alt:


lib/Sharing/DeckShareProvider.php
```diff
@@ -281,8 +297,14 @@ public function update(IShare $share) {
...
-			->set('expiration', $qb->createNamedParameter($share->getExpirationDate(), IQueryBuilder::PARAM_DATE))
-			->executeStatement();
+			->set('expiration', $qb->createNamedParameter($share->getExpirationDate(), IQueryBuilder::PARAM_DATE));
+
+		$shareAttributes = $this->formatShareAttributes($share->getAttributes());
+		if ($shareAttributes !== null) {
+			$qb->set('attributes', $qb->createNamedParameter($shareAttributes));
+		}
+
+		$qb->executeStatement();
```
blizzz: also not necessarily in this scope, but since we run three queries within `update()` make its worth to combine them in a transaction.

Issue #6704, opened 2025-01-28
Renaming a board and pressing enter does not save the new name


elzody:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Describe the bug**
When creating a new board, you type in the name and you can press Enter to save it. When renaming a board, however, pressing enter does not save the new name and focus is lost from the input field. Additionally, when you click "Edit board," the input field is not automatically focused as I would expect.

**To Reproduce**
1. Go to Deck
2. Click on "Add board"
3. Type in a name, press Enter
4. Click on the new board's action menu (three dots)
5. Click "Edit board"
6. The input field is not focused by default
7. Entering a new name and pressing Enter does not save the new name

**Expected behavior**
The new board name is saved and applied, and the input field would be focused by default when it appears.

**Video**
https://github.com/user-attachments/assets/702c634f-4f4f-4991-a82d-d8a103c79b30

**Client details:**
 - OS: Arch Linux x86_64 (6.12.10-arch1-1)
 - Browser: Zen (Firefox)
 - Version: main (presumably 1.15.0-beta.1)
 - Device: Laptop (Tuxedo Pulse 14)

biguenique:
Hint: I successfully managed to rename my boards through the Nextcloud Deck mobile app (v 1.24.5/Android, with Nextcloud server v30.0.10/AIO v10.13.0).

biguenique:
I'm still experiencing the same bug after upgrading to Deck v1.15.1 / Nextcloud v31.0.6 (Hub 10) / AIO v11.1.0.
Need to use the mobile app to rename decks.

Ron-Dot-Star:
Line 127 in deck/src/components/navigation/AppNavigationBoard.vue

Can we just put @keyup.enter="applyEdit" here?

PR #7662, opened 2026-02-19
fix(navigation): fix Enter/Escape key behavior in board edit form

ramteid:
The Cancel button had `native-type="submit"`, making it the first submit button in the form. Due to HTML implicit form submission, pressing Enter in the text field activated the Cancel button instead of saving changes.

- Change Cancel button `native-type` from `"submit"` to `"button"` so that the Save button becomes the form's implicit submit target
- Add `@keyup.esc` handler on the form to cancel editing on Escape

* Resolves: 
#6704
#7105 
* Target version: main

### Summary

When editing a board name in the sidebar navigation, pressing Enter discards changes instead of saving them, and pressing Escape does nothing. This is counter-intuitive behavior (and annoying tbh).

The root cause is that the Cancel button had `native-type="submit"`, making it the first submit button in the `<form>`. Per the [HTML spec for implicit submission](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#implicit-submission), pressing Enter in a text input activates the form's first submit button — which was Cancel, not Save.

| Key | Before | After |
|-----|--------|-------|
| Enter | Cancels editing (discards changes) | Saves the board |
| Escape | Does nothing | Cancels editing |

### TODO

- [x] Fix Cancel button `native-type` from `"submit"` to `"button"`
- [x] Add `@keyup.esc` handler to cancel editing on Escape

### Checklist

- [] Tests (unit, integration, api and/or acceptance) are included

### DIFF
--- a/src/components/navigation/AppNavigationBoard.vue
+++ b/src/components/navigation/AppNavigationBoard.vue
@@ -133,7 +133,7 @@
 			<NcColorPicker v-model="editColor" class="app-navigation-entry-bullet-wrapper">
 				<button :style="{ backgroundColor: getColor }" class="color0 icon-colorpicker app-navigation-entry-bullet" />
 			</NcColorPicker>
-			<form @submit.prevent.stop="applyEdit">
+			<form @submit.prevent.stop="applyEdit" @keyup.esc="cancelEdit">
 				<NcTextField ref="inputField"
 					:disable="loading"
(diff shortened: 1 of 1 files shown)


ramteid:
Somehow I didn't see the related issues were closed already - the fix was made only 3 days ago on main and apparently not yet reached my Nextcloud instance :)

PR #7664, opened 2026-02-20
fix(i18n): Use positional placeholders

rakekniven:


rakekniven:
Any chance to get a review?
@juliusknorr maybe?

PR #7672, opened 2026-02-21
fix: hide upload button if user has no quota

solracsf:
### Summary

If user has no quota, don't show the option as it will only lead to an error.
`0 B` quota can be defined exactly to restrict uploads.

Similar to https://github.com/nextcloud/server/blob/0a6831085994b9d388fb79805df8fed2a5597abe/apps/files_sharing/src/files_views/shares.ts#L62-L64

### Checklist


solracsf:
@grnd-alt can we backport this?

grnd-alt:
> @grnd-alt can we backport this?

Sure go ahead!

Issue #7673, opened 2026-02-22
Past card activities on a shared board not visible


ramteid:
**Describe the bug**
When another person shared a board with you, you cannot see card activities that happened before the board was shared with you. For example the creation activity.

**To Reproduce**
1. User A creates a board
2. User A creates cards and modifies it to generate entries under Activity
3. User A shares the board with user B
4. User B cannot see Activities on any card from before the board was shared
5. Subsequent activities on that board are visible to user B

**Expected behavior**
When a board is shared with you, you'd expect to see all activities for all visible cards.

luka-nextcloud:
@ramteid As I understand, this is expected behavior since the Activity app does not provide any function to show activities to the user before the board/file was shared with that user. Perhaps this issue should be raised at the Activity app.

PR #7674, opened 2026-02-22
fix(activity): Show full card history to users added to a shared board

ramteid:
When someone gets access to a board after it's already been in use, they see nothing in the activity history — because Nextcloud's activity system only records events for users who had access at the time. Every past activity entry (card created, moved, edited, commented on, etc.) is invisible to them.

This adds retroactive backfilling: whenever a board is shared (with a user, group, or circle), we query the board owner's existing activity history from `oc_activity` and republish every event to each newly added user. Since the board owner has been present since creation, their feed is the authoritative complete record. The entries preserve original timestamps, authors, and subjects, so the history is identical for all members regardless of when they joined.

### What changed

- **ActivityManager** — new `retroactivelyPublishBoardActivities()` that queries `oc_activity` for all deck events belonging to the board (both board-level and card-level via `deck_stacks`), filtered to the board owner's entries, and republishes each one for the new user. Adds `IDBConnection` dependency.
- **PermissionService** — new `getUsersForAcl()` that resolves any ACL type (user / group / circle) to a concrete list of UIDs.
- **BoardService::addAcl()** — calls the backfill for every resolved UID after a share is created.

Note: There is no public OCP API to read back existing activity events (`IManager` only offers `publish()` / `generateEvent()`), so this queries `oc_activity` directly. The schema is stable Nextcloud core infrastructure, the query is read-only and well-scoped. This is a pragmatic solution — ideally the activity app would expose a query API upstream.

---

* Resolves: #7673
* Target version: main

### Checklist

### DIFF
--- a/lib/Activity/ActivityManager.php
+++ b/lib/Activity/ActivityManager.php
@@ -27,6 +27,8 @@ use OCP\Activity\IManager;
 use OCP\AppFramework\Db\DoesNotExistException;
 use OCP\AppFramework\Db\MultipleObjectsReturnedException;
 use OCP\Comments\IComment;
+use OCP\DB\QueryBuilder\IQueryBuilder;
+use OCP\IDBConnection;
 use OCP\IUser;
 use OCP\L10N\IFactory;
 use OCP\Server;
@@ -46,6 +48,7 @@ class ActivityManager {
 	private AclMapper $aclMapper;
 	private StackMapper $stackMapper;
 	private IFactory $l10nFactory;
+	private IDBConnection $db;
 
 	public const DECK_OBJECT_BOARD = 'deck_board';
 	public const DECK_OBJECT_CARD = 'deck_card';
@@ -101,6 +104,7 @@ class ActivityManager {
 		StackMapper $stackMapper,
 		AclMapper $aclMapper,
 		IFactory $l10nFactory,
+		IDBConnection $db,
 		?string $userId,
 	) {
 		$this->manager = $manager;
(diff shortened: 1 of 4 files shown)


grnd-alt (changes requested):
thanks for your contribution, the fix is highly appreciated I am not fond of the way it is implemented however. The created card activity event is far from the only one shown and republishing only those results in differing activity histories for users depending on when they joined the board. I think it's better to not have old history at all than only part of it.

That said I think an approach where the events are retrieved and directly republished no matter what type they are should also be possible.

ramteid:
@grnd-alt I considered your point. You might be correct about that. I tried to find an alternative by directly accessing the oc_activity schema.

The updated implementation no longer has this problem. Instead of synthetically reconstructing only card_create events, it queries the oc_activity table for the board owner's complete activity history and republishes every event - regardless of type - to the newly added member. Since the board owner has been present since creation, their feed is the authoritative record of everything that ever happened on the board. Every member receives an exact copy of that record, so the activity history is identical for all users no matter when they joined.

I acknowledge this is only a workaround to a missing functionality in the Activity app. I'll leave it up to you whether or not to include it in your codebase.

grnd-alt:
@ramteid thanks for updating the pr! I think what you mentioned is correct. The activity approach should be improved rather than doing sql queries directly on the activity table. So lets rather think about an implementation in the activity app, the main thing being that it's always filtered by affected user which can only be one user....
I will close this pr for now, in order to maybe get a feature in activity in.

PS: We have a Nextcloud Talk instance where we invite Nextcloud contributors to be able to communicate more quickly. If you like, I could give you a guest account to further discuss this and future prs. If you're interested I would need an email address from you though as your commits are only signed with the gh mail ;)

grnd-alt:
feature request and some thoughts in activity: https://github.com/nextcloud/activity/issues/2400

PR #7705, opened 2026-03-03
fix: user may be null

miaulalala:
* Resolves: # 
* Target version: main

### Summary
Found in my logs during tests:

```
"CustomMessage":"OC\\Group\\Manager::getUserGroupIds(): Argument #1 ($user) must be of type OCP\\IUser, null given, called in /var/www/html/apps-extra/deck/lib/Db/BoardMapper.php on line 108 in file '/var/www/html/lib/private/Group/Manager.php' line 352"
```

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/lib/Db/BoardMapper.php
+++ b/lib/Db/BoardMapper.php
@@ -105,8 +105,12 @@ class BoardMapper extends QBMapper implements IPermissionMapper {
 		));
 
 		// Shared to user groups of the user
-		$groupIds = $this->groupManager->getUserGroupIds($this->userManager->get($userId));
-		if (count($groupIds) !== 0) {
+		$user = $this->userManager->get($userId);
+		$groupIds = null;
+		if ($user !== null) {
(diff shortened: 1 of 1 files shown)


lib/Db/BoardMapper.php
```diff
@@ -105,8 +105,12 @@ public function findBoardIds(string $userId): array {
...
 
 		// Shared to user groups of the user
-		$groupIds = $this->groupManager->getUserGroupIds($this->userManager->get($userId));
-		if (count($groupIds) !== 0) {
+		$user = $this->userManager->get($userId);
+		$groupIds = null;
+		if ($user !== null) {
+			$groupIds = $this->groupManager->getUserGroupIds($user);
+		}
+		if ($groupIds !== null) {
```
grnd-alt: nitpick:

```suggestion
		if ($groupIds !== null && count($groupIds) !== 0) {
```

the user can still be in no groups so we can save/shorten the query

Issue #6919, opened 2025-04-17
Copy card does not copy content anymore


Gobytego:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Describe the bug**
Copying a card from one board to another only copys the name of the card and the people assigned... no details, no attachments, no comments.

**To Reproduce**
1. Go to 'triple dot'
2. Click on 'copy/move'
3. Click copy card
4. select a different board
5. goto said board
6. open card

**Expected behavior**
it should carry over all details and comments and attachments... i can understand not bringing over activity but everything else should copy. we use a lot of templates and have a template board... the copy feature has been awesome until just this week we noticed it stopped. guessing since last update.

**Client details:**
 - OS: Ubuntu 24.04.02
 - Browser firefox
 - Version 138
 - Device: desktop

<details>
<summary>Server details</summary>

**Operating system**: Ubuntu 24.04

**Nextcloud version:**Snap - [Hub 10] (31.0.2)

**Where did you install Nextcloud from:** snap store

**Signing status:** letsencrypt

```
Login as admin user into your Nextcloud and access
http://example.com/index.php/settings/integrity/failed
paste the results here.
```No errors have been found.

**List of activated apps:**

```Enabled:
  - activity: 4.0.0
  - admin_audit: 1.21.0
  - announcementcenter: 7.1.0
  - bruteforcesettings: 4.0.0
  - calendar: 5.2.1
  - cloud_federation_api: 1.14.0
  - comments: 1.21.0
  - contacts: 7.0.6
  - contactsinteraction: 1.12.0
  - dashboard: 7.11.0
  - dav: 1.33.0
  - deck: 1.15.0
  - encryption: 2.19.0
  - external: 6.0.2
  - federatedfilesharing: 1.21.0
  - federation: 1.21.0
  - files: 2.3.1
  - files_downloadlimit: 4.0.0
  - files_external: 1.23.0
  - files_pdfviewer: 4.0.0
  - files_reminders: 1.4.0
  - files_sharing: 1.23.1
  - files_trashbin: 1.21.0
  - files_versions: 1.24.0
  - firstrunwizard: 4.0.0
  - forms: 5.1.0
  - groupfolders: 19.0.4
[...]

PR #7714, opened 2026-03-05
fix(cards): correctly copy label and description when cloning cards

solracsf:
Fix https://github.com/nextcloud/deck/issues/6919
Target version: main

- [x] Tested OK on a patched v32.0.5 with app version 1.16.3

### Summary

When cloning a card with labels to a different board, `LabelService::cloneLabelIfNotExists()` returned the source board's label instead of the matching label found on the target board. This caused `assignLabel()` to fail with a 500 error: "Operation not allowed. Label does not exist."

Return `$filteredValues[0]` instead of `$originLabel` when a matching label already exists on the target board.

---

When cloning a card, the description was not copied to the new card. After `create()` inserts the new card via the ORM, the internal updatedFields tracker is reset. Calling `setDescription()` on the returned entity object was not properly tracked by the ORM, causing `cardMapper->update()` to skip persisting the description field.

Re-fetch the card from the database before setting the description, ensuring the ORM correctly tracks the field as updated before persisting.

### Checklist

### DIFF
--- a/tests/unit/Service/CardServiceTest.php
+++ b/tests/unit/Service/CardServiceTest.php
@@ -250,6 +250,8 @@ class CardServiceTest extends TestCase {
 		$card->setOrder(0);
 		$card->setOwner('admin');
 		$card->setStackId(12345);
+		$card->setDescription('A test description');
+
 		$clonedCard = clone $card;
 		$clonedCard->setId(2);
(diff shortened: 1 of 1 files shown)


solracsf:
Will backport manually

Issue #5852, opened 2024-05-06
Simultaneous editing of description


PatrickJosh:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Is your feature request related to a problem? Please describe.**
Currently, simultaneous editing of the description of cards by multiple people does not work, i.e. only the contents of the person who edited the description most recently are saved. This is contrary to the usual experience with the text editor. I was not expecting that and thus have already lost some contents due to this.

**Describe the solution you'd like**
Live/simultaneous editing of the description by multiple people should be possible.

**Describe alternatives you've considered**
A notice that simultaneous editing is not possible since currently, people will probably expect that simultaneous editing is possible since it is in other locations where the text editor is used. Moreover, in this case, a warning that another person is editing the description should be implemented, in my opinion.

**Additional context**
#256 might be related, but it seems to be more applicable to the cards and boards themselves and not their description.

SimJoSt:
It would be great to have at least a similar feature to file-locking, to prevent overwriting.

antoninchadima:
For which release is the file-locking similar feature planned? Didn't see it on the Office team planning board.

PR #7716, opened 2026-03-06
feat: live update card description

luka-nextcloud:
* Resolves: #5852
* Target version: main

### Checklist


juliusknorr:
Nice one, may just be a first step as it still has potential for overwriting other peoples changes. Long term it may be a good idea to extend the text y.js based collaboration to also work with non-file content.
e
Please still assign some labels, they make it easier to find changes later on and may help with aggregating changelogs.

Issue #7279, opened 2025-09-28
Drag cards by dragging them over the title text instead of selecting it.


Somebodyisnobody:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Is your feature request related to a problem? Please describe.**
It is really annoying that I can drag cards only in specific areas. I don't see the point of being able to select the card title text in the overview.

**Describe the solution you'd like**
I want to be able to drag a card independent from where I click on the card (except for buttons, of course). I know there is the closed bug #5378 of a user complaining that he cannot select the title text of a card and it get's dragged instead. But why don't open the card for selecting the title text? Moreover closing the mentioned bug (option to edit title without opening the card was removed) caused another problem described in #7265.

**Describe alternatives you've considered**
Preferrred alternative:
* Set css `user-select: none` for the card title text
* Make the card draggable in the whole purple marked area
<kbd>
<img width="279" height="49" alt="Image" src="https://github.com/user-attachments/assets/c0832edd-51f5-4b13-ad24-53a9f1dc479b" />
</kbd>

Not preferred but "i-can-live-with"-alternative:
* set css `cursor: pointer` for the purple marked area in the image except the title text
* set css `cursor: text` for the title text

**Additional context**
none

Somebodyisnobody:
@grnd-alt sorry for poking you, but since you edited the linked issue four days ago, I think you still have the context in mind and can evaluate my feature request better than someone who will look at it in two months.

Somebodyisnobody:
See also how long card titles affect the draggable area.

<img width="372" height="164" alt="Image" src="https://github.com/user-attachments/assets/11e970ef-8f85-45b5-bd36-0081bd0f9e6d" />

Removing the css class "dragDisabled" for the element would make the card draggable and non-selectable: https://github.com/nextcloud/deck/blob/90f51f931f56834b1892f5c644a736b3fadcfa0e/src/components/cards/CardItem.vue#L23

luka-nextcloud:
I can reproduce the issue.

Somebodyisnobody:
Workaround: This problem is soo annoying that I wrote a greasemonkey script as workaround:
```js
// ==UserScript==
// @name     Deck-Drag
[...]

Issue #7514, opened 2026-01-09
UX: Accidental Text Selection and Inconsistent De-selection


Turbulentarius:
When using Deck and moving cards around, I frequently end up selecting text instead of dragging the card. This happens almost every time I use Deck and is surprisingly disruptive, as it creates uncertainty about whether I might accidentally trigger or break something while trying to clear the selection.

The issue occurs in several places and is not unique to Nextcloud Deck. However, in most interfaces, text selection is cleared by clicking on a neutral, non-interactive area (for example, the white background of a page). In Nextcloud, this behavior does not occur. I suspect this may be related to specific combinations of CSS layout properties, although I can’t recall the exact cause.

I do appreciate being able to select text, as I often copy and paste content and rely on it due to poor spelling. Completely disabling text selection would therefore be undesirable. That said, it would be very helpful if clicking on a neutral background area simply cleared the text selection.

<img width="560" height="364" alt="Image" src="https://github.com/user-attachments/assets/8ea88729-f1cc-4476-a849-e8276815e5c0" />

**Platform:** Windows 11
**Browser:** Firefox 146.0.1 (64-bit)

PR #7765, opened 2026-03-19
Remove `dragDisabled` class to prevent accidential text selection when dragging a card

theoholl:
* Resolves: #7279 #7514
* Target version: main

### Summary

The card title could be selected with the cursor, which often caused accidental text selection when trying to drag a card. This PR removes the `dragDisabled` class from the card title so that card is dragged instead of text selected.

Especially small cards with only title text and little 'draggable padding' suffered from the text selection issue (see screencast below).

Before:

https://github.com/user-attachments/assets/b4398024-d520-4281-8d5d-4c41889f1cc7

After:

https://github.com/user-attachments/assets/bbcd0db2-a5b4-4f0f-a76b-964af4e0d592

### DIFF
--- a/src/components/cards/CardItem.vue
+++ b/src/components/cards/CardItem.vue
@@ -20,7 +20,7 @@
 			<CardCover v-if="showCardCover" :card-id="card.id" />
 			<div class="card-upper">
 				<h4 v-if="editingTitle === 0" key="title-view" dir="auto">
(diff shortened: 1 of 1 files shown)


lemoer:
Thank you for opening this pull request! I feel that this is a hughe improvement for the product. I often see coworkers struggling with this and it causes distraction in our meetings.

I hope someone will review and merge it soon!

PR #7787, opened 2026-03-24
feat: add color property to cards and implement color selection feature

### LINKED ISSUE #3592: Set individual background color of card
In order to add yet another dimension for sorting cards, it would be nice if cards could have an adjustable background color - similar to be able to chose from different post-it colors.
Here is a quick mockup:
image

### How to use GitHub

**Additional context**

luka-nextcloud:
* Resolves: #3592
* Target version: main

### Summary
* Added a new `color` column to the `deck_cards` table.
* Updated the `Card` entity and service logic to handle the new `color` property.

### Checklist


grnd-alt (changes requested):
code changed is valid, great work!

I don't really like the layout though, I don't think the color picker should take up an entire row while being so small in width. Maybe we could put it in line with the due date selector or move into the three dot menu? maybe @marcoambrosini has a great idea?

just leaving some screenshots here that look weird because of themes:

text gets invisible if the card is the text color for the theme, might be hard to solve but as we support both light and dark mode it should always be readable for everyone.
Image

the default color for the picker also makes it hard to see in light mode:
Image

luka-nextcloud:
@marcoambrosini @grnd-alt Please check this video:
[demo.webm](https://github.com/user-attachments/assets/1581aa0c-8988-4524-9285-f76a3d44f2d6)

marcoambrosini (commented):
Nice @luka-nextcloud, thanks for the screencast

A couple of changes I would make:

* move the button to the left of the card action menu
Image

* Make the button square, size: var(--default-clickable-area)
* Add a tooltip to the button: "Choose card color"

luka-nextcloud:
@marcoambrosini @grnd-alt We are using component `NcAppSidebar` to display card. And this component does not provide slot to show the button on the left of actions menu. However, there is a slot `tertiary-actions` which can display custom html on the left of card title. What do you think?
image

marcoambrosini:
I would then add this as an action within the action menu itself, what do you think @luka-nextcloud ?

luka-nextcloud:
@marcoambrosini Do you mean like this? It looks ok to me
image
image

marcoambrosini:
Very nice, yes!

marcoambrosini (approved):

grnd-alt (approved):

PR #7812, opened 2026-03-31
feat: add Gantt chart view for boards

### LINKED ISSUE #134: Add diagramme gantt
Hello
Is it possible to have an output in the form of a Gantt chart
Thank you

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/45278406-add-diagramme-gantt?utm_campaign=plugin&utm_content=tracker%2F52758431&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F52758431&utm_medium=issues&utm_source=github).
</bountysource-plugin>

aparcar:
### Summary

Add a timeline/Gantt view as an alternative to the kanban board view, allowing users to visualize card schedules across time using frappe-gantt.

- Add GanttView component with Day/Week/Month view modes
- Add Kanban/Gantt view toggle in board controls
- Store view mode preference in localStorage via Vuex
- Stack-based color coding with legend and undated cards section
- Drag-and-drop support for rescheduling cards
- Auto-fit column width to fill container on wider views
- Add frappe-gantt dependency and webpack resolve alias for its CSS

image

* Resolves: #134 (an old one!)
* Target version: main

This is inspired by https://github.com/nextcloud-community/ncgantt

### TODO

- [ ] Sometimes entries "jump" around while dragging, however this might be an issue with my browser

### Checklist

### DIFF
--- a/src/components/board/GanttView.vue
+++ b/src/components/board/GanttView.vue
@@ -1,0 +1,514 @@
+<!--
+  - SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+<template>
+	<div class="gantt-wrapper">
+		<div class="gantt-toolbar">
+			<NcButton v-for="mode in viewModes"
+				:key="mode.value"
+				:type="currentViewMode === mode.value ? 'primary' : 'secondary'"
+				@click="changeViewMode(mode.value)">
+				{{ mode.label }}
+			</NcButton>
+		</div>
+
+		<div v-if="ganttTasks.length > 0" ref="ganttContainer" class="gantt-chart" />
+
+		<NcEmptyContent v-if="!ganttTasks.length && !undatedCards.length">
+			<template #icon>
+				<ChartGanttIcon />
+			</template>
+			<template #name>
+				{{ t('deck', 'No cards with dates') }}
+			</template>
+			<template #description>
+				{{ t('deck', 'Set a start date and due date on your cards to see them on the Gantt chart') }}
+			</template>
+		</NcEmptyContent>
+
+		<div v-if="undatedCards.length" class="gantt-undated">
+			<NcButton type="tertiary" @click="showUndated = !showUndated">
+				<template #icon>
+					<ChevronDown v-if="showUndated" :size="20" />
+					<ChevronRight v-else :size="20" />
+				</template>
+				{{ t('deck', 'Cards without dates ({count})', { count: undatedCards.length }) }}
+			</NcButton>
+			<div v-if="showUndated" class="gantt-undated__list">
+				<div v-for="card in undatedCards"
+					:key="card.id"
+					class="gantt-undated__card"
+					@click="openCard(card)">
+					<span class="gantt-undated__stack-dot" :style="{ backgroundColor: getStackColor(card.stackId) }" />
+					<span class="gantt-undated__title">{{ card.title }}</span>
+					<span class="gantt-undated__stack-name">{{ getStackTitle(card.stackId) }}</span>
+				</div>
+			</div>
+		</div>
+
+		<div v-if="stacksByBoard.length" class="gantt-legend">
+			<span v-for="stack in stacksByBoard"
+				:key="stack.id"
+				class="gantt-legend__item">
+				<span class="gantt-legend__dot" :style="{ backgroundColor: getStackColor(stack.id) }" />
+				{{ stack.title }}
+			</span>
+		</div>
+	</div>
+</template>
+
+<script>
+import { mapState } from 'vuex'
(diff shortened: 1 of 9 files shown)


grnd-alt (dismissed):
works thanks a lot for this contribution. But there is some room for improvement. also darkmode looks like this:

Image

aparcar:
@grnd-alt I ran Claude again and worked on the fixes. Since this is mostly AI generating (somewhat usable) things, I understand that this quickly starts to waste your time. In other words, please feel free to close this PR until someone comes along with actual web dev knowledge to create a proper implementation.

src/components/board/GanttView.vue
```diff
@@ -0,0 +1,514 @@
...
+				view_mode: this.currentViewMode,
+				view_modes: GANTT_VIEW_MODES,
+				bar_height: 28,
+				lower_header_height: 40,
+				padding: 20,
+				scroll_to: 'today',
+				today_button: true,
+				infinite_padding: false,
+				on_click: (task) => {
+					this.openCard(task._card)
```
grnd-alt:
```suggestion
					if (!this.isDragging) {
						this.openCard(task._card)
					}
```

src/store/main.js
```diff
@@ -62,6 +62,7 @@ export default function storeFactory() {
 			activityLoadMore: true,
 			filter: { tags: [], users: [], due: '', unassigned: false, completed: 'both' },
 			shortcutLock: false,
+			viewMode: localStorage.getItem('deck.viewMode') || 'kanban',
```
mejo-: This means that the view mode is switched persistently for all boards, is this the expected behaviour? Especially as the gantt view doesn't allow to edit the board (i.e. add cards/lists), just to edit the cards themselves.

src/components/board/GanttView.vue
```diff
@@ -0,0 +1,562 @@
...
+				:key="mode.value"
+				:type="currentViewMode === mode.value ? 'primary' : 'secondary'"
+				@click="changeViewMode(mode.value)">
+				{{ mode.label }}
+			</NcButton>
+		</div>
+
+		<div v-if="ganttTasks.length > 0" ref="ganttContainer" class="gantt-chart" />
+
+		<NcEmptyContent v-if="!ganttTasks.length && !undatedCards.length">
```
mejo-, line 18: For an empty board, it might be better to switch back to the kanban view, no? When I create a list using this interface, I cannot really do anything with it afterwards, as the gantt view doesn't allow to add another list or cards anyway.

src/components/board/GanttView.vue
```diff
@@ -0,0 +1,562 @@
...
+				infinite_padding: false,
+				readonly_progress: true,
+				popup: false,
+				on_click: (task) => {
+					if (!this.isDragging) {
+						this.openCard(task._card)
+					}
+				},
+				on_date_change: (task, start, end) => {
+					this.isDragging = true
```
mejo-: The whole `isDragging` and `_pendingChange` implementation looks a bit overly complex to me. Maybe it's required that way, but since the PR was initially generated by AI, I wonder whether it really is. `on_date_change` seems to be a callback from the gantt library used here. Why doesn't it work to make it call `onDateChange()` directly? Why the need to track `isDragging` and pending changes in Vue explicitly?

mejo- (commented):
Overall a great addition ❤️ 

I tested a bit, skimmed through the code changes and made some comments.

src/components/board/GanttView.vue
aparcar: I personally would drop all dragging functionality and just allow the user to click a task and then manually modify the specific dates.

grnd-alt: @mejo- I refactored that logic to only have pendingChange, that is required so we only update when the dragging ends and not in between all the time. The logic is now in the global mouseUp listener as the library's on_click seems to not be in the same event chain as the global mouseUp/click so we can't stop propagation between them, and we would have ending drags open the modal, which is quite annoying when you only want to change the date.

aparcar:
Discussing this with colleagues, they asked for direct links to the gantt or deck view. Could this be added to the url handling?

grnd-alt:
@aparcar we can think about it as a follow up, but I'd prefer to get this pr done for now so we can have it ready in 34.

mejo- (approved):
Tested again, and works well.

As written in chat, moving a card on the Gantt diagram results in two `PUT .../api/v1.0/cards/<cardId>` requests, the first with `startdate: null` and the second with a value in `startdate`.

I guess there's a flaw somewhere and the first request should not be sent.

Apart from that it looks good. Approving already so you don't need to ping me again after fixing this final issue 😉

aparcar:
Cool! Thanks for making this happen

Issue #6663, opened 2025-01-16
Migrate to Vue 3


juliusknorr:
Steps might need to be refined:

- [ ] Come up with an overview of dependencies and their vue3 compatibility - which ones we can keep, which are incompatible, needs replacement
- [ ] Migration steps that are possible to do before upgrading
- [ ] do another stable release and branch off
- [ ] Upgrade packages to vue3 and do changes that have not been possible before

Further reference
- https://github.com/nextcloud/standards/issues/13
- https://v3-migration.vuejs.org/
- https://github.com/nextcloud-gmbh/office-team/issues/47

juliusknorr:
Added a PR list to the first post @grnd-alt

grnd-alt:
# How to migrate
**possible before changing to vue3**
* [ ] migrate all stores away from vuex to pinia (see #7790 for example and plan)
* [ ] migrate to nextcloud-eslint-config@9

**do after installing vue3**
* [ ] migrate <var.sync> to v-model:<var>
* [ ] migrate away from @<event>.native
* [ ] list emited events of components in emits
* [ ] migrate beforeDestroy -> beforeUnmount
* [ ] migrate away from propsData
* [ ] Audit codebase for $listeners, $children, $scopedSlots usage
* [ ] Audit for filters usage
* [ ] Find/build replacements for Vue 2-only deps below

## Dependencies
### Vue 2-Only Dependencies

| Package | Issue | Replacement |
|---------|-------|-------------|
| `vue-at` | Vue 2 only | Find Vue 3 alternaitve |
| `vue-click-outside` | Vue 2 only | Use `@vueuse/core` `onClickOutside` |
| `vue-easymde` | Vue 2 only | `vue3-easymde` or integrate EasyMDE directly |
| `vue-infinite-loading` | Abandoned, Vue 2 | `@vueuse/core` `useInfiniteScroll` |
| `vue-smooth-dnd` | Vue 2 only | `vue3-smooth-dnd` or `vuedraggable@next` |
| `vuex-router-sync` | Vuex only | Not needed with pinia (use composable) |


### Version Bumps Required

| Package | Current | Vue 3 Version |
|---------|---------|---------------|
| `vue` | ^2.7.15 | ^3.x |
| `vue-router` | ^3.6.5 | ^4.x |
| `vuex` | ^3.6.2 | use Pinia |
| `vue-template-compiler` | ^2.7.16 | Remove → `@vue/compiler-sfc` |
[...]

PR #7827, opened 2026-04-06
Migrate `comment.js` Vuex store to Pinia

theoholl:
* Related to: #6663
* Target version: main

### Summary

This PR migrated the `comment.js` Vuex store to Pinia, following patterns from https://github.com/nextcloud/deck/pull/7790.


grnd-alt (changes requested):
did you test this? this breaks reactivity when deleting comments and when loading more comments than on initial fetch

theoholl:
Oh shoot, I'll review that. I thought I had tested deleting (adding, referencing, etc.) comments at least but I'll put this PR on hold until I tested more thoroughly.

src/stores/comment.js
```diff
@@ -0,0 +1,104 @@
...
+		async fetchMore({ cardId }) {
+			// fetch newer comments first
+			await this.fetchComments({ cardId })
+			await this.fetchComments({ cardId, offset: this.getCommentsForCard(cardId).length })
+		},
+		async createComment({ cardId, comment }) {
+			await apiClient.createComment({ cardId, comment, replyTo: this.replyTo })
+			await this.fetchComments({ cardId })
+		},
+		async apiDeleteComment(data) {
```
grnd-alt: I don't see a need for these api<Command> wrappers, better just merge that into one action

src/components/card/CardSidebarTabComments.vue
```diff
@@ -75,12 +80,16 @@ export default {
...
 		...mapState({
 			currentBoard: state => state.currentBoard,
-			replyTo: state => state.comment.replyTo,
 		}),
-		...mapGetters([
-			'getCommentsForCard',
-			'hasMoreComments',
-		]),
+		replyTo() {
+			return this.commentStore.replyTo
```
grnd-alt: we can just use commentStore.replyTo, no need to have these extra computeds

grnd-alt (changes requested):
@theoholl took me awhile to get to this, but here I am now!
this works great now, I have some small remarks but then we can get the next store to pinia 🥳

src/stores/comment.js
theoholl: Yes, currently there is no need to just add/remove/update comments from the state store without updating via API, so I merged those.

src/components/card/CardSidebarTabComments.vue
theoholl: Makes sense to me.

grnd-alt (approved):

PR #7857, opened 2026-04-16
fix(teams): use lighter probeCircles()

ArtificialOwl:
it is advice to use `probeCircles()` for this purpose, the backend will request a structured/cached version of data about memberships,
it also extract less data from database and provide a way to request extra data. (`DataProbe`)


artonge (approved):

luka-nextcloud (changes requested):
- The static analysis issue can be fixed by stubbing method `probeCircles` in `tests/stub.phpstub`
- The performance warning can be ignored since it is unrelated

luka-nextcloud (approved):

PR #7870, opened 2026-04-20
feature add deck to user_migration

### LINKED ISSUE #3828: User migration (export/import data via user migration app)
With 24 we have a server API to handle user migration which deck should also support by implementing https://docs.nextcloud.com/server/stable/developer_manual/digging_deeper/user_migration.html

Related App: https://apps.nextcloud.com/apps/user_migration

We already have a way to dump json exports of boards which could be reused, but it might be missing some data:
- Comments
- Activity
- File shares

Especially with file shares it might be a topic for further discussion on how to handle those on user migration. The files might not be owned by the board owner and therefore would not be migrated. Also sharing migration is currently not implemented for files in general.

samin-z:
* Resolves: #3828 
* Target version: main

### Checklist

### DIFF
--- a/lib/Service/Importer/Systems/DeckJsonService.php
+++ b/lib/Service/Importer/Systems/DeckJsonService.php
@@ -156,7 +156,7 @@ class DeckJsonService extends ABoardImportService {
 	/**
 	 * @return array{0: string, 1: string}
 	 */
-	private function resolveCommentActor(\stdClass $commentOriginal): array {
+	private function resolveCommentActor(object $commentOriginal): array {
 		$actorType = (string)($commentOriginal->actorType ?? 'users');
 		$actorId = $commentOriginal->actorId ?? null;
 
@@ -171,7 +171,7 @@ class DeckJsonService extends ABoardImportService {
 		return [$actorType, $actorId];
 	}
 
-	private function importAttachmentsForCard(\stdClass $sourceCard): void {
+	private function importAttachmentsForCard(object $sourceCard): void {
 		if (!property_exists($sourceCard, 'attachments') || !isset($this->cards[$sourceCard->id])) {
 			return;
 		}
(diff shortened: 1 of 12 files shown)


lib/Service/Importer/Systems/DeckJsonService.php
```diff
@@ -112,17 +114,121 @@ public function getComments(): array {
...
+							->setParentId('0')
+							->setMessage($commentOriginal->message)
+							->setCreationDateTime(\DateTime::createFromFormat('Y-m-d\TH:i:sP', $commentOriginal->creationDateTime));
+						$comment->setMetaData([
+							'deckImportSourceId' => $commentId,
+							'deckImportParentId' => '0',
+						]);
+						$ordered[$commentId] = $comment;
+					}
+					$pending = [];
```
grnd-alt: nitpick: `break;` would be more readable imo with the same effect.

lib/UserMigration/DeckMigrator.php
```diff
@@ -0,0 +1,261 @@
...
+			}
+			$boardWithStacksAndCards = $this->boardService->export($board->getId());
+			$this->appendArchivedCards($boardWithStacksAndCards);
+			$exportData['boards'][] = $this->serializeBoard($boardWithStacksAndCards, $uid);
+		}
+
+		return $exportData;
+	}
+
+	private function serializeBoard(object $board, string $uid): array {
```
grnd-alt, line 140: I assume this serializeBoard path, mainly exists to get comments into cards right?

lib/UserMigration/DeckMigrator.php
```diff
@@ -0,0 +1,261 @@
...
+				'objectId' => $comment->getObjectId(),
+				'verb' => $comment->getVerb(),
+			];
+		}
+
+		return $formattedComments;
+	}
+
+	private function shouldImport(IImportSource $importSource): bool {
+		return $importSource->getMigratorVersion($this->getId()) !== null;
```
grnd-alt, line 218: any version is fine?

grnd-alt (approved):
two small questions, but nothing blocking. Nice 🥳

lib/UserMigration/DeckMigrator.php
samin-z: yes

lib/UserMigration/DeckMigrator.php
samin-z: yes, its just to check there is a version

grnd-alt (approved):

PR #7875, opened 2026-04-22
feat: Add swimlane grouping by labels and assignees

### LINKED ISSUE #32: Swimlanes support
Would be nice to be able to split tasks in swimlanes.
A swimlane is a horizontal categorisation of cards in the active Kanban board on a board.

Pavlinchen:
* Target version: NC34 +
* Resolves: #32
  * perhaps also relates to: #1878 

#### Disclaimer
This PR was written mainly with LLM assistance (Claude Opus 4.7 max effort) for technical implementation.

### Summary

Adds horizontal **swimlane grouping** to the board view.

Users can now group cards by **labels** or **assignees** via the existing "View Modes" menu. Each group renders as a collapsible horizontal row with its own set of stacks, showing only the cards that belong to that group.

**Key behaviors:**
- Cards with multiple labels (or multiple assignees) appear in **all matching lanes** - a subtle badge indicates the card is in multiple lanes
- A "No label" / "Unassigned" catchall lane appears at the bottom for ungrouped cards
- Empty lanes (labels/assignees with no cards) are automatically hidden
- Lanes can be **collapsed/expanded** (persisted per-user in localStorage)
- Lanes can be **reordered via drag-and-drop** (editors only, persisted via OCS config API as shared board setting)
- Card drag-and-drop works **within each lane** across stacks, but cross-lane dragging is blocked (prevents accidental label/assignee changes)
- Swimlane mode and lane order are **shared board settings** - all users see the same layout, only editors can change them (persisted via `setAppValue`, gated by `PERMISSION_EDIT`)
- Full **keyboard and screen reader accessibility** - `aria-expanded`, `aria-label`, focus-visible styles

**No database changes** - swimlanes are a computed view layer derived from existing card labels and assignee data.

### Architecture

The implementation adds **no new API endpoints or database tables**. It extends the existing frontend state management and adds two new Vue components:

| Layer | Changes |
|-------|---------|
| **Vue components** | New `Swimlane.vue` and `SwimlaneHeader.vue`; modified `Board.vue`, `Stack.vue`, `Controls.vue`, `CardBadges.vue` |
| **Vuex store** | New `cardsByStackAndLane` getter in `card.js`; swimlane order state + actions in `main.js` |
| **Backend** | Extended `BoardService::enrichWithBoardSettings()` to include swimlane config (3 new `getAppValue` calls); `ConfigService::set()` gates swimlane keys with `PERMISSION_EDIT` |
| **Persistence** | Swimlane mode and lane order stored as shared board settings via `setAppValue` (visible to all users, editable by board editors only). Collapse state remains per-user in localStorage |

The existing `vue-smooth-dnd` Container/Draggable pattern is reused for both within-lane card drag and lane reordering. Lane isolation uses `group-name` scoping (`'stack-' + lane.key`) to prevent cross-lane card drops.

### How it works

1. **Controls.vue** adds "Group by" radio options to the View Modes menu
2. When a mode is selected, it is stored via the OCS config API as a shared board setting and an event-bus message triggers Board.vue to re-render
3. **Board.vue** switches from the flat stack view to a vertical list of `Swimlane` components
4. Each `Swimlane` renders a `SwimlaneHeader` (with label color/avatar, card count, collapse toggle) and the same horizontal stack layout
5. **Stack.vue** accepts an optional `lane` prop - when present, it uses `cardsByStackAndLane` to filter cards for that lane
6. **CardBadges.vue** shows a duplicate badge when a card appears in 2+ lanes

### Changes

**New files (4):**
- `src/components/board/Swimlane.vue` - Swimlane row wrapper (header + stacks)
- `src/components/board/SwimlaneHeader.vue` - Lane header with label/avatar, count, collapse, drag handle
- `tests/unit/Service/ConfigServiceTest.php` - PHPUnit tests for swimlane config storage and permission gating
- `cypress/e2e/swimlanesFeatures.js` - Cypress E2E tests for swimlane grouping interactions

**Modified files (8):**
- `src/components/board/Board.vue` - Conditional flat vs swimlane rendering, lane computation, lane reorder DnD (editor-gated)
- `src/components/board/Stack.vue` - Accept `lane` prop, lane-scoped card filtering and DnD group names
- `src/components/Controls.vue` - "Group by" radio selector in View Modes menu (disabled for non-editors)
- `src/components/cards/CardBadges.vue` - Duplicate lane badge
- `src/store/card.js` - `cardsByStackAndLane` getter
- `src/store/main.js` - Swimlane mode/order state, mutations, actions, config loading
- `lib/Service/BoardService.php` - Swimlane settings in board enrichment (shared via `getAppValue`)
- `lib/Service/ConfigService.php` - Swimlane config keys gated by `PERMISSION_EDIT`, stored via `setAppValue`

### Checklist

- [x] Verify with NC dark mode / high-contrast themes
- [x] Test with right-to-left (RTL) languages

### DIFF
--- a/lib/Service/ConfigService.php
+++ b/lib/Service/ConfigService.php
@@ -24,6 +24,8 @@ class ConfigService {
 	public const SETTING_BOARD_NOTIFICATION_DUE_ALL = 'all';
 	public const SETTING_BOARD_NOTIFICATION_DUE_DEFAULT = self::SETTING_BOARD_NOTIFICATION_DUE_ASSIGNED;
 
+	private const SWIMLANE_CONFIG_KEYS = ['swimlaneMode', 'swimlaneLabelOrder', 'swimlaneUserOrder'];
+
 	private ?string $userId = null;
 
 	public function __construct(
@@ -187,14 +189,61 @@ class ConfigService {
 				if (count($parts) < 3) {
 					break;
 				}
+				$boardId = $parts[1];
 				$boardConfigKey = $parts[2];
+				if ($boardId === '' || $boardConfigKey === '') {
+					throw new BadRequestException('Board config key must be in the form board:<id>:<setting>');
+				}
 				if ($boardConfigKey === 'notify-due' && !in_array($value, [self::SETTING_BOARD_NOTIFICATION_DUE_ALL, self::SETTING_BOARD_NOTIFICATION_DUE_ASSIGNED, self::SETTING_BOARD_NOTIFICATION_DUE_OFF], true)) {
 					throw new BadRequestException('Board notification option must be one of: off, assigned, all');
 				}
-				$this->config->setUserValue($userId, Application::APP_ID, $key, (string)$value);
+				if (in_array($boardConfigKey, self::SWIMLANE_CONFIG_KEYS, true)) {
+					// Shared board-level setting (visible to every board member).
+					// EDIT permission is enforced in ConfigController::setValue(),
+					// whose board:(\d+): match requires a numeric board id — so a
+					// numeric id is required here too, otherwise the permission
+					// check could be bypassed.
+					if (!ctype_digit($boardId)) {
+						throw new BadRequestException('Board id must be numeric');
+					}
+					$this->validateSwimlaneValue($boardConfigKey, $value);
+					$this->config->setAppValue(Application::APP_ID, $key, (string)$value);
+				} else {
+					$this->config->setUserValue($userId, Application::APP_ID, $key, (string)$value);
+				}
 				$result = $value;
 		}
 		return $result;
+	}
+
+	/**
+	 * Shared swimlane settings live in app config and are delivered to every
+	 * board member on board load, so reject malformed values before storing.
+	 */
+	private function validateSwimlaneValue(string $boardConfigKey, mixed $value): void {
+		if ($boardConfigKey === 'swimlaneMode') {
+			if (!in_array($value, ['none', 'labels', 'assignees'], true)) {
+				throw new BadRequestException('Swimlane mode must be one of: none, labels, assignees');
+			}
+			return;
+		}
+
+		// swimlaneLabelOrder / swimlaneUserOrder: JSON array of lane ids
+		// (label ids, user ids or the '__none__' catch-all lane).
+		if (!is_string($value) || strlen($value) > 8192) {
+			throw new BadRequestException('Swimlane order must be a JSON array of lane ids');
+		}
+		$order = json_decode($value, true);
+		if (!is_array($order) || !array_is_list($order) || count($order) > 1000) {
+			throw new BadRequestException('Swimlane order must be a JSON array of lane ids');
+		}
+		foreach ($order as $laneId) {
+			$isIntId = is_int($laneId);
+			$isStringId = is_string($laneId) && $laneId !== '' && strlen($laneId) <= 64;
+			if (!$isIntId && !$isStringId) {
+				throw new BadRequestException('Swimlane order entries must be label ids, user ids or "__none__"');
+			}
+		}
 	}
 
 	/**
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -42,7 +42,7 @@
         <database min-version="9.4">pgsql</database>
         <database>sqlite</database>
         <database min-version="8.0">mysql</database>
-        <nextcloud min-version="35" max-version="35"/>
+        <nextcloud min-version="34" max-version="35"/>
     </dependencies>
     <background-jobs>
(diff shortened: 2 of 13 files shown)


Pavlinchen:
Force pushed new (rebased) version to incorporate GanttView

Pavlinchen:
on it!

lib/Service/ConfigService.php
```diff
@@ -25,11 +27,15 @@ class ConfigService {
...
 	public const SETTING_BOARD_NOTIFICATION_DUE_DEFAULT = self::SETTING_BOARD_NOTIFICATION_DUE_ASSIGNED;
 
+	private const SWIMLANE_CONFIG_KEYS = ['swimlaneMode', 'swimlaneLabelOrder', 'swimlaneUserOrder'];
+
 	private ?string $userId = null;
 
 	public function __construct(
 		private readonly IConfig $config,
 		private readonly IGroupManager $groupManager,
+		private readonly PermissionService $permissionService,
```
grnd-alt: that's a circular dependency and breaks the app

grnd-alt (changes requested):
Hey, thanks for your pr and the much appreciated work you did here.
Did not review the code fully but wanted to leave some comments from testing:

1. there's a lot of layout shift when cards are moved between stacks as the column width changes.
2. the order of lanes randomly changes when moving in columns in one lane
3. the "order by" setting is persisted between site refreshes but the settings radio button's don't reflect it after refresh
4. more of an opinion: I don't know if it's really necessary to have the column names listed for every lane, and them having different widths per lane seems counter-intuitive as well (but that's more of a design question idk what's better here)

Pavlinchen:
Hello again & thanks for the patience. 
I realize this isn't a trivial commit, so the table below summarises what changed, how, and why. 
If anything is needed, please let me know :)

| Before | Why changed | After |
|---|---|---|
| `PermissionService` injected into `ConfigService` | circular DI, broke the app | EDIT check moved to `ConfigController::setValue()` per upstream #7990 pattern |
| Stack column widths shrank to card content when cards moved between stacks | layout shift | fixed flex basis on stack columns: `flex: 0 1 $card-max-width; min-width: $card-min-width` |
| Lane order changed randomly on every card move (unsorted card-derived iteration) | random lane order | `computedLanes` ends with deterministic `.sort()` - saved order first, ID fallback |
| `NcActionRadio` used deprecated `:checked` Boolean - didn't reflect persisted mode after reload | radio state not persisting | migrated to `:model-value` / `@update:model-value`, reads from `board.settings.swimlaneMode` |
| Column headers repeated per lane with mismatched widths | removing unneccessary duplicates | single sticky header row; lanes are bare card columns sharing one horizontal scroll context |
| `board:<id>:` with empty key/id silently accepted; arbitrary strings could land in shared app config | self-review | empty-key + non-numeric-id guards + JSON-array shape validation in `ConfigService::validateSwimlaneValue` |
| Cypress lane selector interpolated raw labels into a CSS selector | self-review | `expectCardInLane()` helper via `cy.contains().parents('.swimlane')` |
| `setSwimlaneMode` / `setSwimlaneOrder` committed locally before awaiting the server - UI stayed mutated on reject | self-review | snapshot previous value, revert on dispatch failure |
| `user.participant.uid` deref crashed the whole board view on deleted/federated accounts | self-review | `?.participant?.uid` + skip falsy in `computedLanes` and `cardsByStackAndLane` |

Pavlinchen:
Hi @grnd-alt, following up here.

The June revision (20dbf4848) covers the points from your May review. 
Main has moved on quite a bit since then, so before I put another rebase together I would rather ask than guess.

Is this PR something this project wants? 
A no is a perfectly fine answer and I would rather hear it than keep the PR alive. 

If the answer is yes: which base should I target? 
I can see you've started work on some of the same files in #8259.
The one difference compared to main is `useSettingsStore().setConfig()`. 

So I can rebase onto main as it stands and or stack on top of #8259 - lmk


Thanks in advance &
kind regards,
Paul

grnd-alt:
@Pavlinchen I must say, I personally do not see a need for swimlanes to be honest. But it seems to be somewhat wanted by the community as there's a 10year old issue with some likes about it. I am open to have it in, but I do see the risk of increased complexity and maintenance cost, especially since we have the gantt chart which is in a not so good state performance whise right now.

So if you are motivated to have it in and think you'll get it in a very good state that's easily maintainable, I would say go on!

In regards to the rebase, I am currently mainly working on a refactor to move to vue3: https://github.com/nextcloud/deck/pull/8276

The pr you mentioned is incorporated into the vue3 rewrite I would advise you to wait until that is merged.

Pavlinchen:
Hi @grnd-alt,
sounds fair enough - for now I will wait on the merge of #8276 as discussed :)

Thanks for the fast reply

PR #7879, opened 2026-04-23
feat: add card dependencies

### LINKED ISSUE #5346: Dependencies / Subtasks for cards
Hey there,

if got in touch with Deck today and I really like it! So I am going to use it alot in future.
I don't know if you think the same but I would suggest to add the possibility for dependencies between cards.

As an example:

Bake Pizza is in column "Tasks"
Buy flour is in column "Tasks"
Buy yeast is in column "Tasks"

Now I choosed within the card Bake Pizza that before I can finished (move) the task its required to have "Buy Flour" and "Buy yeast" are at least in column Done.

Or you could do that by defining subtasks for a card like in Trello.

I would really apreciate if you consider to implement something like this.

Greetings
cos

### How to use GitHub

**Additional context**

luka-nextcloud:
* Resolves: #5346
* Target version: main

### Summary
Introduces the concept of "dependent cards", allowing users to assign and remove dependencies between cards. The implementation includes backend support, API routes, database migration, and frontend UI integration for managing card dependencies.

### Checklist


DesertCookie:
Awesome to see some movement on this.

luka-nextcloud:
Recording video: 

https://github.com/user-attachments/assets/746644c1-2c01-4ec1-87fe-7c4caa37c1d9

lib/Controller/CardOcsController.php
```diff
@@ -170,4 +170,28 @@ public function reorder(int $cardId, int $stackId, int $order, ?int $boardId): D
...
 		return new DataResponse($this->cardService->reorder($cardId, $stackId, $order));
 	}
+
+	#[NoAdminRequired]
+	#[PublicPage]
+	public function assignDependentCard(int $cardId, int $dependentCardId, ?int $boardId = null): DataResponse {
+		if ($boardId) {
+			$board = $this->boardService->find($boardId, false);
+			if ($board->getExternalId()) {
+				// External board support can be added later if needed
```
grnd-alt: if we don't implement it yet we should throw a not implement exception here

lib/Migration/Version11002Date20260410000000.php
```diff
@@ -0,0 +1,29 @@
...
+use Closure;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+class Version11002Date20260410000000 extends SimpleMigrationStep {
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options) {
+		$schema = $schemaClosure();
+
+		if ($schema->hasTable('deck_cards')) {
+			$table = $schema->getTable('deck_cards');
```
grnd-alt: I'm curios about the thought process of storing related cards as json field. Why not have it as an actual db relation so we can have both `depends on` and `required by` displayed.
Storing it this way also removes the ability to auto-delete relations, right now if a card is deleted the depending json will never be changed and the frontend will keep trying to retrieve that card, while the json only keeps growing.

grnd-alt (changes requested):
didn't review everything yet. It basically works, but looking at the comment in the migration step, I think this part needs a rework.

lib/Controller/CardOcsController.php
```diff
@@ -170,4 +171,28 @@ public function reorder(int $cardId, int $stackId, int $order, ?int $boardId): D
...
+		return new DataResponse($this->cardService->assignDependentCard($cardId, $dependentCardId));
+	}
+
+	#[NoAdminRequired]
+	#[PublicPage]
+	public function removeDependentCard(int $cardId, int $dependentCardId, ?int $boardId = null): DataResponse {
+		if ($boardId) {
+			$board = $this->boardService->find($boardId, false);
+			if ($board->getExternalId()) {
+				// External board support can be added later if needed
```
grnd-alt: this also has to throw notImplemented

grnd-alt (changes requested):
general comments while testing:
* sidebar view is not working well the selector expands over the sidebar and is only half visible (TagSelector is quite similar we can probably take some of that)
 and when clicking on the selector the selector overflows the initial div: 

Image

* clicking on the radio button in the list opens the dependent card and does not mark it as done
* debatable: I think we should have the selector not show all selected cards _and_ show it again below. Makes it feel cluttered to me.

luka-nextcloud:
@grnd-alt 
> sidebar view is not working well the selector expands over the sidebar and is only half visible

It is strange. I cannot reproduce this issue. It works well on my side.
image

luka-nextcloud:
> clicking on the radio button in the list opens the dependent card and does not mark it as done

For now, this is just an icon to indicate the status of the depend card. Hmm... it is reasonable to add a quick done button for the depend card

luka-nextcloud:
@grnd-alt @marcoambrosini 
> debatable: I think we should have the selector not show all selected cards andshow it again below. Makes it feel cluttered to me.

It sound reasonable. I think we need design advice.

luka-nextcloud:
Demo:
[demo.webm](https://github.com/user-attachments/assets/957ebbb3-f880-4c2a-a9e7-aac65fa2035e)

grnd-alt (approved):

PR #7921, opened 2026-05-07
fix: show avatars in selector

grnd-alt:
before:
image

now:
image

### DIFF
--- a/src/components/card/AssignmentSelector.vue
+++ b/src/components/card/AssignmentSelector.vue
@@ -19,19 +19,9 @@
 			label="displayname"
 			track-by="multiselectKey"
 			@option:selected="onSelect"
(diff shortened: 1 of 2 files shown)


src/components/card/AssignmentSelector.vue
```diff
@@ -19,19 +19,9 @@
 			label="displayname"
 			track-by="multiselectKey"
 			@option:selected="onSelect"
-			@option:deselected="onRemove">
-			<template #tag="scope">
-				<div class="avatarlist--inline">
```
grnd-alt, line 24: this template in here was not rendered in the first place, user bubble is rendered by nextcloud/vue based on user-select

benjaminfrueh (approved):

PR #7940, opened 2026-05-13
feat: allow read-only users to assign labels to cards (#590)

### LINKED ISSUE #590: Adding tags without editing
Hi, I'm pretty new on this, if I make a mistake please let me know.

Could you add the option to grant permission to users so that they could only add tags? At this time you can only add tags giving permission to edit, but this includes being able to remove decks ... is that possible?

Thank you

guangningsun:
## Summary

Fixes #590 — Allow users with read-only board access to assign labels to cards.

### Problem

Users with only `PERMISSION_READ` on a board could not assign labels to cards, even though the board owner intended them to self-manage card labels. The `assignLabel()` method required `PERMISSION_EDIT`, which is too restrictive for collaborative workflows.

### Solution

Lower the permission check in `CardService::assignLabel()` from `PERMISSION_EDIT` to `PERMISSION_READ`:

```diff
- $this->permissionService->checkPermission($this->cardMapper, $cardId, Acl::PERMISSION_EDIT);
+ $this->permissionService->checkPermission($this->cardMapper, $cardId, Acl::PERMISSION_READ);
```

`removeLabel()` still requires `PERMISSION_EDIT`, so read-only users can add labels but cannot remove them.

### Additional Changes

- Added 4 unit tests covering permission scenarios:
  - `testAssignLabelWithReadPermission` — READ-permission user CAN add labels
  - `testAssignLabelWithoutReadPermission` — no-permission user CANNOT add labels  
  - `testRemoveLabelRequiresEditPermission` — EDIT-permission user CAN remove labels
  - `testRemoveLabelWithoutEditPermission` — no-permission user CANNOT remove labels

### BC Note

This is a **backwards-compatible bug fix**. Read-only users who previously could not add labels now can, but they still cannot remove labels or modify card content.

---

Closes #590

### DIFF
--- a/lib/Service/CardService.php
+++ b/lib/Service/CardService.php
@@ -617,7 +617,7 @@ class CardService {
 	public function assignLabel(int $cardId, int $labelId): Card {
 		$this->cardServiceValidator->check(compact('cardId', 'labelId'));
 
-		$this->permissionService->checkPermission($this->cardMapper, $cardId, Acl::PERMISSION_EDIT);
+		$this->permissionService->checkPermission($this->cardMapper, $cardId, Acl::PERMISSION_READ);
 		$this->permissionService->checkPermission($this->labelMapper, $labelId, Acl::PERMISSION_READ);
 
 		if ($this->boardService->isArchived($this->cardMapper, $cardId)) {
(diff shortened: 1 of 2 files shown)


grnd-alt:
thanks a lot for your contribution! In the discussion of the issue it was made clear that the intended fix is not to make assigning tags a read-only option but rather make deleting boards an action that requires manage permission, which has been done as a side effect of https://github.com/nextcloud/deck/pull/1545, hence I am closing this pr and the underlying issue as well.

PR #7981, opened 2026-05-26
fix(board): truncate stack title for read-only and archived boards

### LINKED ISSUE #7980: fix: Stack title overflows into cards for read-only board members
## Bug description

For users with **read-only access** to a board (`canManage = false`) or on archived boards, the list/stack column header title wraps to multiple lines and visually overlaps the cards below it. Users with manage rights see the title correctly truncated with `…`.

## Steps to reproduce

1. Share a board with a user as read-only (no manage rights)
2. Log in as that user and open the board
3. The board must have at least one list with a long title

**Result:** Title wraps to multiple lines and overlaps the cards.  
**Expected:** Title is truncated to a single line with `…`

## Root cause

In `src/components/board/Stack.vue` there are two `<h3>` branches:

- **Manageable** (`v-else-if="!editing"`): has `class="stack__title"` → gets `white-space: nowrap; overflow: hidden; text-overflow: ellipsis` → correct.
- **Read-only / archived** (`v-if="!canManage || isArchived"`): plain `<h3>` without the class → no overflow constraints → wraps freely.

## Environment

- Affects all browsers (Chrome, Firefox, WebKit)
- Reproducible on Nextcloud Deck v1.17.1

Superjamp:
## Summary

- Fixes #7980
- The read-only `<h3>` branch in `Stack.vue` was missing `class="stack__title"` and the `<span dir="auto">` wrapper, causing long list titles to wrap into cards instead of being clipped with `…`
- Added the same class and span structure as the editable branch — no logic change, CSS-only fix

## Test plan

- [ ] Share a board with a user who has no manage rights
- [ ] Log in as that user and open the board with a long list title
- [ ] Verify the title is now truncated with `…` instead of wrapping
- [ ] Verify archived boards show the same truncation
- [ ] Verify users with manage rights still see the title correctly (no regression)

### DIFF
--- a/src/components/board/Stack.vue
+++ b/src/components/board/Stack.vue
@@ -10,8 +10,11 @@
 			:class="{'stack__header--add': showAddCard, 'stack__header--done-column': isDoneColumn}"
 			:aria-label="stack.title">
 			<transition name="fade" mode="out-in">
-				<h3 v-if="!canManage || isArchived" tabindex="0">
-					{{ stack.title }}
+				<h3 v-if="!canManage || isArchived"
+					tabindex="0"
+					:title="stack.title"
+					class="stack__title">
+					<span dir="auto">{{ stack.title }}</span>
 					<CheckCircleOutline v-if="isDoneColumn"
 						class="stack__done-icon"
 						decorative />


grnd-alt (changes requested):
code change is generally okay thanks a lot!
please fix your DCO commit message (you can do that by commiting with the -s flag), and ping me when that's done.

PR #8007, opened 2026-06-01
fix: re-label "dependendant card" card to "child card"

blizzz:
from Design feedback we got.

I did not find other occurrences in user visible strings.

### DIFF
--- a/src/components/card/DependentCardsSelector.vue
+++ b/src/components/card/DependentCardsSelector.vue
@@ -48,12 +48,12 @@
 				class="add-dependency-button"
 				type="button"
 				variant="tertiary"
-				:aria-label="t('deck', 'Add dependent card')"
+				:aria-label="t('deck', 'Add child card')"
 				@click="openSelector">
 				<template #icon>
 					<Plus :size="20" />
 				</template>
-				{{ t('deck', 'Add dependent card') }}
+				{{ t('deck', 'Add child card') }}
 			</NcButton>
 
 			<div v-if="isEditable && showSelector" class="selector-row">


AndyScherzinger (approved):

jancborchardt (approved):

jospoortvliet:
I think this is not right. Subtasks or child tasks we have - the checklist. Dependencies are what GANTT charts are about, it's a related but different thing.

| **Concept** | **Meaning** | **Example** |
| :-: | :-: | :-: |
| Dependency | Task 1 cannot start/finish until task 2 is done | “Deploy” depends on “Testing” |
| Subtask / Child task | Task is part of another larger task | “Testing” is part of “Release v1.0” |

So if we want to use this to show GANTT charts, we should use dependencies.


A bit more from a discussion in chat:

The current checklist you can put in is the definition of sub tasks: you have a list of 7 things to tick off. When all are ticked off, the task is done.

The dependency / blocker / prequisite / preceding or however we name it thing is different. If a card depends on 2 others, when those others are done, this one is not yet done - instead, you can now finally START on it.

So, in questions:

Can the parent be considered “done” when all children are done?
If yes, that’s a parent/child relationship.
(checklist ticked off-> task done)

Does one task need another task before it can proceed?
If yes, that’s a dependency.
(task done -> you can finally start the task(s) that depended on this one)

Dependencies are the new feature - needed for GANTT charts. A GANTT chart doesn't need subtasks, but it does need predecessors/successors or dependencies. As a GANTT chart is about order, and subtasks don't imply order, but predecessors/etc do

And Eric:
> Predecessors/successors are a story on their own, but if we keep it simple the only thing they add would be a line in the Gantt.

That is indeed a feature request for the GANTT chart -  and I would also, when moving and resizing tasks, put in limits - like, you can't move the start of a task before its predecessor/dependency is done. See https://github.com/nextcloud/deck/issues/8010

So I would not merge this change.

grnd-alt:
I am with Jos on this one.

Yes the current implementation feels more like a child relation, so the renaming does make sense, but the actual goal of that implementation was to use it for lines in gantt view soon, so now making it a child relation by naming will make it confusing when the gantt line feature thing actually comes around. The gantt relation is clearly a depends on/required by, thats exactly the same feature as child/parent from an implementation pov, but afaik we implemented the card relation feature clearly for the gantt view and not to have subtasks, right?

I would be open to have both child/parent _and_ depends on/required by, but I don't think we should display something called a child as a dependency in the gantt view.

blizzz:
Alright, makes sense. So then we close this PR for now, and create a new issue based on Jos statement, and follow up on that.

jospoortvliet:
let me close this then... no need to track it further. I created an issue to signify the dependency relation in the GANTT view.

PR #8013, opened 2026-06-03
fix: improve gantt rendering

grnd-alt:
this is mainly a workaround to make the gantt chart usable for bigger boards, we should put some effort into this issue upstream https://github.com/frappe/gantt/issues/486

### DIFF
--- a/src/components/board/GanttView.vue
+++ b/src/components/board/GanttView.vue
@@ -87,22 +87,6 @@ const DAYS_PER_UNIT = {
 
 const GANTT_VIEW_MODES = [
 	{
-		name: 'Hour',
-		padding: '7d',
-		step: '1h',
-		snap_at: '1h',
-		date_format: 'YYYY-MM-DD HH:',
-		lower_text: 'HH',
-		upper_text: (d, ld, lang) =>
-			!ld || d.getDate() !== ld.getDate()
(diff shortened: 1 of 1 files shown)


src/components/board/GanttView.vue
```diff
@@ -207,6 +207,31 @@ export default {
 					if (!card.duedate && !card.startdate) {
 						undatedCards.push(card)
 					} else {
+						// gantt renders everything at once so large date ranges cause performance issues on render
+						// therefore we limit the timeframe of visible tasks
+						const duedate = new Date(card.duedate)
+						switch (this.currentViewMode) {
+						case 'Hour':
```
blizzz, line 214: 💄 indentation looks odd on Github (tabs vs spaces?)

src/components/board/GanttView.vue
```diff
@@ -207,6 +207,31 @@ export default {
...
+							if (duedate < new Date() - 2 * 24 * 3600 * 1000) {
```
blizzz, line 215: so cards from the past are cut off, right?

blizzz (approved):

src/components/board/GanttView.vue
grnd-alt: yes only from the past

src/components/board/GanttView.vue
grnd-alt: apparently, the `lint:fix` wants case and switch on the same column...

PR #8037, opened 2026-06-09
feat(rtl): add Persian/Arabic font and automatic text direction

SadeqRouhani:
Bundle the Vazirmatn variable font scoped to Arabic/Persian Unicode ranges via @font-face unicode-range, so those glyphs render in Vazirmatn regardless of the interface language while Latin text keeps the Nextcloud system font. Re-assert the font on the rich-text editors (comment input, EasyMDE/CodeMirror and Text/ProseMirror description) whose scoped styles would otherwise override the inherited stack.

Add a small helper that tags editable fields and user-content elements with dir="auto" (via an initial pass plus a MutationObserver for lazily rendered nodes) so the browser resolves their base direction from the content. Switch the whole layout to RTL when the interface language is Persian or Arabic.

The comment form derives its direction from the first strong character typed so the submit button lands on the correct side and the send arrow points toward the writing direction.

Both stylesheets are loaded from shared-init.js so they apply across every Deck entry point.

Vazirmatn is licensed under OFL-1.1; declare it in REUSE.toml and add the license text.

* Resolves: # 
* Target version: main

### Checklist


grnd-alt:
thanks for your contribution here as well, We should definitely get the text direction correct everywhere (idk why is it not auto on every text field by the browser) but that is not deck specific, so I am not a fan of adding extra scripts for that in deck as it will cause issues. for the comment form we can easily add the text direction auto to it, as I now have done in a smaller pr here: https://github.com/nextcloud/deck/pull/8044

I think it makes more sense to have the dir="auto" set in nextcloud/vue where the text field components all come from and not in the specific apps that use them.

PR #8038, opened 2026-06-09
feat(board): show card count next to the list title

SadeqRouhani:
Display the number of cards in each list as a small badge after the list title in the stack header. The count reuses the existing cardsByStack computed, so it follows the archived/active filter and stays reactive as cards are added, moved or removed. The badge does not shrink, keeping it visible when a long title is truncated.

* Resolves: # 
* Target version: main

### TODO

- [ ] ...

### Checklist


grnd-alt (approved):
thanks for your contribution

PR #8042, opened 2026-06-10
feat(card): upload images pasted from the clipboard as attachments

SadeqRouhani:
* Resolves: # 
* Target version: main

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/src/components/card/AttachmentList.vue
+++ b/src/components/card/AttachmentList.vue
@@ -209,10 +209,71 @@ export default {
 			},
 		},
 	},
+	mounted() {
+		document.addEventListener('paste', this.handlePaste)
+	},
+	beforeDestroy() {
+		document.removeEventListener('paste', this.handlePaste)
+	},
 	methods: {
 		...mapActions([
 			'fetchAttachments',
 		]),
+		handlePaste(event) {
+			if (this.isReadOnly) {
+				return
+			}
+			// Only react when the attachments tab is actually visible (offsetParent
+			// is null when the tab is hidden or the component is detached).
+			if (!this.$el || this.$el.offsetParent === null) {
+				return
+			}
+			// Don't hijack pastes happening inside text fields (e.g. the comment box).
+			const target = event.target
+			if (target && (target.tagName === 'INPUT' || target.tagName === 'TEXTAREA' || target.isContentEditable)) {
+				return
+			}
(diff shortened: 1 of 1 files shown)


src/components/card/AttachmentList.vue
```diff
@@ -209,10 +209,71 @@ export default {
...
+				'image/jpeg': 'jpg',
+				'image/gif': 'gif',
+				'image/webp': 'webp',
+				'image/bmp': 'bmp',
+				'image/svg+xml': 'svg',
+			}
+			const extension = extensions[file.type] ?? (file.type.split('/')[1] ?? 'png')
+			const d = new Date()
+			const pad = (n) => String(n).padStart(2, '0')
+			const stamp = `${d.getFullYear()}${pad(d.getMonth() + 1)}${pad(d.getDate())}-${pad(d.getHours())}${pad(d.getMinutes())}${pad(d.getSeconds())}`
```
luka-nextcloud, line 268: To ignore duplicate, it should be
```
const stamp = `${d.getFullYear()}${pad(d.getMonth() + 1)}${pad(d.getDate())}-${d.getTime()}`
```

src/components/card/AttachmentList.vue
```diff
@@ -209,10 +209,71 @@ export default {
...
+			// multiple pastes. Give each a unique, human-readable name.
+			const extensions = {
+				'image/png': 'png',
+				'image/jpeg': 'jpg',
+				'image/gif': 'gif',
+				'image/webp': 'webp',
+				'image/bmp': 'bmp',
+				'image/svg+xml': 'svg',
+			}
+			const extension = extensions[file.type] ?? (file.type.split('/')[1] ?? 'png')
```
luka-nextcloud, line 265: I think we should extract file extension from file name instead

luka-nextcloud (changes requested):

HjortCreations:
Thanks for working on this. We needed the feature on a current production installation and have published a small standalone app as a temporary workaround: https://github.com/HjortCreations/nextcloud-deck-clipboard

It uses Deck's existing attachment endpoint, unique millisecond-based filenames, and avoids an automatic reload while a text field is active. Version 0.1.0 is tested with Nextcloud 34.0.4 and Deck 1.18.4.

We would still prefer this capability upstream; sharing the app in case its implementation and production testing are useful.

PR #8066, opened 2026-06-15
fix: cast boardUsers to string

grnd-alt:


lib/Service/AssignmentService.php
```diff
@@ -62,7 +62,7 @@ public function assignUser(int $cardId, string $userId, int $type = Assignment::
 
 		$card = $this->cardMapper->find($cardId);
 		$boardId = $this->cardMapper->findBoardId($cardId);
-		$boardUsers = array_keys($this->permissionService->findUsers($boardId, true));
+		$boardUsers = array_map(fn (string|int $userId) => (string)$userId, array_keys($this->permissionService->findUsers($boardId, true)));
```
grnd-alt, line 65: php converts array keys, that are inserted as strings, to int if possible, so the later `in_array` fails if the `user_id` passed is a string but `boardUsers` contains ints for numeric userIds

blizzz (approved):

PR #8052, opened 2026-06-11
🕵️ Add sharereview support

AndyScherzinger:
* Needs merged: #8069
* Target version: main

### Summary

Enabled the share review app to retrieve general share information and enabled the app to delete shares.

### TODO

- [x] run CI
- [x] clarify share meta data

### Checklist

- [x] Code is properly formatted
- [x] Sign-off message is added to all commits
- [x] Tests (unit, integration, api and/or acceptance) are included
- [ ] Documentation (manuals or wiki) has been updated or is not required

### 🤖 AI (if applicable)

- [x] The content of this PR was partly or fully generated using AI

PR #8069, opened 2026-06-16
🧾Add share metadata

AndyScherzinger:
* Resolves: pre-requisite for #8052 (WiP)
* Target version: main

### Summary

While starting to work on #8052 (nowhere near completion..) to implement the sdhare review capability I realized that the share date always matched 1970 and digging and asking @grnd-alt it turned out deck shares have no creation and no last-edit date, like for example files/folder or tables shares. Yet for a share review app, that also stores a "review date" such timestamps are important to show a changed share for a re-review in case it changed. Hence creation and edit dates a crucial. 

### TODO

- [x] get CI to turn green 💚 

### Checklist

## 🤖 AI (if applicable)

- [x] The content of this PR was partly or fully generated using AI


lib/Migration/AclTimestampBackfill.php
```diff
@@ -0,0 +1,68 @@
...
+			$output->info('AclTimestampBackfill: no rows to update');
+			return;
+		}
+
+		// Group ACL IDs by target timestamp to issue one UPDATE per group
+		// rather than one UPDATE per row (avoids N+1 queries).
+		$now = time();
+		$groups = [];
+		foreach ($rows as $row) {
+			$timestamp = ((int)$row['board_last_modified'] > 0) ? (int)$row['board_last_modified'] : $now;
```
grnd-alt: idk if we should really backfill that way, does the app you're doing all that for not have an option to have an unknown value?
The values we're writing in here 99% of the time are not the real created/last_modified of the acl, so an unknown value would be better imo, or at least just the "now" value so it is clear that the value was written when the update was done.

grnd-alt (changes requested):

lib/Migration/AclTimestampBackfill.php
AndyScherzinger: We can use any we would want, the share review app would else have to assume one and either that is a fully hard-coded and fixed date, no matter the share, or needs to be computed each time the shares are loaded for review. So to me, compute-once made the most sense and going with a creation/modification date of the board itself seemed most feasible. However we could also go for `$now` if you think that would be better. But in any care I suggest to pre-compute and migrate to a date and not have a fallback logic because we lacked dates in the past.

WDYT @grnd-alt @blizzz ? "No date" feels wrong to me, but that might be me, than we can also just leave it at `0` and take that as "unknown". 

Both has its advantages, one is unknown: cause it hasn't been tracked but mid-term I find this confusing to users when there are shares with known and unknown dates.

Both works for me. I can find a way to explain the unknown in the review app of course.

blizzz: My general preference is having a one time (migration) operation over some extra business logic code that then has to be kept maintained over a long time/forever. So this would be in favor of the update step. 

I guess this could run long, so there should be some progress output to avoid wild cancellation by the admin. 

Having a "now" date in there seems odd in itself though and might lead to wrong assumptions. This is the more critical question imo: what does it tell users and can it be misleading? Alternative coulde setting to 01-01-1970 and still handling as "unknown", but this brings the extra treatment again.

AndyScherzinger: I'll then go for `0` as in `01-01-1970`. Currently not shown in the deck UI anyways and the share review app can than have some handling for the magic number 👍

grnd-alt (approved):

PR #8089, opened 2026-06-20
fix(card): show long card titles in full and edit them on multiple lines

SadeqRouhani:
The card sidebar header truncated long titles to a single line with an ellipsis and edited them in a single-line input, so the full text was never visible. Let the title wrap onto multiple lines and replace the header edit input with an auto-growing textarea.

Because NcAppSidebar renders the title input itself with no slot to override it, the textarea is injected in place of the input when editing starts; the original input is kept (hidden) so the library's form submit and click-outside handling keep working. Pressing Enter applies the title instead of inserting a newline, and Enter is ignored while an IME is composing (e.g. Persian/Arabic input).

* Resolves: # 
* Target version: main

### TODO

- [ ] ...

### Checklist

### DIFF
--- a/src/components/card/CardSidebar.vue
+++ b/src/components/card/CardSidebar.vue
@@ -214,8 +214,72 @@ export default {
 				this.titleEditing = newTitle
 			},
 		},
+		isEditingTitle(editing) {
+			if (editing) {
+				// The sidebar header uses a single-line <input> for the title.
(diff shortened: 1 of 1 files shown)


luka-nextcloud:
@SadeqRouhani Could you please share before/after screenshots?

PR #8109, opened 2026-06-23
Fix: overview page is too slow to load

mintsoft:
* Resolves: 
* Target version: main

### Summary
On my current configuration, I have approximately 400 cards in the no-due section on the overview that actually live across multiple boards. This means that if I click the overview page I have to wait about 10 seconds for it to load before I can go somewhere else and this column is of no-use to me at all. I do use the due date on specific cards that are time sensitive so the page is useful for the other columns.

### TODO

- [x] Test this actually works

### Checklist


mintsoft:
Very rough testing I was doing with this today, with ~600 cards with noDue date; it was taking about 27 seconds to load before and approx 700ms now.

PR #8112, opened 2026-06-26
fix: improve card open performance on boards with many cards

### LINKED ISSUE #3710: Slow navigation
Hi everyone, I’m trying to implement a management software on Nextcloud through the Deck application.

I have set up 20 boards with over 300 tasks and attached pdf files but the system is really slow. it is currently installed on a service of a German server company.
Usually Nextcloud freezes and the system becomes unusable.

Do you have to suggest me a correct version of Nextcloud and the Deck application to avoid usability problems?

Thanks

nemphys:
* Resolves: #3710 
* Target version: main

### Summary

Fixes severe UI lag when opening a card on boards with a large number of cards (e.g. 200+). Clicking a card to open its details could block the main thread for several seconds due to forced layout recalculations.

## Problem
On boards with many cards, opening a card's sidebar was unbearably slow. Chrome DevTools showed multi-second main-thread blocks dominated by **Recalculate Style**, triggered from Vue's `<transition-group>` update logic (`hasMove` → `getTransitionInfo`).
Two issues compounded:
1. **`<transition-group>` on card labels** — Each `CardItem` wrapped its label list in a `<transition-group>`. On any component update, Vue checks whether children have moved by reading layout from the DOM. With hundreds of cards (each potentially having multiple labels), this caused thousands of forced reflows on every interaction that updated the card list.
2. **`$route` in a computed property** — `currentCard` depended on `$route.params.cardId`, so **every** card component re-rendered whenever the route changed (e.g. opening or closing the card sidebar), not just the previously selected and newly selected card.
## Solution
- Replace the label `<transition-group>` with a plain `<ul>`. Label add/remove zoom animations are dropped, which is an acceptable trade-off for large boards.
- Track selection state in local `isCurrentCard` data, updated via a route watcher that only mutates state when **this** card's selection actually changes. Only the two affected cards re-render for the highlight border.

### Checklist

- [ ] Open a board with 100+ cards and click a card — sidebar should open without a noticeable delay
- [ ] Verify the selected card still shows the primary-color border highlight
- [ ] Close the sidebar and confirm the highlight is removed from the previously selected card
- [ ] Confirm labels still render and label-filter clicks still work
- [ ] Smoke-test on a board with few cards to ensure no regressions in normal use
- [ ] Optional: record a Chrome Performance profile while opening a card — the long `Recalculate Style` block from `hasMove` should be gone

### DIFF
--- a/src/components/cards/CardItem.vue
+++ b/src/components/cards/CardItem.vue
@@ -6,7 +6,7 @@
 <template>
 	<AttachmentDragAndDrop v-if="card" :card-id="card.id" class="drop-upload--card">
 		<div :ref="`card${card.id}`"
-			:class="{'compact': compactMode, 'current-card': currentCard, 'no-labels': !hasLabels, 'card__editable': canEdit, 'card__archived': card.archived, 'card__highlight': highlight}"
+			:class="{'compact': compactMode, 'current-card': isCurrentCard, 'no-labels': !hasLabels, 'card__editable': canEdit, 'card__archived': card.archived, 'card__highlight': highlight}"
 			:style="{backgroundColor: color}"
 			tag="div"
 			:tabindex="0"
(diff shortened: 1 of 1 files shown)


grnd-alt (approved):
hey, thanks for your contribution!
Together with https://github.com/nextcloud/deck/pull/8140 I think it could be possible to revert the transition-group change to keep the animations, but that's something to consider after the other pr is merged, so let's get this one in for now 🔥

PR #8121, opened 2026-06-29
fix(sharing): show more board sharing targets and clarify the picker label

maxemig:
The board sharing picker supports users, groups, and teams, but it only requested 20 autocomplete results and displayed the first 10 unallocated entries. This may give the impression that only users are supported.

Raise both limits to 200 and use a more specific picker label to clarify the supported sharing targets.

The 200-result limit follows established behaviour in Calendar, Tasks, and Bookmarks.

* Resolves: n/a, discussed in multiple places no single issue just for that.
* Target version: main

### Summary

In my local Nextcloud instance, I was only ever suggested 10 users for sharing a board with, the label _Select account_ with the absence of any suggested groups led me to believe that group-sharing was not supported.

It turned out, after searching existing groups are found just fine.

Two improvements to resolve that UI/UX gripe, bump the limit to 200 displayed users/groups/teams to share a board with in line with Calendar, Tasks, and Bookmarks and clarify in the label that teams and groups are also supported.

### TODO

n/a

### Checklist

### DIFF
--- a/src/components/board/SharingTabSidebar.vue
+++ b/src/components/board/SharingTabSidebar.vue
@@ -7,6 +7,7 @@
 		<NcSelectUsers v-model="addAcl"
 			:options="formatedSharees"
 			:loading="isLoading"
+			:input-label="t('deck', 'Search for users, groups and teams')"
 			@search="(search) => asyncFind(search)" />
 
 		<ul id="shareWithList"
@@ -143,7 +144,7 @@ export default {
 					type: SOURCE_TO_SHARE_TYPE[item.source],
 				}
 				return res
-			}).slice(0, 10)
+			}).slice(0, 200)
 			return result
 		},
 		unallocatedSharees() {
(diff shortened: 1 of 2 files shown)


luka-nextcloud (approved):

issb-gh:
Thanks for addressing this. We reproduced the same issue in our Nextcloud environment.

In our case, the group name is also part of many users' email addresses. Therefore the autocomplete returns many matching users before the actual group. With the current `.slice(0, 10)` limit, the group is not displayed at all.

We verified the cause by locally changing the frontend limit from 10 to 20. After a hard browser reload, the group immediately became selectable.

Increasing the limits to 200 is a clear improvement, but I wonder whether this still leaves the same issue for larger installations. With more than 200 matching users, an exact group match could still be hidden.

Would it make sense to prioritize exact matches, especially exact group matches, before applying the result limit? That would make the picker robust regardless of the number of matching users.

Environment:
- Nextcloud 34.0.1
- Deck 1.18.2
- LDAP / Active Directory user backend
- Firefox

PR #8128, opened 2026-07-02
refactor(pinia): migrate dashboard store

grnd-alt:


src/init-dashboard.js
```diff
@@ -17,59 +17,60 @@ const getAsyncImports = async () => {
...
-	const { default: dashboard } = await import('./store/dashboard.js')
+	const { createPinia, PiniaVuePlugin } = await import('pinia')
+	const { default: Vuex} = await import('vuex')
 
 	Vue.prototype.t = t
 	Vue.prototype.n = n
 	Vue.prototype.OC = OC
+	const pinia = createPinia()
+	Vue.use(PiniaVuePlugin)
 	Vue.use(Vuex)
```
grnd-alt, line 28: we are still injecting vuex here because some components in the dashboard require vuex state from not yet migrated stores. for example: https://github.com/nextcloud/deck/blob/main/src/components/cards/badges/DueDate.vue#L48

mejo- (approved):
Only looked through the code and didn't spot anything obvious.

PR #8148, opened 2026-07-09
fix: align attachment frontend requests with registered OCS routes

Lightshadow02:
* Resolves: attachment delete/update/restore silently failing in the UI
* Target version: main

### Summary

696ca2f3e52b59a08c390c9351951337b66d13af migrated the attachment frontend to OCS endpoints, but the frontend kept the old page-route URL shape (singular `/attachment/`, composite `{type}:{id}` ids, `POST`/`GET` verbs) while the registered OCS routes use plural `/attachments/{attachmentId}` with `PUT`/`DELETE`. As a result, deleting, updating and restoring an attachment from the card sidebar hit non-existent routes and silently failed with 404 (the Vuex dispatch has no error handling).

This PR aligns the frontend with the registered OCS routes and fixes the remaining gaps on the backend side:

- `AttachmentApi.js`: call `/cards/{cardId}/attachments/{attachmentId}` with a numeric id, pass the attachment `type` as a query parameter (the OCS controller defaults to `file`, so `deck_file` attachments need it explicitly), use `PUT` for restore, and unwrap the OCS envelope in `updateAttachment()`.
- `routes.php`: register a `POST` route for `attachment_ocs#update` next to `PUT`, since PHP only parses `multipart/form-data` bodies (file uploads) for `POST` requests — mirroring the old page routes which registered both verbs for the same reason.
- `AttachmentOcsController::update()`: default `$data` to an empty string, as the upload form data contains `file`/`type` but no `data` field.
- Add jest regression tests covering the URL, verb and parameters of all five attachment endpoints. This is the first jest test in the repo; the jest config in `package.json` pointed at `vue-jest`, which is not a dependency (the installed transformer is `@vue/vue2-jest`), so it is fixed in a separate commit.

### How to test

1. Open a card → attachments tab → upload a file
2. Delete it via the three-dot menu → the attachment should be greyed out (soft-deleted) and the `DELETE .../ocs/v2.php/apps/deck/api/v1.0/cards/{cardId}/attachments/{attachmentId}?type=deck_file` request should return 200 (previously 404)
3. Restore it → it comes back (`PUT .../restore`)
4. `npx jest src/services/AttachmentApi.spec.js` → 5 passing

### Checklist

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -160,6 +160,8 @@ return [
 		['name' => 'attachment_ocs#getAll', 'url' => '/api/v{apiVersion}/cards/{cardId}/attachments', 'verb' => 'GET'],
 		['name' => 'attachment_ocs#create', 'url' => '/api/v{apiVersion}/cards/{cardId}/attachment', 'verb' => 'POST'],
 		['name' => 'attachment_ocs#update', 'url' => '/api/v{apiVersion}/cards/{cardId}/attachments/{attachmentId}', 'verb' => 'PUT'],
+		// POST is needed next to PUT as file uploads (multipart/form-data) are only parsed by PHP for POST requests
+		['name' => 'attachment_ocs#update', 'url' => '/api/v{apiVersion}/cards/{cardId}/attachments/{attachmentId}', 'verb' => 'POST'],
 		['name' => 'attachment_ocs#delete', 'url' => '/api/v{apiVersion}/cards/{cardId}/attachments/{attachmentId}', 'verb' => 'DELETE'],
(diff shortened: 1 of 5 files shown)


Lightshadow02:
Friendly ping on this one — it is a small bug fix still waiting for a first review.

Since 696ca2f3e moved the attachment frontend to OCS, the client kept the old page-route URL shape (singular `/attachment/`, composite `{type}:{id}` ids, POST/GET verbs) which matches no registered OCS route, so deleting, updating and restoring an attachment silently fails with a 404 in the current main branch. The PR aligns the frontend with the registered routes and adds jest regression tests for all five attachment endpoints.

Issue #7871, opened 2026-04-21
Feature request: Sorting of the board list by recent activity


apaul101:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Is your feature request related to a problem? Please describe.**
We use Deck as an issue tracker, where we share boards with clients and allow them to create cards. After a while (e.g. after a weekend), it is tedious to find out which boards/cards have had some form of activity. It would be great to see this at a glance, similar to an inbox.

**Describe the solution you'd like**
A great solution would be the ability to sort the list of boards by recent activity, with more recently changed boards on top, and denoted with a dot or even better, the number of changes. That way one could see at a glance which boards need to be checked.

**Describe alternatives you've considered**
The alternative is using the activity tab in conjunction with Deck, which is cluttered and tedious. 

**Additional context**
Adding a sorting function to the board list should be fairly easy to implement with practically no downsides. Other requests for alphabetical sorting could be easily fulfilled as well.

luka-nextcloud:
@apaul101 You can check the latest boards/cards activity in Activity app instead.

<img width="1710" height="760" alt="Image" src="https://github.com/user-attachments/assets/9dc8e4c0-d689-4cff-bc0f-8379fa4e95db" />

PR #8150, opened 2026-07-10
feat: per-user board ordering in the navigation

Lightshadow02:
* Resolves: recurring request to control the order of boards in the navigation (related discussion: #7871)
* Target version: main

### Summary

The left navigation sorts boards alphabetically (hard-coded), which does not match how people prioritise their boards. This adds per-user manual ordering:

- New per-user config key `boardOrder` (JSON array of board ids) through the existing `ConfigService` user-config plumbing and the existing OCS `POST /config/{key}` route — **no schema migration, no shared state**: each user has their own order and boards themselves are untouched
- `getAll()` exposes `boardOrder` to the frontend initial state; value is validated (list of ints) and cleaned on write; documented in docs/API.md
- Frontend: drag & drop of boards in the "All boards" navigation category only (archived/shared categories keep the alphabetical sort), `sortBoards()` helper — user order first, boards not in the list appended alphabetically (new/newly shared boards), stale ids ignored
- Rename input and actions menu are non-drag areas; failures show an error toast and the order reverts
- Jest tests for the helper; PHPUnit tests for the config round-trip and validation

### How to test

1. Drag boards into a custom order in the "All boards" section; F5 persists it
2. A second user on the same boards keeps their own independent order
3. New boards appear at the end of the list alphabetically

### Checklist

### DIFF
--- a/lib/Service/ConfigService.php
+++ b/lib/Service/ConfigService.php
@@ -63,7 +63,7 @@ class ConfigService {
 	}
 
 	/**
-	 * @return bool|array{id: string, displayname: string}[]
+	 * @return bool|array{id: string, displayname: string}[]|int[]
 	 * @throws NoPermissionException
 	 */
 	public function get(string $key) {
--- a/lib/Service/ConfigService.php
+++ b/lib/Service/ConfigService.php
@@ -52,7 +52,8 @@ class ConfigService {
 		$data = [
 			'calendar' => $this->isCalendarEnabled(),
 			'cardDetailsInModal' => $this->isCardDetailsInModal(),
-			'cardIdBadge' => $this->isCardIdBadgeEnabled()
+			'cardIdBadge' => $this->isCardIdBadgeEnabled(),
+			'boardOrder' => $this->getBoardOrder()
 		];
 		if ($this->groupManager->isAdmin($userId)) {
 			$data['groupLimit'] = $this->get('groupLimit');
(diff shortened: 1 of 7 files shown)


lib/Service/ConfigService.php
luka-nextcloud: Should be updated to `@return bool|array{id: string, displayname: string}[]|int[]` as well

src/components/navigation/AppNavigationBoardCategory.vue
luka-nextcloud, line 1: I think we should add an icon to indicate and trigger the drag function since it is showing the board url when user trying to drag the board.

Image

luka-nextcloud (changes requested):

lib/Service/ConfigService.php
Lightshadow02: Good catch, updated in 9caa17ef7 — the annotation now includes `int[]`.

src/components/navigation/AppNavigationBoardCategory.vue
Lightshadow02: Makes sense — done in 9caa17ef7. Boards are now dragged through a dedicated handle icon (`drag-vertical`, revealed on row hover, `cursor: grab`) via `drag-handle-selector`, so grabbing the row itself no longer starts a drag and the browser's native link drag (the URL ghost from your screenshot) can't be triggered by the reorder gesture anymore.

Lightshadow02:
@luka-nextcloud thanks for the review! Both points are addressed in 9caa17ef7:

- `ConfigService::get()` return annotation now includes `int[]`
- boards are reordered through a dedicated drag handle (`drag-vertical` icon, revealed on row hover, `cursor: grab`) wired via `drag-handle-selector`, so grabbing the row itself no longer starts a drag and the native link drag with the URL ghost from your screenshot can no longer be triggered by the reorder gesture

Whenever you have a moment for another look.

PR #8151, opened 2026-07-10
fix: do not clear card color when update omits the field

### LINKED ISSUE #8131: Card update API clears the existing color when the color field is omitted
Related Android client issue: [stefan-niedermann/nextcloud-deck#1887](https://github.com/stefan-niedermann/nextcloud-deck/issues/1887)

## Description

The card update API treats an omitted optional `color` field as `null` and unconditionally clears the stored card color. This causes data loss for clients that do not support the newly added card color property.

## Confirmed reproduction

- Nextcloud: 34.0.1
- Deck server: 1.18.2
- Android Deck: 1.24.8
- Update endpoint: `PUT /apps/deck/api/v1.0/boards/{boardId}/stacks/{stackId}/cards/{cardId}`

Before:

```json
{"color":"FF5733"}
The update payload changes only the title and omits color.
After:
{"color":null}
The request succeeds with HTTP 200.
Expected behavior
When color is omitted, the existing value should remain unchanged. Only an explicit "color": null should clear it.
```

Lightshadow02:
* Resolves: #8131
* Target version: main

### Summary

`CardService::update()` unconditionally called `$card->setColor($color)`, so any API client that omits the `color` field on a card update (mobile clients, scripts, integrations) silently wiped the stored card color.

This applies the exact mechanism already used for the `done` field: the controllers build an `OptionalNullableValue` only when the request body actually contains a `color` key (`array_key_exists`), and the service only touches the color when the wrapper is present:

- field omitted → color kept (the fix)
- explicit `null` or `""` → color cleared
- hex string → color set

The service guards with `is_string()`, so malformed non-scalar payloads (e.g. an array) are treated as "clear" instead of being string-cast. Note: an unquoted numeric JSON value (`"color": 123456`), which previously coerced to the string `"123456"`, now also clears — strings are the documented type for this field.

The web frontend always sends the full card object including `color`, so UI behaviour is unchanged. Covered by four new PHPUnit cases (kept / cleared via null / cleared via empty string / set).

### How to test

1. Set a card color in the web UI
2. `PUT /index.php/apps/deck/cards/{cardId}` with a body that contains title/stackId/type/order but no `color` key
3. Before: color wiped. After: color kept. Sending `"color": null` still clears it.

### Checklist

### DIFF
--- a/lib/Controller/CardApiController.php
+++ b/lib/Controller/CardApiController.php
@@ -88,8 +88,9 @@ class CardApiController extends ApiController {
 	#[NoAdminRequired]
 	#[CORS]
 	#[NoCSRFRequired]
-	public function update(string $title, $type, string $owner, string $description = '', int $order = 0, $duedate = null, $startdate = null, $archived = null, ?string $color = null): DataResponse {
+	public function update(string $title, $type, string $owner, string $description = '', int $order = 0, $duedate = null, $startdate = null, $archived = null): DataResponse {
 		$done = array_key_exists('done', $this->request->getParams()) ? new OptionalNullableValue($this->request->getParam('done', null)) : null;
+		$color = array_key_exists('color', $this->request->getParams()) ? new OptionalNullableValue($this->request->getParam('color', null)) : null;
 		$card = $this->cardService->update($this->request->getParam('cardId'), $title, $this->request->getParam('stackId'), $type, $owner, $description, $order, $duedate, 0, $archived, $done, $startdate, $color);
 		return new DataResponse($card, HTTP::STATUS_OK);
 	}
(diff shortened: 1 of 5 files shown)


luka-nextcloud (approved):

Lightshadow02:
Just a note on the two red checks here: they look like the known-flaky Cypress e2e suite rather than a regression from this PR. The same workflow currently fails on `main` itself in its last four runs (2026-07-08, 07-09, 07-13, 07-15), each time on a different unrelated spec (`boardFeatures.js — Exports a board as JSON`, `cardFeatures.js — Set a due date`, …), none of which touch card colors. Everything else is green: the three PHPUnit matrices, psalm and the lint jobs.

The change itself is backend-only and covered by four unit tests (color kept when the field is omitted / cleared on explicit null / cleared on empty string / set from a hex value). Happy to rebase or re-run anything if that helps.

PR #8163, opened 2026-07-15
refactor(pinia): move stackStore to pinia

grnd-alt:


src/main.js
```diff
@@ -20,6 +20,10 @@ import { createPinia, PiniaVuePlugin } from 'pinia'
 // the server snap.js conflicts with vertical scrolling so we disable it
 document.body.setAttribute('data-snap-ignore', 'true')
 
+const pinia = createPinia()
+Vue.use(PiniaVuePlugin)
+pinia.use(() => ({ $vuex: store }))
```
samin-z, line 26: pinia is created before 'store', causing '$vuex' plugin to close over it still uninitialized. maybe we can keep the previous order to avoid latent TDZ?

samin-z (approved):

PR #8181, opened 2026-07-19
Add new cards to top or bottom

### LINKED ISSUE #6457: New cards are created at the bottom of the screen

theoholl:
* Resolves: #6457
* Target version: main

### Summary

Users can now choose whether new cards are added to the top or the bottom of a list. A new setting in the Deck settings toggles the behavior; the 'Add card' button sits on the side of the list where the card will be inserted. The default behavior is unchanged (add to bottom).

Settings toggle for changing between adding cards to the top or the bottom of the stack:

image

Adding cards to the top of the stack:

image

Adding cards to the bottom of the stack:

image

### Implementation notes

- New per-user setting `stackAddCardAtTop` (`ConfigService`), exposed to the client via the initial state and toggled in the Deck settings.
- The add-card form is extracted from `Stack.vue` into a new `StackCardAdd.vue` component and rendered above or below the card list depending on the setting.
- `CardService::create()` gained an `insertAtPosition` flag: the card is inserted and the surrounding cards are shifted in a single DB transaction, so no follow-up reorder request is needed.
- **API behavior note:** passing an explicit `order` to the OCS card create endpoint now inserts the card at that position and shifts the surrounding cards, instead of only storing the raw order value. Omitting `order` keeps the previous append-to-bottom behavior.
- Covered by unit tests (transaction commit + rollback) and a Cypress test that verifies both modes and persistence across reloads.

### Checklist


grnd-alt (changes requested):
Thanks for your contribution!
I like the idea in general but I think the implementation is way to complex. For a sticky element like that we shouldn't need that much js, I don't think we need a resizeObserver and bottomAddCardInline for example. For the handling with empty stacks, we could only use the isEmptyStack and remove isEmptyStackWithBottomAddCard, please also have a look at the other js parts of this pr to simplify it further.

some ui things I noticed:
* you can't really drag and drop to empty stacks as the dnd container shrinks
* when clicking "add card" the input gets bigger/wider which causes a sudden layout shift, would be nice to not have that
* subjective but I think having the "add card" with less opacity might look nicer to not draw away attention from the cards already there

theoholl:
Completely agreed, I'll work on that. Thanks for your review! :)

theoholl:
Previously, I placed the 'Add card' button right underneath the last card of the stack and then sticking to the bottom of the board when the cards overflowed. 

image

This had some disadvantages:

- Most importantly, for me, this introduces unpredictable behavior because cards that were dragged in from another stack had to be dropped _above_ the 'Add card' button which is different for add-cards-to-the-top stacks where one could drop off the card anywhere. This is especially confusing for users since there is no visible drop-off zone marked by a border or a background color.
- That could be fixed by allowing cards to be dropped anywhere on the stack but without any visual feedback it feels weird dropping a card on an add-cards-to-bottom stack below the 'Add card' button, basically into a void.
- Introducing visual feedback, e.g. making the 'Add card' button move down a bit to 'make room' for the new card when a card is moved onto the stack is also weird and perhaps one would want the same visual effect then for stacks with some cards on it?

I think those issues could be tackled with a clear 'drop box' like Trello.

image

But for now I would keep the exact same, predictable behaviour for add-to-top and add-to-bottom lists.

lib/Service/CardService.php
```diff
@@ -204,7 +206,21 @@ public function create(string $title, int $stackId, string $type, int $order, st
 		$card->setDuedate($duedate);
 		$card->setStartdate($startdate);
 		$card->setColor($color);
-		$card = $this->cardMapper->insert($card);
+
+		if (!$insertAtPosition) {
+			$card = $this->cardMapper->insert($card);
+		} else {
+			// Insert at the requested position and shift the surrounding cards
+			$this->connection->beginTransaction();
```
theoholl, line 214: Insertion and order normalization are one transaction, so a failed reorder cannot leave the new card stored with conflicting positions.

src/components/board/StackCardAdd.vue
```diff
@@ -0,0 +1,208 @@
...
+		async addCard() {
+			this.creating = true
+			this.$emit('creating')
+			try {
+				const newCard = await this.addCardInStore({
+					title: this.title,
+					stackId: this.stack.id,
+					boardId: this.stack.boardId,
+					// Without an order the API appends the card to the end of the stack
+					...(this.addAtTop ? { order: 0 } : {}),
```
theoholl, line 108: Bottom insertion deliberately omits `order` to preserve the existing append behavior. Top insertion sends `order: 0`; the server shifts the existing cards.

src/stores/card.js
```diff
@@ -280,6 +280,13 @@ export const useCardStore = defineStore('card', {
 		},
 		async addCard(card) {
 			const createdCard = await apiClient.addCard(card)
+			if (card.order !== undefined) {
+				for (const existingCard of this.cards) {
+					if (existingCard.stackId === createdCard.stackId && existingCard.order >= card.order) {
+						Vue.set(existingCard, 'order', existingCard.order + 1)
```
theoholl, line 286: Mirror the server-side shift locally so the new top card renders in the correct position immediately after creation.

theoholl (commented):
A few implementation notes for reviewers.

grnd-alt (approved):
much better state than before, I would say we can merge that.
rebasing onto main should also fix the federation tests again.

⚠️ please merge this with a squash so we don't get all the 40 commits into the history of main

PR #8190, opened 2026-07-23
fix: validate board permission in setDoneStack method

luka-nextcloud:
# Summary

Added an explicit permission check for the board in addition to the stack, and validated that the stack belongs to the specified board in `setDoneStack`.

### DIFF
--- a/lib/Service/StackService.php
+++ b/lib/Service/StackService.php
@@ -306,6 +306,11 @@ class StackService {
 	 */
 	public function setDoneStack(int $stackId, int $boardId, bool $isDone): void {
 		$this->permissionService->checkPermission($this->stackMapper, $stackId, Acl::PERMISSION_MANAGE);
+		$stack = $this->stackMapper->find($stackId);
+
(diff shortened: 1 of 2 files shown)


lib/Service/StackService.php
```diff
@@ -306,6 +306,12 @@ public function reorder(int $id, int $order): array {
 	 */
 	public function setDoneStack(int $stackId, int $boardId, bool $isDone): void {
 		$this->permissionService->checkPermission($this->stackMapper, $stackId, Acl::PERMISSION_MANAGE);
+		$this->permissionService->checkPermission($this->boardMapper, $boardId, Acl::PERMISSION_MANAGE);
```
grnd-alt: this check is already done with the line above by this line: https://github.com/nextcloud/deck/blob/update-setDoneStack/lib/Service/PermissionService.php#L149

so I would say the `$stack->getBoardId() !== $boardId` check is enough here

grnd-alt (changes requested):

grnd-alt (approved):

PR #8215, opened 2026-07-29
feat(deck): add board filter input to the board header

pringelmann:
### Summary

Server is removing the "local search bar" it rendered in the header for so Deck now owns its filter input. The filtering logic is unchanged - only where the query comes from.

image

Input now shown on a board and on the board list, which filters board titles from the same query; not on Upcoming, which has no consumer for it. Named "Filter", not "Search", per design feedback.

**Behaviour change:** the global search box no longer filters the board behind it. Those events are deprecated and the bar driving them is being deleted.

### Checklist

### AI (if applicable)

- [x] The content of this PR was partly or fully generated using AI (mostly tests)

### DIFF
--- a/src/components/Controls.vue
+++ b/src/components/Controls.vue
@@ -31,18 +31,6 @@
 		<div class="board-actions">
 			<SessionList v-if="isNotifyPushEnabled && presentUsers.length"
 				:sessions="presentUsers" />
-			<!-- Hide but not remove for now as search might change in the future -->
-			<div v-if="false" class="deck-search">
-				<input id="deck-search-input"
-					ref="search"
-					:tabindex="0"
-					type="search"
-					class="icon-search"
-					:value="searchQuery"
-					@focus="$store.dispatch('toggleShortcutLock', true)"
-					@blur="$store.dispatch('toggleShortcutLock', false)"
-					@input="$store.commit('setSearchQuery', $event.target.value)">
-			</div>
 			<div v-if="board && canManage && !showArchived && !board.archived"
 				id="stack-add"
 				v-click-outside="hideAddStack">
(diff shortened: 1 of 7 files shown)


pringelmann:
Fixing cypress test

kra-mo (changes requested):
> Named "Filter", not "Search", per design feedback.

Slept on this, I think it's actually fine to just call it "Search". In this case specifically since it doesn't only filter, we could call it "Find", but what's more important than distinct wording is probably clarification for what is being searched. So let's just stick with "Search cards".

src/components/Controls.vue
```diff
@@ -71,6 +59,24 @@
...
+					:label="filterLabel"
+					:value="searchQuery"
+					:show-trailing-button="searchQuery !== ''"
+					:trailing-button-label="t('deck', 'Clear filter text')"
+					aria-describedby="deck-filter-hint"
+					@update:value="setSearchQuery"
+					@trailing-button-click="clearSearchQuery"
+					@focus="$store.dispatch('toggleShortcutLock', true)"
+					@blur="$store.dispatch('toggleShortcutLock', false)" />
+				<span id="deck-filter-hint" class="hidden-visually">{{ filterHint }}</span>
```
grnd-alt: that's only shown for screen readers that way would be better to have it as an actual hint (maybe with :title)

src/components/Controls.vue
```diff
@@ -361,6 +368,23 @@ export default {
 				name: 'board.details',
 			}
 		},
+		// Controls is shared by the board, the board list and the overviews.
+		// Only the first two have something that consumes the query.
+		isBoardList() {
+			return !this.board && !this.overviewName
```
grnd-alt: not a fan of this way of deciding when to show what in child components. Would be better to have showFilter and filterLabel be a prop and let the parent decide what to show.

grnd-alt (changes requested):
I'm unsure that's how the small screen is supposed to look like, maybe either 100vw or a fixed width's better
Image

kra-mo (approved):

grnd-alt (approved):

PR #8234, opened 2026-08-04
refactor(pinia): move board logic to boardStore

grnd-alt:


src/components/navigation/AppNavigationAddBoard.vue
```diff
@@ -81,10 +82,7 @@ export default {
 		async createBoard(e) {
 			this.loading = true
 			const title = this.value.trim()
-			await this.$store.dispatch('createBoard', {
-				title,
-				color: this.color.substring(1),
-			})
+			await useBoardStore().createBoard({ title, color: this.color.substring(1) })
```
mejo-: Is there a particular reason why you don't use `mapActions(useBoardStore, ['createBoard'])` here? Maybe I'm missing something, but it seems like inconsistent code style to me.

src/components/Controls.vue
```diff
@@ -302,8 +302,9 @@ import SessionList from './SessionList.vue'
 import { isNotifyPushEnabled } from '../sessions.js'
 import CreateNewCardCustomPicker from '../views/CreateNewCardCustomPicker.vue'
 import { getCurrentUser } from '@nextcloud/auth'
-import { mapActions } from 'pinia'
+import { mapActions, mapState as mapStatePinia } from 'pinia'
```
mejo-: In other places you used the pattern to rename the Vuex version to `mapStateVuex` which seems cleaner to me as you don't have to rename again once Vuex gets removed.

src/components/Controls.vue
```diff
@@ -422,26 +424,27 @@ export default {
...
 			} else {
 				this.filter.completed = 'both'
-				this.$store.dispatch('setFilter', { ...this.filter })
+				this.setFilter({ ...this.filter })
 			}
-			this.$store.dispatch('setFilter', { ...this.filter })
+			this.setFilter({ ...this.filter })
 		},
-		setFilter() {
+		setFilterLocal() {
```
mejo-: If I'm correct, in other places you used the pattern to import the store method as `setFilterInStore()`, so again you introduce a different code style here.

mejo- (approved):
I have some nitpicking code style comments. Other than that, it looks good. Didn't test myself and only eyeballed the code changes.

PR #8237, opened 2026-08-04
fix(trash): allow to configure the trash retention in hours

mejo-:
Also change default from 5 minutes to 5 hours. The cronjob was configured to run only once a day anyway, so the deletion took place sometime between 5 minutes and 1 day after the item was trashed.

Assisted-by: OpenCode:claude-fable-5

### Checklist


grnd-alt (approved):

mejo-:
08e8c4e84bed64fbff489daa94c39f60871f625e fixes unrelated phpunit failures that got introduced sometime in the past.

PR #8266, opened 2026-08-13
feature: Allow a team (circle) to own a board  

### LINKED ISSUE #2778: Support creating Deck boards owned by teams
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.

---

## Feature request
We can (soon) create deck board owned by team. API/PR: https://github.com/nextcloud/deck/pull/8266

We should make sure the Teams UI supports it!

* [ ] add "create Deck board" button to Teams page
* [ ] automatically create Deck board upon creation of new team

### LINKED ISSUE #8167: Allow a team (circle) to own a board
### How to use GitHub

In a organization, people come and go. This can be a problem when they own resources that are actually used/relevant for a larger group of people. For Files, this was resolved via Teamfolders, and Collectives resolved it by always having a circle as owner. Deck should follow the same/a similar approach.

This is part of a wider effort to ensure team ownership is possible for all resources: https://github.com/nextcloud/circles/issues/2644

We have 3 types of roles in teams right now:
* owner (only person who can delete the team)
* moderators (can add/remove team members, rename team etc)
* normal members

We will add 'guests' (read-only members). After that, a rich role mechanism is on the agenda. This would including a custom role mechanism so complex permissions from a checklist ("can manage X") can be assigned for each role, then roles can be assigned to team members. Other settings we are working on is to dis-allow re-sharing on a team level, ideally Deck can respect that setting.

Much of that is future, but for Deck now the core issue team ownership should resolve is that no single person can block work or cause issues when unavailable or having left the company.

It should be possible to designate a team as owner.

Based on the roles in teams that currently exist, a mapping like this would perhaps make sense:
* moderators can act as owner
* normal members can act as normal members
* guests have read-only access (future feature of Circles/Teams, so probably not for 35)

There are other questions and I thought of what I think are reasonable answers, but the designers might have different ideas of course:
* can all team members do everything? Are they all owners?
    * See mapping above. I think the team members should have rights that follow from their rights in the Team. That is, team managers should have the ability of an owner (able to change ownership, mainly), other members should have the access rights they have in the team. Currently, teams have no specific access rights, to that just means everything. In the future, teams will likely get at least an option to block re-sharing of resources from the team (which means the team this board is assigned to is the ONLY group of people who can have access - sharing should not be allowed). And we will probably introduce read-only team members.
        * this then adds the question: what happens to other current shares, esp if the new owning-team has a no-sharing policy. Remove or retain? I would suggest we warn the team managers about this in a notification, but retain the shares. But warning that sharing with the no-share team results in removing the other shares is also a legit approach... Maybe easier.
* What happens when a Team is removed?
    * just like with Deck boards of a user - delete. We will probably have to introduce an 'archive' function for teams, which should also archive their boards.
* What happens when Teams are not available?
    * you should not be able to transfer ownership to a team...
* What happens to you - can you still see the board, even if you're not in the team?
    * my suggestion is to add the user who does the ownership transfer to the board, unless they are in the team, and unless the team is has a no-sharing policy. Reason is that, in most cases, you were/are probably working on this task with others and now hand over ownership. If that is not the case, and should no longer have access, you can leave yourself, or an owner can remove you - this is not disruptive. But losing access right away until somebody re-adds you IS disruptive.

There could be a no-owner model, like Talk, where boards exist on their own... But I think this makes big orgs, who want tight ownership and clear responsibilities, very unhappy.

**Additional context**
I did an AI assisted PR earlier - but did not have the time to work on it more:
https://github.com/nextcloud/deck/pull/7899

samin-z:
Feature ticket: #8167 

### Summary
Allow teams to create boards belonging to them.

### DONE
1. add new api to add a Team board.
2. on team owner leaving the team, the ownership of the board is transfered to the user with next highest level.
3. on team deletion, or all members leaving the team, all the boards belonging to that team also get removed.

Since there is no FE yet, to test this API can be called to add new Team board: 
/deck/api/v1.0/boards/team

body:
{
    "title": "sample title",
    "teamId":"sample team id",
    "color": "sample color" (this filed is optional, can also not be sent since on the current design we have the board is created with no color)
}

### Checklist

### DIFF
--- a/lib/Controller/BoardApiController.php
+++ b/lib/Controller/BoardApiController.php
@@ -87,6 +87,17 @@ class BoardApiController extends ApiController {
 	}
 
 	/**
+	 * Create a board attached to a team (circle).
+	 */
+	#[NoAdminRequired]
+	#[NoCSRFRequired]
+	#[CORS]
+	public function createForTeam(string $title, string $teamId, ?string $color = null): DataResponse {
+		$board = $this->boardService->createForTeam($title, $this->userId, $color, $teamId);
+		return new DataResponse($board, HTTP::STATUS_OK);
+	}
+
+	/**
 	 * Update a board with the specified boardId, title and color, and archived state.
 	 */
 	#[NoAdminRequired]
--- a/lib/Controller/BoardApiController.php
+++ b/lib/Controller/BoardApiController.php
@@ -87,17 +87,6 @@ class BoardApiController extends ApiController {
 	}
 
 	/**
-	 * Create a board attached to a team (circle).
-	 */
-	#[NoAdminRequired]
-	#[NoCSRFRequired]
-	#[CORS]
-	public function createForTeam(string $title, string $teamId, ?string $color = null): DataResponse {
-		$board = $this->boardService->createForTeam($title, $this->userId, $color, $teamId);
-		return new DataResponse($board, HTTP::STATUS_OK);
-	}
-
-	/**
 	 * Update a board with the specified boardId, title and color, and archived state.
 	 */
 	#[NoAdminRequired]
--- a/lib/Controller/BoardController.php
+++ b/lib/Controller/BoardController.php
@@ -51,11 +51,6 @@ class BoardController extends ApiController {
 	}
 
 	#[NoAdminRequired]
-	public function createForTeam(string $title, string $teamId, ?string $color = null): Board {
-		return $this->boardService->createForTeam($title, $this->userId, $color, $teamId);
-	}
-
-	#[NoAdminRequired]
 	public function update(int $id, string $title, string $color, bool $archived): Board {
 		return $this->boardService->update($id, $title, $color, $archived);
 	}
--- a/lib/Controller/BoardController.php
+++ b/lib/Controller/BoardController.php
@@ -51,6 +51,11 @@ class BoardController extends ApiController {
 	}
 
 	#[NoAdminRequired]
+	public function createForTeam(string $title, string $teamId, ?string $color = null): Board {
+		return $this->boardService->createForTeam($title, $this->userId, $color, $teamId);
+	}
+
+	#[NoAdminRequired]
 	public function update(int $id, string $title, string $color, bool $archived): Board {
 		return $this->boardService->update($id, $title, $color, $archived);
 	}
--- a/lib/Listeners/ParticipantCleanupListener.php
+++ b/lib/Listeners/ParticipantCleanupListener.php
@@ -8,43 +8,61 @@
 namespace OCA\Deck\Listeners;
 
 use OCA\Circles\Events\CircleDestroyedEvent;
+use OCA\Circles\Events\RemovingCircleMemberEvent;
+use OCA\Circles\Model\Member;
 use OCA\Deck\Db\Acl;
 use OCA\Deck\Db\AclMapper;
 use OCA\Deck\Db\AssignmentMapper;
 use OCA\Deck\Db\BoardMapper;
+use OCA\Deck\Service\TeamBoardService;
 use OCP\EventDispatcher\Event;
 use OCP\EventDispatcher\IEventListener;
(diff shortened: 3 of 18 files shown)


lib/Controller/BoardApiController.php
```diff
@@ -86,6 +86,17 @@ public function create(string $title, string $color): DataResponse {
 		return new DataResponse($board, HTTP::STATUS_OK);
 	}
 
+	/**
+	 * Create a board attached to a team (circle).
+	 */
+	#[NoAdminRequired]
+	#[NoCSRFRequired]
+	#[CORS]
+	public function createForTeam(string $title, string $teamId, ?string $color = null): DataResponse {
```
grnd-alt: would prefer to have new endpoints only in the ocs controllers.

lib/Service/CirclesService.php
```diff
@@ -105,4 +105,46 @@ public function getUserCircles(string $userId): array {
...
+		if (!$this->circlesEnabled) {
+			return null;
+		}
+
+		try {
+			$circlesManager = Server::get(CirclesManager::class);
+			$circlesManager->startSuperSession();
+			$circle = $circlesManager->getCircle($circleId);
+			$circleMembers = [];
+			foreach ($circle->getMembers() as $member) {
```
grnd-alt, line 123: is there not always an admin for a team?

can't we just search for the admin and give them the ownership? then we don't have to search through all the members and sort them and all that.

lib/Controller/BoardController.php
```diff
@@ -50,6 +50,11 @@ public function create(string $title, string $color): Board {
 		return $this->boardService->create($title, $this->userId, $color);
 	}
 
+	#[NoAdminRequired]
+	public function createForTeam(string $title, string $teamId, ?string $color = null): Board {
```
grnd-alt: also this preferably in ocs

lib/Listeners/ParticipantCleanupListener.php
```diff
@@ -8,43 +8,61 @@
...
 
 	public function handle(Event $event): void {
 		if ($event instanceof UserDeletedEvent) {
-			$boards = $this->boardMapper->findAllByOwner($event->getUser()->getUID());
+			$userId = $event->getUser()->getUID();
+			$transferredBoardIds = $this->teamBoardService->transferTeamBoardsFromDeletedUser($userId);
+
+			$boards = $this->boardMapper->findAllByOwner($userId);
 			foreach ($boards as $board) {
+				if (in_array($board->getId(), $transferredBoardIds, true)) {
```
grnd-alt: we could drop some readability to prevent doing the same db query twice (findAllByOwner)
if we drop the `transferTeamBoardsFromDeletedUser` function and incorporate it's logic into the for loop we already have. Not required by me though.

Either way I think the `transferredBoardIds` array has no effect on the logic here right? All the ids in there have a different owner by now so are not in the `$boards` array anyway...

grnd-alt (changes requested):
works well and code is clean, got some small nitpicks though otherwise looks good 🔥

lib/Service/CirclesService.php
jospoortvliet: FWIW, Teams currently have an owner. Could be a temporary solution - but note that  but we want to remove that role, as any role with a single person is something we want to avoid. Teams have admins and  moderators - can have multiple of those.

Background that might be helpful: in the next release we will try to align team roles with roles in apps. I mean so that Team moderators would be moderators in Talk rooms owned by the team. And in the far future, we would like to be able to create and manage roles of team members (can manage ACL's, can create talk rooms, can moderate talk rooms, ...).

grnd-alt: all right sounds to me like there is always _at least one_ admin on a team, so we can just pick the first that comes up @samin-z

grnd-alt (approved):

PR #8277, opened 2026-08-19
fix(migration): support the public OCP\DB\Schema API

AndyScherzinger:
* Resolves: # 
* Target version: main

### Summary

Nextcloud 35 replaced the Doctrine DBAL return types of ISchemaWrapper with the public OCP\DB\Schema interfaces (server#63013). Bump the nextcloud/ocp dev dependency so the shipped OCP stubs carry the new signatures instead of shadowing the server ones with the pre-35 declarations, and point the changeSchema() @throws at the OCP SchemaException that ISchemaWrapper::getTable() now raises.

The composer platform requirement is raised to PHP 8.3 to match the minimum of Nextcloud 35, which nextcloud/ocp now also requires.

That @throws was the only docblock-only reference to an undefined class left in the code base, so the UndefinedDocblockClass suppression no longer suppresses anything and Psalm reports it as an UnusedIssueHandlerSuppression. Psalm counts suppressions per errorLevel filter rather than per referencedClass, so the whole handler has to go. Undefined classes referenced outside docblocks stay covered by the UndefinedClass handler.

### TODO

- [x] manual update tests 
- [ ] CI run

### Checklist

### 🤖 AI (if applicable)

- [x] The content of this PR was partly or fully generated using AI

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
 			"bamarni/composer-bin-plugin": true
 		},
 		"platform": {
-			"php": "8.2"
+			"php": "8.3"
 		}
 	},
 	"scripts": {
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "1e7c5e7f0772434d255c0804f3c6824e",
+    "content-hash": "d3c4dd16a9d9702b12542dd5b6ada365",
     "packages": [
         {
             "name": "bamarni/composer-bin-plugin",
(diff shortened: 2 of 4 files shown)


CarlSchwan (approved):

CarlSchwan:
Technically not required for the good working of the app, but a nice cleanup

AndyScherzinger:
> Technically not required for the good working of the app, but a nice cleanup

@CarlSchwan What do you mean? Without the fix, I can't activate the main branch on a v35 server. Also since tables also has a 33-35 version range, this implies the change is backwards compatible? (I think Louis mentioned it when I asked, but just to be sure)

CarlSchwan:
Very weird that it doesn't work without rhe change: SchemaException is just a doc block and the php interpreter doesn't care about it

AndyScherzinger:
> Very weird that it doesn't work without rhe change: SchemaException is just a doc block and the php interpreter doesn't care about it

@CarlSchwan Prior to having merged this PR, when you try to activate deck, it silently fails and this is the log output:

`Declaration of OC\\DB\\SchemaWrapper::getTable(string $tableName):`
**`OCP\\DB\\Schema\\ITable must be compatible with OCP\\DB\\ISchemaWrapper::getTable($tableName)`**
`at /var/www/html/lib/private/DB/SchemaWrapper.php#76`


Full json element:

```
{"reqId":"HzZAwJgf7tkXhCunC1Ev","level":3,"time":"2026-08-19T21:29:10+00:00","remoteAddr":"192.168.21.3","user":"admin","app":"PHP","method":"POST","url":"/ocs/v2.php/apps/appstore/api/v1/apps/enable","scriptName":"/ocs/v2.php","message":"Declaration of OC\\DB\\SchemaWrapper::getTable(string $tableName): OCP\\DB\\Schema\\ITable must be compatible with OCP\\DB\\ISchemaWrapper::getTable($tableName) at /var/www/html/lib/private/DB/SchemaWrapper.php#76","userAgent":"Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:154.0) Gecko/20100101 Firefox/154.0","version":"35.0.0.4","data":{"app":"PHP"}}
```

CarlSchwan:
@AndyScherzinger 

This should be a better fix https://github.com/nextcloud/server/pull/63427 I think the old version of the ISchemaWrapper is still loaded during the update, so we need to ensure SchemaWrapper is compatible with both the new version and the old one.

AndyScherzinger:
@CarlSchwan nice finding then 👍 Happy to here if there is a solution improving the situation (for others - given we already merged the fix for deck)

PR #8325, opened 2026-08-27
feat(filter): persist board filters (saved views + default view)

### LINKED ISSUE #2861: [Feature request] Saveable/storable filters/views
### How to use GitHub

I have common filters that I use on a daily base, mostly to filter based on tags depending on which project I am working on. It's not hard to set the filter, and the filters work as I'd expect, but it'd be nice to be able to have a saved view of a board, with pre-set filters.

When viewing a filtered board, I'd like an option to "Save View" either in the filtering dropdown (feels relevant, but maybe wrong) or in the three-dots menu for the whole board. This would prompt for a name for the "View", and save the current status of the view (i.e., the active filters) somewhere.

In the list of boards, underneath a given board (say my "Research" board) I'd like to have listed my saved views. These would probably have some indicator that they are saved views, either by slight indentation or prefixing the name of the view with the name of the board. Clicking on a view would load the relevant board and apply the filters.

My current approach is just to manually set filters, which as I mention is not too hard, it's just an extra few button presses whenever I go to a different "view" but it does feel annoying and clunky to do.

**Additional context**
Issue #414 mentions representing the filter in the URL, which could be one way of achieving this (save the URL, and load that URL when clicking the view), except that as best I can tell, changing a filter does not update my current URL, and it's a lot smoother from a UI perspective to not have a full page reload to see a different view.

heinrich-toews:
* Resolves: #2861
* Target version: main

### Summary

This changeset implements the requested "Save filters" feature by letting users **save deck board filters as named views** and re-apply them later, plus the ability to **set one saved view as the default** for a board.

**Saved board views**
- New `BoardView` entity, mapper, service and OCS API controller.
- CRUD routes under `deck/api/v1.0/views` (index/create/update/delete).
- The filter popover in `Controls.vue` lets you save the active filter as a named view and re-apply it.

**Default view**
- `ConfigService::defaultBoardView` persisted per user + board.
- New picker in `DeckAppSettings.vue` (per-user Default settings) to choose the default view.
- When opening a board, the default view is applied automatically; tags are only applied where they match the current board.

**Migration**
- `Version11003Date20260821120000` creates `oc_deck_board_views`.
- Numbered `11003`, safely above the current latest `11002`.

### Testing steps

1. Apply this branch and run `occ upgrade` (or it runs via the updater).
2. Open a board, set a filter (label/priority/tag/assignee/…), choose **Save as view** and name it.
3. Confirm the view appears in the filter popover, re-applies when clicked, and can be edited/deleted.
4. Go to **Deck settings → Default**, pick the saved view as the default.
5. Open a different board / reload: the default view is applied automatically (tags only where they match).

*Note:* `l10n/*` is intentionally not included per the project's Transifex workflow — translations are managed upstream.

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -144,6 +144,7 @@ return [
 		['name' => 'board_ocs#createForTeam', 'url' => '/api/v{apiVersion}/boards/team', 'verb' => 'POST'],
 		['name' => 'board_ocs#addAcl', 'url' => '/api/v{apiVersion}/boards/{boardId}/acl', 'verb' => 'POST'],
 
+		['name' => 'board_view_api#all', 'url' => '/api/v{apiVersion}/views', 'verb' => 'GET'],
 		['name' => 'board_view_api#index', 'url' => '/api/v{apiVersion}/boards/{boardId}/views', 'verb' => 'GET'],
 		['name' => 'board_view_api#create', 'url' => '/api/v{apiVersion}/boards/{boardId}/views', 'verb' => 'POST'],
 		['name' => 'board_view_api#update', 'url' => '/api/v{apiVersion}/boards/{boardId}/views/{viewId}', 'verb' => 'PUT'],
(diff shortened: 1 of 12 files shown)


heinrich-toews:
Hi — this is a first contribution from a fork, so the GitHub Actions workflows are pending approval. Could a maintainer approve the workflow runs so CI can execute? Thanks!

PR #8359, opened 2026-09-09
enh(performance): get deck share permissions in one query

grnd-alt:


lib/Service/PermissionService.php
```diff
@@ -63,6 +63,31 @@ public function setAccessToken(string $token) {
...
 	}
 
+	/**
+	 * Get permissions for all supplied card ids
+	 * @param int[] $cardIds
+	 * @param ?string $userId
+	 * @param bool $allowDeleted
+	 * @return array<int, array<Acl::PERMISSION_*, bool>>
+	 */
+	public function getPermissionsForCards(array $cardIds, ?string $userId = null, bool $allowDeleted = false): array {
```
benjaminfrueh, line 73: The param $allowDeleted is currently unused and should be maybe removed.

benjaminfrueh (approved):

PR #8388, opened 2026-09-15
enh: validate datetime fields

### LINKED ISSUE #8369: Invalid extreme date time value accepted in PostgreSQL causes "Double time specification" error
In PostgreSQL, an extreme timestamp value (for example, 99999-09-11 02:00:00) is accepted as valid and stored. This causes errors "Double time specification" when parse record (especially when run background job CardDescriptionActivity). So, the date time values should be validated before being saved into db.

luka-nextcloud:
* Resolves: #8369
* Target version: main

### Summary
- Improve validation for date fields in card creation and update operations, ensuring that dates are properly formatted and do not exceed the year 9999. 
- Also enhance datetime fields handling on frontend to prevent invalid date selections.

### Checklist


lib/Validators/BaseValidator.php
```diff
@@ -135,6 +149,27 @@ private function min($value, $limit): bool {
 		return $this->getSize($value) >= $limit;
 	}
 
+	/**
+	 * @throws Exception
+	 */
+	private function datetime($value): bool {
```
blizzz:
```suggestion
	private function datetime(mixed $value): bool {
```

nitpick: should be possible meanwhile and shows intent

Q: this does not check about formats, so '+1 day'  would also pass. Valid? Empty string would also pass.

src/components/card/DueDateSelector.vue
```diff
@@ -162,7 +162,12 @@ export default defineComponent({
 				return this.card?.duedate ? new Date(this.card.duedate) : null
 			},
 			set(val) {
-				this.$emit('input', val ? new Date(val) : null)
+				// Prevent setting a date with a year greater than 9999, as this can cause issues with some date libraries and databases.
+				const date = val ? new Date(val) : null
+				if (date && date.getFullYear() > 9999) {
```
blizzz: nitpick: `9999` preferable in a constant, and re-used in the other Selector.

Q: is a very low value also problematic? < -9999 or such?

lib/Service/CardService.php
blizzz: By-catch: this setDueDate setter call differs from that one in lin 306  `$card->setDuedate($duedate ? new \DateTime($duedate) : null);` (same with the start date below). Bit unrelated this change, but if an empty string is passed, the behaviour seems to be different currently.

blizzz (commented):
Already stored values are not repaired, right? One-time repair job could be considered, or command. But I guess it is an edge case, though would be nice for affected users/admins to get out of it?

src/components/card/DueDateSelector.vue
luka-nextcloud: The min year allowed to input in `NcDateTimePickerNative` component is `0001`

lib/Validators/BaseValidator.php
luka-nextcloud: I have updated to check the datetime formats.

